PR #14, opened 2022-05-18
Add types for basic terminology

### LINKED ISSUE #9: osm2streets terminology and structure
I find myself wanting to be precise about particular components or aspects of a road, but lacking the terminology. I think this is a good time to agree on some terminology that will allow us to describe the structure we want to aim for. I can add the terms as a primer to the README after some input.

Let me start with one way of slicing it up, and let's change it to match up with what is already used.

- `Lane`: a single lane with designation, width, etc.
- `Buffer`: medians, painted buffers, verges, ... with width etc.
- `RoadWay` ~~`Road`~~ : a collection of Lanes (and Buffers), travelling in the same direction; A "half street" if you will, or a slip road or slip lane.
- `Intersection`: the area where multiple lanes of travel overlap and the road endpoints that join it.
- `Street` or `Corridor`: a collection of `RoadWays` (and cycletracks/footpaths) and `Buffers` that all run (mostly) parallel to one "centerline", representing the whole corridor and the "curb" "building to building".
- `Junction`: A collection of Intersections and RoadWays that make up one "junction". E.g., a roundabout is `RoadWay`s around with one `Intersection` per arm.
- `RoadNetwork`: A directed graph of Roads connecting Intersections
- `StreetNetwork`: An undirected graph of Streets connecting Junctions

Those feel like the types I am expecting to find, which makes me want
- `geom::{Point, LineString, Area, ...}` and all the right algos, and
- `osm::{Node, Way, Relation, Tag, Key, Tags}` as well understood boring types.

How are these concepts weird or difficult or bad?

BudgieInWA:
Closes #9.


src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    Outward,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum DrivingSide {
+    LHT,
+    RHT,
+}
+
+use Direction::*;
```
dabreegster, line 56: What does importing an enum like this mean? I don't think I've ever seen just `LHT` used in code to mean `DrivingSide::LHT`, for example

src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    Barrier,
+}
+
+/// A single lane on the carriageway, with designation, width, etc.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Lane {
+    dir: Directions,
+    designation: Designation,
+    width: Meters,
+    can_enter_from_inside: bool,
```
dabreegster, line 120: What's inside/outside in the context of lanes possibly with alternating/multiple directions?

src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    can_enter_from_outside: bool,
+}
+
+struct CrossWay(Lane);
+
+impl CrossWay {
+    fn unmarked() -> CrossWay {
```
dabreegster, line 127: Hmm, treating crossings as lanes? I have seen designated foot / cycle crossings now. Is this meant to eventually be part of representing junctions with multi-stage turns or crossing islands and stuff?

dabreegster (commented):
Just skimmed through. It's really awesome to see a fresh and much richer representation! I'm very confused by lots of this; I think once the idea-dust settles, example diagrams with labels would be a useful exercise

src/lib.rs
BudgieInWA: relative to the roadway, so `can_enter_from_road_inside: bool` might be better. I think I'll change `LaneSide` to `RoadSide`...

src/lib.rs
BudgieInWA: You are on the right track, it `use`s each option in that scope, so that you could quote `LHT` to mean `DrivingSide::LHT`.

It's like a prelude that lets you say things like `Some(Inside)`.

BudgieInWA:
> Just skimmed through. It's really awesome to see a fresh and much richer representation! I'm very confused by lots of this; I think once the idea-dust settles...

I will give some thought to documentation lines for a bunch of things, and some grouping in modules.

Then I want to build a piece of logic using this, as an example context from which to understand each term

> example diagrams with labels would be a useful exercise

Yes! Diagrams of each classic kind of intersection and road. Is basically our most basic test suite, I guess. 

<details><summary>Makes me think of the diagrams in Randall Munroe's Thing Explainer.</summary>

Thing Explainer pages

</details>

src/lib.rs
BudgieInWA: > Is this meant to eventually be part of representing junctions with multi-stage turns or crossing islands and stuff?

Yes. We start out by thinking about the fully detailed graph of travel ways, which is about as complicated as the OSM graph we will find. Then we try to make sense of the structure of it, bit by bit. Then when we've understood as much as we can, we try to realise what is there (guessing areas, curbs, markings, etc.). Grouping everything into a "Junction" feels like a pretty high level concept, and doesn't have much bearing on stuff like determining routing and interpreting road structure. But a concept that is very cool for stuff like giving verbal directions.

E.g. If we happen to find a footway crossing over a road, that's a pretty obvious crossing even if it's not tagged as such: probably an unmarked crossing, with a gap in the median and lowered curbs, but nothing else.

src/lib.rs
droogmic: iirc, that is why C++ removed un-namespaces enums from C...

src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+pub type Meters = f32;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum End {
+    Front,
+    Back,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Direction {
```
droogmic, line 14: Direction vs Directions vs LaneDirections is going to be confusing in the long-run ;)

droogmic (commented):
some of these concepts also apply to osm2lanes.

I assume osm2streets will depend on osm2lanes (and not vice versa), so we can hoist these into osm2lanes or even some more fundamental type crate when the time comes.

BudgieInWA:
I have split things into `mod`s which helps *me* keep things straight in my head. A good place to start looking is the little example at the bottom of `networks.rs` which builds a graph that looks like this:

abbigail dot

@droogmic 

> I assume osm2streets will depend on osm2lanes (and not vice versa), so we can hoist these into osm2lanes or even some more fundamental type crate when the time comes.

Yes, I imagine that too. Some types will be replaced with those from osm2lanes as soon as we depend on it. osm2lanes feels to me like a good place for these concepts to live long term. We probably want to hoist the `osm` crate out of abstreet though.   

> I'm not sure I would do it for all types though at a global level, it will probably cause issues later.

I've put those re-exports in a "preamble" mod for easy access on demand for those who want it. I do like doing it for enums with distinct variants, like most of those in `mod units`.

> Direction vs Directions vs LaneDirections is going to be confusing in the long-run ;)

Yes. Some of it is inherent complexity (which is why it makes it useful to have `Left` vs `LHT` vs `Outside`), so we'll need to be clever about naming of those types. I wonder if naming the enums `LeftRight`, `InsideOutside` and `ForwardBackward` makes sense?

src/road_parts.rs
```diff
@@ -0,0 +1,203 @@
+use crate::road_parts::RoadEdge::Kerb;
+use crate::units::preamble::*;
+use crate::units::{Meters, RoadSide, TrafficDirections};
+use Carriage::*;
+use Designation::*;
+
+/// Some hunk of something hurtling or dawdling down some lane, or being stored somewhere.
```
dabreegster, line 7: Docstrings bordering on poetry ;)

BudgieInWA:
Let's move on to smaller pieces. Open PRs or issues for anything that hasn't hit the mark, or just drop a comment here.

Issue #8, opened 2022-05-02
Start a test suite


BudgieInWA:
We want to be able to do something like
1. download a `.osm` of a small area,
2. fill in one of the expected outputs somehow, and
3. get the test to run.

Having test cases in place will be useful even before any implementations show up.

I can imagine fancy tests having these attributes:
- bounds: real world examples
- `.osm` xml: the example input
- network graph: rawmap graph structure and properties
- rawmap geojson: for snapshot testing/goldenfile testing
- ground truth: orthographic imagery for context when comparing snapshots

Would love to be able to write code against a *network graph from `.osm`* test.

BudgieInWA:
@dabreegster, with the addition of some docs explaining an approach for

1. running tests: from `cd` to `cargo test`,
2. comparing and updating the goldenfiles, and
3. adding a new test (doing the goldenfile update pattern I guess),

I am happy to call the test suite "started", close this issue, and hack whatever I need on top of the existing tests, as I need them.

dabreegster:
A separate test per directory would be ideal, I didn't look into it yet. That crate looks like it could work nicely!

So my TODOs:
- rename cfg.json -> test.json and output.json -> raw_map.geojson
- docs
- split tests

BudgieInWA:
The very promising `tests` crate hits the nail on the head, testing multiple projects under the same roof.

PR #15, opened 2022-05-21
Start a small test suite of OSM XML -> RawMap GeoJSON goldenfiles. #8

dabreegster:
See `src/lib.rs` for a description of what this does. In short, we can clip out a osm.xml file for some tricky area, run it through A/B Street's current code, and save the resulting polygons as geojson. As we iterate on that code, we can see when the output changes, making this a convenient regression test.

I'm starting this code in this repo and not A/B Street to:
1) avoid making the main abstreet git repo huge with all the OSM and GeoJSON files
2) nudge towards extracting the osm2streets-bits into proper standalone crates and repos

@BudgieInWA, there are many other ideas you mentioned in the issue -- consider this just a start.

CC @martinfleis, who has some Google summer of code students also working on street network simplification. At the very least, we could share the input test cases that cause problems. There are only 3 included in this PR so far, but I have many more in mind to add.


dabreegster:
You can view the current output with github's viewer, https://github.com/a-b-street/osm2streets/blob/main/tests/src/aurora_sausage_link/output.json

martinfleis:
Thanks for CCing me in.

> There are only 3 included in this PR so far, but I have many more in mind to add.

Can you keep me in the loop on those as well?

dabreegster:
@martinfleis, sure -- I added more in a few followup commits, all at https://github.com/a-b-street/osm2streets/tree/main/tests/src. If you come across any good examples in OSM, please send them over and we can add them here. When we actually start trying to simplify some of these scenarios, I can CC you on those PRs too

tests/src/lib.rs
```diff
@@ -0,0 +1,91 @@
+//! This is a goldenfile test for osm2streets. Each test case has a directory with:
+//!
+//! - `input.osm`, from the export tab of https://www.openstreetmap.org
+//! - `cfg.json`, defining the `driving_side` and describing the situation
+//! - `output.json`, a GeoJSON file showing the resulting road and intersection polygons
```
BudgieInWA, line 5: Can we call these `test.json` (or `test.yml`), `input.osm`, `rawmap.geojson`? I want to add tests of different kinds that sit along side the RawMap GeoJSON ones, so lets name each output what it is.

I might go for `roads.dot` (then `corridors.dot`)...

PR #17, opened 2022-05-23
Plan: Iterate on the ideas here, approaching what `RawMap` does while iterating on the abstreets side towards this

BudgieInWA:
This PR proposed that we ought to
- Iterate on the api/docs here, working towards [abstreet/RawMap](ttps://github.com/a-b-street/abstreet/blob/master/raw_map/src/lib.rs) 
- Iterate on `RawMap` working towards modularity and the ideas in the emerging api/docs
  - If `RawMap` get good enough, we we can drop the experimentation on the api,
  - otherwise, one by one, all the pieces eventully get integrated into the api

-----

What do we think? The value of the knowledge embedded in abstreet (of interest to osm2streets) is enormous, but we also want the ability to work on the questions that osm2streets poses, uncoupled from untangling abstreets pieces.

I'm gonna work on transformations in the least friction way, which to me, looks like a layer built on `petgraph::StableDiGraph`, and probably looks like `RawMap` (but untangled) to @dabreegster. Hopefully we can meet in the middle soon, replacing both with an iteration on one of them. I'm happy to drop whatever I have written here, in favour of working replacements from abstreets, as they get untangled. Getting into the abstreets ecosystem (with viewers and editors etc.) will be reward enough. My hope is largely that a fresh perspective on the problem might exhibit some different and useful ideas -- that can be pinched for the untangling work.

And some small pieces will just slot in. I'm keen, for example, to get in a copy of `convert_osm` that builds a `RoadNetwork` instead of a `RawMap`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,10 @@ OSM into a cleaned-up street network graph with geometry.
 
 ## Project status
 
-This repository is conspicuously missing code. The osm2streets effort is
-underway, but splitting the code from the A/B Street codebase isn't easy. (It
-is the ultimate goal, though.) Currently the logical parts of osm2streets are
-scattered around:
+This repository is conspicuously missing functionality. The osm2streets effort is
+underway, but splitting the code from the A/B Street codebase will take some time.
+We will iteratively import and move in different logical parts of the sm2streets scope
+from around the place:
 
 - [osm2lanes](https://github.com/a-b-street/osm2lanes) determines the lanes
   along one OSM way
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,10 +9,10 @@ OSM into a cleaned-up street network graph with geometry.
 
 ## Project status
 
-This repository is conspicuously missing code. The osm2streets effort is
-underway, but splitting the code from the A/B Street codebase isn't easy. (It
-is the ultimate goal, though.) Currently the logical parts of osm2streets are
-scattered around:
+This repository is conspicuously missing functionality. The osm2streets effort is
+underway, but splitting the code from the A/B Street codebase will take some time.
+We will iteratively import and move in different logical parts of the sm2streets scope
```
dabreegster, line 14:
```suggestion
We will iteratively import and move in different logical parts of the osm2streets scope
```

dabreegster (approved):
I like this two-pronged plan and meeting in the middle somewhere. I'm cognizant that the current A/B Street has lots of functionality wired together, but also has technical debt and is burdened with old assumptions. I also may need to make sudden quick progress on things to unblock bits of my day job activities. Being able to experiment and start with a fresh API without being burdened by all of that is perfect.

>  looks like a layer built on petgraph::StableDiGraph

Does that assign node and edge IDs itself? One struggle I've had with using OSM node and way IDs comes from having to invent fake new OSM IDs internally when creating / modifying structures based originally on "pure" OSM.

> I'm keen, for example, to get in a copy of convert_osm that builds a RoadNetwork instead of a RawMap.

That'd likely be a good step! `convert_osm` just focused on the street network part of OSM is not that huge of a crate. Are you thinking to convert a `RawMap` into a `RoadNetwork` (I'm guessing not), to build something fresh based on `convert_osm` (extracting common library pieces along the way), or come up with generic-ish code that processes OSM and can emit structures in either world?

One argument for starting a bit fresh might be that abst isn't actually using osm2lanes yet (it's proving to be a painful transition precisely because abst has so many different special cases in cities that I need to keep working)

PR #20, opened 2022-05-31
Implement loading of .osm to RoadNetwork via RawMap

### LINKED ISSUE #19: Implement loading `.osm` to a `RoadNetwork`
I would like to have the ability to work on real examples, and tap into the tests, as my next step. The `tests` crate is already doing the reading and parsing of `.osm` files, by importing from `abstreet`, so outputting `RoadNetwork` instead of `RawMap` shouldn't be too far away. 

I think I want the initial `RoadNetwork` to be split up into as many pieces as needed, with intersections broadly categorised (incomplete, terminus, slice, intersection). That way, transformations can work mostly on combining the pieces into bigger structures, and collapsing unnecessary detail; I feel like doing all the splitting before all the combining might keep things clear.

Two options:

## 1. Import `reader`, re-implement `extract_osm`
It is tempting to add a `road_network = osm_document.into()` type thing to the clean slate implementation. It looks like the necessary algorithms are in `extract_osm` and `split_ways` (at least), which could be translated into `osm2streets` types instead of `RoadNetwork` types (cutting out a lot of abstreet specific concerns). This requires some effort.

## 2. Import `convert_osm`, load `.osm` as `RawMap` then translate it into `RoadNetwork`
A `road_network: RoadNetwork = raw_map.into()` type thing, that we can run after splitting ways (etc.) might be just as useful. There is a lot of logic in that extraction process that we don't need to revisit yet, but which we might want to use as the basis for the eventual solution. So it should be fine that the clean slate view of things uses `RawMap` as an implementation detail of the `.osm` to most-basic-`RoadNetwork` step.

BudgieInWA:
This PR Closes #19, adding the ability to load a `RoadNetwork` from a `.osm` file; and empowering `tests/` to check `road_network.dot` for all test cases.

It does it by loading up a `RawMap`, then "translating" it into streets types. The translation is still buggy, and will be incomplete for a long time, but that's ok.

A nice side effect is that attempting to translate is a good context for understanding the shortcomings of both systems.


src/io.rs
```diff
@@ -0,0 +1,137 @@
...
+            // raw_int.intersection_type has some useful info, bit is often misleading.
+            t: IntersectionType::Incomplete,
+            control: match raw_int.intersection_type {
+                // IntersectionType::StopSign => ControlType::Signed, // wrong when it should be uncontrolled
+                raw_map::IntersectionType::TrafficSignal => ControlType::Lights,
+                _ => ControlType::Uncontrolled,
+            },
+        }
+    }
+}
```
BudgieInWA, line 137: This is the kind of boon I was hoping for, working on two sides of the same problem. This bit of code is a usecase/testcase that shines a spotlight on a beautifully narrow piece of the `RawMap` solution, and the `RoadNetwork` solution.

E.g. We see that `RawIntersection` can't represent free-running "slice" nodes, and that `Intersection` doesn't neatly represent "type to be determined" vs "border" nodes.

src/io.rs
```diff
@@ -0,0 +1,137 @@
...
+/// println!("{}", Dot::new(&map.graph));
+pub fn load_road_network(path: String, timer: &mut Timer) -> RoadNetwork {
+    let driving_side = DrivingSide::Left; // TODO get driving side from country containing lat lon?
+    let clip = None;
+
+    let mut raw_map = convert_osm::convert(
+        path,
+        MapName::new("zz", "osm2streets_test", &abstutil::basename(&osm_path)),
+        clip,
+        convert_osm::Options {
```
dabreegster, line 36: I'm considering a `convert_osm::Options::common_settings(driving_side)` here; most of these are default/boilerplate. (Not sure about implementing `Default`, since `driving_side` really is required)

src/io.rs
dabreegster: `IntersectionType::Border` should get filled out by `convert_osm/src/clip.rs`. You're right that uncontrolled intersections aren't represented at all yet -- everything's a stop sign, border, or traffic signal. A later pass in `map_model` ("above" the rawmap layer) guesses stop signs; no data at all from OSM is used, other than `highway` ranking. This is all an artifact of my North American bias. Now I'm well aware of how common give way junctions are...

src/io.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// ## Conversions
+
+impl RoadWay {
+    pub fn pair_from(r: &RawRoad) -> EmumMap<Direction, Option<RoadWay>> {
+        //
+        let mut lanes = r.lane_specs_ltr.iter();
+        let ds: DrivingSide;
+        // lanes are ltr, so take the left lanes until we see one in the direction of the traffic
```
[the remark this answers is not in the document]
BudgieInWA: That begs a pretty fundamental question about why we split roads into halfs (and therefore, how we do). I think we are trying to capture the symmetry that is there. The term "contraflow bus lane" shows that we do understand that symmetry and consider this lane to be going "backward on this side".

We should aim to have a theory for why we split roads that makes it clear where the cut should happen. 

I guess it depends the context, where the split should be. A road should split on car lanes only, a cycle track should split on cycle tracks. (Hmm... interpreting a lanes list for a whole corridor as a cycletrack, would put a seperated cycletrack at the center, and the actual road as some irrelevant feature after the barrier on one side... seems helpful)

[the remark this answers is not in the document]
BudgieInWA: It's definitely bugged now. I'll fix anything obviously wrong, but I don't want to block on this.

src/io.rs
BudgieInWA: I've made use of that now.

src/io.rs
```diff
@@ -0,0 +1,198 @@
+use crate::units::preamble::*;
```
dabreegster, line 1: The order of these imports look a bit backwards to me; I usually put (std, external crates, current workspace), per https://github.com/rust-dev-tools/fmt-rfcs/issues/131. But to my knowledge, there's no `cargo fmt` automation for this, only some IDE support.

Just FYI. Feel free to leave them this way; just aim for consistency in the rest of the crate then

src/io.rs
```diff
@@ -0,0 +1,198 @@
...
+    raw_map.run_all_simplifications(false, false, timer);
+
+    raw_map.into()
+}
+
+impl From<RawMap> for RoadNetwork {
+    fn from(map: RawMap) -> Self {
+        let mut net = RoadNetwork::new();
+        // Intersection ids from NodeIds
+        let is = HashMap::<_, _>::from_iter(map.intersections.iter().map(|(node_id, raw_int)| {
```
dabreegster, line 63: Many ways to write this kind of thing. You could also do `let is: HashMap<_, _> = map.intersections.iter().map(...).collect()`, which I find a bit simpler. (I don't know if you can omit the type params like that)

src/io.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // TODO do all needed transformations until it is correct (no unknowns, normal form).
+    }
+}
+
+impl From<&LaneSpec> for Lane {
+    fn from(l: &LaneSpec) -> Self {
+        Lane {
+            dir: if let LaneType::SharedLeftTurn = l.lt {
+                // Lane type is used to represent the both-ways aspect of middle turn lanes, I guess.
```
dabreegster, line 164: Correct, and I think this type always has direction = forwards

dabreegster (approved):

src/io.rs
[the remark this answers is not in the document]
BudgieInWA: I agree, and `is` is definitely due for an upgrade.

In fact, combined with `let blah: HashMap<&NodeID,_>` the comment becomes redundant!

BudgieInWA: I do prefer invoking the common `collect` idiom. It looks like familiar boilerplate which you ought to ignore when you can :)

src/io.rs
BudgieInWA: I actually do prefer that way around. I was distracted by this feeling very `prelude` like (that's what I meant to call it), so wanting to put it first.

tests/src/lib.rs
```diff
@@ -40,10 +43,19 @@ mod tests {
...
-        if prior_output != current_output {
-            std::fs::write("old_raw_map.json", prior_output)?;
-            bail!("{}/raw_map.json has changed. Manually view the diff with geojson.io. If it's OK, commit the new output to git, and this test will pass.", path);
+        let road_network: RoadNetwork = raw_map.into();
+        std::fs::write(format!("{path}/road_network.dot"), road_network.to_dot())?;
+
+        let current_dot = std::fs::read_to_string(format!("{path}/road_network.dot"))?;
+        if current_dot != current_dot {
+            std::fs::write(format!("{path}/road_network.orig.dot"), prior_dot)?;
+            bail!("{}/road_network.dot has changed. Compare it to .orig.dot: if it's OK, commit the new output to git, and this test will pass.", path);
```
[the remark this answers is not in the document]
BudgieInWA: Yes! I found both of those solutions too. 

We should print out `https://` links that the terminal can make clickable.

PR #25, opened 2022-06-14
Add function to open a test dir, displaying the .json .dot and .osm files on the map

BudgieInWA:
This PR introduces rendering or .osm and .dot files, and a "load test case" mode which layers all the relevant files up.

- [x] Render .osm
- [x] Render .dot
- [x] Load the whole lot from one function call

**Future Work**

- Display a list of tests to click on
- Basic layer toggling ability


tests-web/assets/js/main.js
```diff
@@ -1,22 +1,39 @@
...
     container.ondrop = makeDropHandler(map);
     container.ondragover = handleDragOver;
-    console.info("Ready for drops on the map!", container);
 
     map.loadLink = makeLinkHandler(map);
+
+    console.info("New map created! File drops enabled.", container);
+
+    console.info("opening a test, just for fun...");
+    makeOpenTest(map)("aurora_sausage_link");
```
dabreegster, line 14: Screenshot from 2022-06-21 10-02-59
Cool, it's working well! A few questions:

1) What file should be dropped into the browser window -- the geojson, dot, osm.xml?
2) Is the next idea to get a dropdown or some other way of browsing through test cases?
3) Why are there blue rectangles for the landuse areas -- is that the .osm.xml being parsed and displayed? I assume we'll also eventually have a way to toggle the 3 layers?

tests-web/assets/js/main.js
```diff
@@ -1,22 +1,39 @@
...
     return m;
 }
+
+// Settings for tile layers.
+const LL = L.TileLayer;
+L.TileLayer = function(x, opts = {
+    maxNativeZoom: 18,
+    maxZoom: 21,
+}, ...args) {
+    const l = new LL(x, opts, ...args)
```
dabreegster, line 36: JS style question: any reason to have an intermediate variable and not just return?

tests-web/assets/js/tests.js
```diff
@@ -0,0 +1,24 @@
+import { makeOsmLayer, makeJsonLayer, makeDotLayer  } from './layers.js';
+
+/** Load and display all the files associated with a test case. */
+export const makeOpenTest = (map) => async (name) => {
+    const prefix = `/src/${name}/`;
```
dabreegster, line 5: Answering one of my questions from earlier... looks like the directory should be dragged over

dabreegster (approved):
Super cool to see this come together! Frameworkless JS is also quite simple to follow. And I guess the `unpkg` technique lets you avoid bundlers and build systems.

tests-web/assets/js/main.js
BudgieInWA: There is no functional difference, so there is no reason to introduce the local variable. This is just a bit of mess (I missed the `;` also). **I might put in `prettier` as a dev dependency to autoformat the source.**

This is like my [butter knife on the sink](https://img.ifunny.co/images/8e40ad8758ef62b3bffaff8e0ea716d224096b060124b0c8d2ddea7994594a1d_1.jpg); "maybe I'll use `l` again, but I'm not sure yet".

tests-web/assets/js/main.js
BudgieInWA: 1. Any file that is dropped will be added as a layer (based on it's file extension, if it's supported). This is a convenience for testing, and as a hassle free shortcut to viewing some file in the interface.
2. A list of tests is the intended interface, yeah. See #29.
3. The whole input.xml is being displayed!
	- The parsing of it is all done inline in the vendored `leaflet-osm.js`. We can customise that to our heart's content. (wishlist: style roads and road nodes distinctively, the rest as faint wireframe; crop features or mask layer to bounds specified in the file)
	- Layer controls will be needed to a certain extent, esp. for dropped files and the network graph. Hopefully we can do the right thing with styling to avoid most cases where fiddling with layers is required.

BudgieInWA:
> Super cool to see this come together!

Yes! I was once again tempted to hold this PR back until there was a clickable list of tests, but it's showing something now, so it's time to merge.

> Frameworkless JS is also quite simple to follow. And I guess the `unpkg` technique lets you avoid bundlers and build systems.

It's been a long time coming! This has always been the idea, but the ergenomics of well supported web features has historically been so wonky, that the build system hell-scape was the lesser of two evils. This is the first time I've really tried going vanilla and I'm looking forward to it: this is just the kind of tool to be built lean and simple like that I think.

A bundler might be introduced just so we can switch dialects to `.ts` and minify and stuff.

> A few questions: ...

There are a bunch of specific ideas/plans that aren't written down yet, sprinkled through answers here. I should put them in issues or the README... I will definitely document the user features of the browser in READMEs soon.

Issue #27, opened 2022-06-17
Consider a cargo workspace


dabreegster:
We now have a `Cargo.lock` for the library itself, for the tests, and for the web test browser. [Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) would let us share build artifacts and reduce maintenance burden of updating dependencies in 3 places

BudgieInWA:
Yes! I had seen that, but am not familiar with exactly how it all fits together: packages, crates, mods.

At the moment, the crate `streets` is at the root, and the other crates just happen to be located in the same directory, but otherwise have no relation?

PR #28, opened 2022-06-20
Use cargo workspaces, first moving the main crate into a subdirectory…

dabreegster:
…. #27

1) Move the main crate into a `osm2streets` subdirectory
2) Rename `streets-web` to `osm2streets-web` for consistency (I'm not too opinionated about this, I was just following the `osm2lanes` directory structure)
3) Remove the spare `Cargo.lock` files

`trunk` still works, and so do the unit tests. (Though I noticed you need to run `cargo test --release` while inside the `tests` directory; that's a separate problem with relative paths messing up somehow)


BudgieInWA:
> Rename streets-web to osm2streets-web...

What's might be better called `tests-web`; it's not really coupled to osm2streets (it could test anything, really...). When it does specialise for osm2streets, it will use `osm2streets-npm` as a js dep.

dabreegster:
Done, it's `tests-web` now

BudgieInWA (approved):
Works for me!

PR #30, opened 2022-06-24
Add a clickable list of test cases to the Test Explorer

BudgieInWA:
The list is hard coded, but now all of the test cases can easily be viewed.

### DIFF
--- a/tests-web/assets/js/tests.js
+++ b/tests-web/assets/js/tests.js
@@ -23,7 +23,7 @@ export const makeOpenTest = (map) => async (name) => {
 
 const loadFile = name => fetch(name).then(body => body.text()).catch(err => console.warn(err));
 
-export const loadTests = async ({ open }) => {
+export const loadTests = async () => {
     // FIXME: load the list of tests from the server
     const testNames = [
         'arizona_highways',
@@ -49,8 +49,7 @@ export const loadTests = async ({ open }) => {
     for (const t of testNames) {
         const li = listNode.appendChild(window.document.createElement('li'));
         const a = li.appendChild(window.document.createElement('a'));
-        a.href = '/t/' + t // Any old href makes the browser present the link as clickable.
-        a.onclick = () => open(t);
+        a.href = '/t/' + t;
         a.innerHTML = t;
     }
 }
--- a/tests-web/assets/js/tests.js
+++ b/tests-web/assets/js/tests.js
@@ -18,7 +18,39 @@ export const makeOpenTest = (map) => async (name) => {
     const networkLayer = await makeDotLayer(await network, { bounds })
     map.addLayer(networkLayer);
 
-    console.debug({ rawMapLayer, bounds, inputLayer, networkLayer, map });
+    // TODO store a reference to the layers so they can be cleaned up when wanted.
 }
 
 const loadFile = name => fetch(name).then(body => body.text()).catch(err => console.warn(err));
+
+export const loadTests = async ({ open }) => {
+    // FIXME: load the list of tests from the server
+    const testNames = [
+        'arizona_highways',
+        'aurora_sausage_link',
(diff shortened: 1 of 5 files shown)


tests-web/assets/js/tests.js
```diff
@@ -18,7 +18,38 @@ export const makeOpenTest = (map) => async (name) => {
     const networkLayer = await makeDotLayer(await network, { bounds })
     map.addLayer(networkLayer);
 
-    console.debug({ rawMapLayer, bounds, inputLayer, networkLayer, map });
+    // TODO store a reference to the layers so they can be cleaned up when wanted.
```
dabreegster, line 21: Ah, we have a leak right now? I hit that sort of problem before using leaflet, because I wasn't clear from the docs about who owns the layer once it's added to a map. Working in Rust for so long has made it weird to consider keeping a pointer to a layer even after seemingly "moving" it into the map's ownership.

This shouldn't be hard to fix in any case, just need some kind of current layer bundle `var` declared outside this function, right? The scope and lifetime of a top-level var floating around in some `.js` is another thing about JS I have to get used to.

tests-web/assets/js/tests.js
```diff
@@ -18,7 +18,38 @@ export const makeOpenTest = (map) => async (name) => {
...
+        'taipei',
+        'tempe_light_rail',
+        'tempe_split',
+    ];
+
+    // Add all of the test cases to the list.
+    const listNode = window.document.getElementById('test-list');
+    for (const t of testNames) {
+        const li = listNode.appendChild(window.document.createElement('li'));
+        const a = li.appendChild(window.document.createElement('a'));
```
dabreegster, line 51: Huzzah for not doing the single-page app pattern and doing all this magic to replace part of the page, and then having to make browser history still act normal!

dabreegster (approved):
Awesome stuff! I'm really liking this idea of a UI around data-driven test cases, when the test case data can't be directly understood by a human.

tests-web/assets/js/tests.js
BudgieInWA: It is only a "leak" in the sense that once we add these layers to the map, we never remove them. When we open another test case, we would want to remove the old one (because we don't expect the user to pan the map back to location of the previous test case to view it again). No actual memory is leaked though.

JS being garbage collected, we don't have to worry too much about memory management, but we do have to follow some sort of "ownership" (or at least "responsible guardian") pattern in order to have a program we can reason about.

In this case, we should look at how the leaflet API works to choose how to treat the layers. I think it would be fair to say that we "give the layer to the map" when we add it, and so we should consider that the map owns the layer, but that doesn't have any bearing on the memory -- we don't have to worry about the memory.

We can drop the reference to the layer we created, as we do here, and get it back later using `map.eachLayer(layer => ...)`. But we want to keep track of which layers are which, and call `map.removeLayer(layer)` at some point in the future. The thing to do would be to return the created layers from this function, so that the caller can decide how to keep track of them.

Then we need to make a decision about how to architect that state management, like you speculate. OOP ideas of encapsulation can be really good, but Functional Programming ideas are also very ergonomic (e.g. a top level `var` in a .js file is captured in a closure in order to be used later).

What we have here is not problematic in any way, I just wanted to hint at my upcoming ambitions.

tests-web/assets/js/tests.js
BudgieInWA: Ha ha, sorry to disappoint, but I had planned to do the SPA thing to avoid reloads etc., see #31. That being said, I am a firm believer that the "web page" features should not be discarded to that end!

I like to treat the SPA stuff as progressive enhancement: so use the URL as the source of truth, then intercept the link clicks to transition without reloading, but replicate the browser history behaviour, which is easy to do with the `history` api.

PR #34, opened 2022-06-24
Generate separate Rust test cases at build-time

dabreegster:
Screenshot from 2022-06-24 15-39-17

### DIFF
--- a/tests/build.rs
+++ b/tests/build.rs
@@ -1,0 +1,24 @@
+use std::io::Write;
+
+fn main() {
+    let out_dir = std::env::var("OUT_DIR").unwrap();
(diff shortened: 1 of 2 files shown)


tests/build.rs
```diff
@@ -0,0 +1,24 @@
+use std::io::Write;
+
+fn main() {
```
dabreegster, line 3: I glanced at crates.io/crates/test-case and others, but this was more straightforward

BudgieInWA: I'm glad it turned out this way.

BudgieInWA (approved):
A great step forward in the ergonomics of running the tests!

PR #36, opened 2022-06-27
Use relative paths to load JS and to indicate the test name

### LINKED ISSUE #35: github.io deploy fails to load linked `.js` files (and would fail on `.../t/blah`)
Unlike `trunk serve`, the deployment to GitHub Pages is served as a static site, and at a path below the root (`/osm2streets/`). The JS that is generated by `trunk build` is put in the right place and linked to correctly, but the JS that I have written is included as a static resource.

We need to refer to these resources with the correct relative path: `js/blah.js`.

We cannot use pretend paths like `./t/test_name/` to encode which test is open, because a static site host wont serve `./index.html` for that path (which is where app is, with the logic to parse the path and load the appropriate test). The baseline approach should be to use the query string for the current test name: `./?test=test_name`.

BudgieInWA:
Hopefully closes #35

I have only tested via `trunk serve`, but I'll merge this in in order to test the version deployed to [github.io](https://a-b-street.github.io/osm2streets/)...

### DIFF
--- a/tests-web/assets/js/tests.js
+++ b/tests-web/assets/js/tests.js
@@ -2,7 +2,7 @@ import { makeOsmLayer, makeJsonLayer, makeDotLayer  } from './layers.js';
 
 /** Load and display all the files associated with a test case. */
 export const makeOpenTest = (map) => async (name) => {
-    const prefix = `/src/${name}/`;
+    const prefix = `src/${name}/`;
     const input = loadFile(prefix + 'input.osm');
     const rawMap = loadFile(prefix + 'raw_map.json');
     const network = loadFile(prefix + 'road_network.dot');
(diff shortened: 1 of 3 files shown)


tests-web/assets/js/tests.js
```diff
@@ -49,7 +49,8 @@ export const loadTests = async () => {
     for (const t of testNames) {
         const li = listNode.appendChild(window.document.createElement('li'));
         const a = li.appendChild(window.document.createElement('a'));
-        a.href = '/t/' + t;
+        // Here we encode the test name in the URL to be read elsewhere.
+        a.href = '?test=' + t;
```
BudgieInWA, line 53: It is often nice to implement the encoding/decoding of app state to/from a URL in the same place to hide the details (which makes it easy to retain additional params and that sort of thing). But for now, this is fine.

PR #40, opened 2022-07-03
Merge Lane and Buffer, removing the E indirection

BudgieInWA:
The more I think about it, the difference between a "buffer" and a "lane" is a spectrum, with no cutoff that makes sense in all situations, and a lot of overlapping attributes. I am thinking that a homogeneous representation, with optional sub-structs for variant-specific data makes sense.


osm2streets/src/road_parts.rs
```diff
@@ -39,41 +38,22 @@ pub enum RoadRanks {
...
-    /// Often painted, sometimes w/barriers.
+    /// A part of the road designated for travel.
+    Travel {
+        carriage: Carriage,
+        direction: TrafficDirections,
+    },
+    /// A part of the road designated for parking / "standing".
+    Parking { carriage: Carriage },
+    /// A part of the road that is explicitly not (normally) for carriages.
+    /// E.g. a painted buffer, median or verge.
```
dabreegster, line 49: This subsumes what amenity used to be?

dabreegster (approved):
I forget if we chatted about all the ways of representing variations of objects that all satisfy an interface and share a lot of common data, but indeed this is feeling simpler

osm2streets/src/road_parts.rs
BudgieInWA: Yeah, I don't remember my justification for being more specific than `NoTravel` at this level, and I'm trying not to get too carried away with datatypes that I'm not using yet. In any case, physical features should probably be details *of* `NoTravel`, instead of a distinct variant.

Issue #41, opened 2022-07-05
Moving A/B Street crates to this repository


dabreegster:
@BudgieInWA, I'm thinking about steps to moving `raw_map` and `convert_osm` crates into this repo. Some of the steps would be:

1) Internally split [RawMap](https://github.com/a-b-street/abstreet/blob/master/raw_map/src/lib.rs) into some kind of `StreetNetwork` struct, just with the data/concepts that should live in this repo. Things like buildings, areas, parking lots, the map name, etc would stay in the A/B Street `RawMap`.
2) Internally detangle [convert_osm](https://github.com/a-b-street/abstreet/tree/master/convert_osm/src) into two pieces -- one that should be in this repository, and the A/B Street specific stuff.
3) Split both into separate crates, then move them into this repository.

There would be a dependency loop between the two repos, but I don't think it's a problem, because the crates wouldn't be in a loop. The crates here would only depend on `geom`. The A/B Street side would pull in the osm2streets crates, of course.

The `convert_osm` split will be the more interesting one. There's generic OSM parsing stuff in there. I'll try to keep it as separate as possible -- maybe it's worth a crate of its own, or maybe it's a chance to switch over to something like [osmio](https://github.com/rory/osmio/) and get native PBF reading too.

# Concerns

This will increase my maintenance burden. When this code changes significantly, I regenerate all maps over in A/B Street. To test things temporarily, I'll switch the local dependency in abstreet from this git repo to my local copy. This is a cost that has to be paid eventually, and now's as good a time as any.

But to double check, is this the direction we want to go? We had talked about exploring new schemas and big ideas here and eventually converging, but the current work is building off `RawMap`. I'm eager to try this refactor, but if there's a good chance we won't want it later, no need for churn.

# Questions

1) What should things be called? `RawMap` has always been an awkward choice, just meant to contrast A/B Street's "final" map model. I think we should keep `RawMap` in abstreet (because otherwise there's **lots** of churn renaming files and stuff there) and call the osm2streets bit something else. `StreetNetwork` is my strawman.

2) Most of the data in `RawRoad` and `RawIntersection` looks relevant to move here. The only weird bit is elevation per intersection and percent_incline per road. That's useful in abstreet, but is only populated for a few maps now. OK with keeping the fields in this repo and just not using them much?

# Out of scope for now

[...]

PR #42, opened 2022-07-08
Copy over the street_network and import_street crates from the abstre…

dabreegster:
…et repo. #41

Then change everything in this repo to use them.


osm2streets/src/io.rs
```diff
@@ -21,35 +20,35 @@ use crate::units::{Direction, DrivingSide, Meters, Side, TrafficDirections};
...
 pub fn load_road_network(osm_path: String, timer: &mut Timer) -> RoadNetwork {
-    let driving_side = raw_map::DrivingSide::Right; // TODO
-    let clip = None;
+    let driving_side = street_network::DrivingSide::Right; // TODO
+    let clip_path = None;
 
-    let mut raw_map = convert_osm::convert(
+    let mut street_network = import_streets::osm_to_street_network(
         osm_path.clone(),
-        MapName::new("zz", "osm2streets_test", &abstutil::basename(&osm_path)),
```
dabreegster, line 29: No more weird `MapName` stuff, that's an a/b street concept now gone

dabreegster (commented):
The two test diffs are due to service roads through parking lots now being picked up. Previously they were removed from the street network graph because A/B Street matches them to parking lots and turns them into something else that isn't driveable. We could consider moving that logic over here too, but for now I left it out.

PR #44, opened 2022-07-22
Add a Leaflet control to toggle the 3 layers. Closes #33

### LINKED ISSUE #33: Add layer controls to the Test Browser
When the test browser loads a test, all of the different files are displayed on the map. With adequate styling, many of these layers can be viewed at the same time, but that will often be too much at once.

Buttons should be added to the map to toggle the visibility of loaded layers (or switch between preset combos or something).

dabreegster:
Kind of awesome how this is a one-liner. I don't want to say how many layer-like UI controls I've very painstakingly implemented in A/B Street, or how much code they take...

https://user-images.githubusercontent.com/1664407/180495072-79174d29-c925-4373-9fb6-c8af3a6ca5de.mp4

### DIFF
--- a/tests-web/assets/js/tests.js
+++ b/tests-web/assets/js/tests.js
@@ -18,6 +18,8 @@ export const makeOpenTest = (map) => async (name) => {
     const networkLayer = await makeDotLayer(await network, { bounds })
     map.addLayer(networkLayer);
 
+    L.control.layers({}, {"Geometry": rawMapLayer, "OSM input": inputLayer, "Graph": networkLayer}).addTo(map);
+
     // TODO store a reference to the layers so they can be cleaned up when wanted.
 }
 


dabreegster:
Er, my screencapture chopped off the right edge, but rest assured the standard layer control is there

tests-web/assets/js/tests.js
```diff
@@ -18,6 +18,8 @@ export const makeOpenTest = (map) => async (name) => {
     const networkLayer = await makeDotLayer(await network, { bounds })
     map.addLayer(networkLayer);
 
+    L.control.layers({}, {"Geometry": rawMapLayer, "OSM input": inputLayer, "Graph": networkLayer}).addTo(map);
```
dabreegster, line 21: What're we using for formatting? I used `prettier`, but it changed a bunch of stuff here. VS code had a slight change in the imports. I'm happy to match whatever

BudgieInWA: Prettier is my preferred formatter (as long as trailing commas is turned on). The existing code does not match the `prettier` format, but I am in favour of changing over to it.

For now, format it however looks decent, and I will set up prettier with a config and reformat everything soon.

BudgieInWA:
> Kind of awesome how this is a one-liner. I don't want to say how many layer-like UI controls I've very painstakingly implemented in A/B Street, or how much code they take...

Re-using UI is one of the biggest draws of web-tech. It's nice that leaflet has lots of our use-cases covered :D

Issue #32, opened 2022-06-24
Remove "rust web app" deps from `tests-web/`


BudgieInWA:
I want to turn `tests-web/` into a static vanilla JS web page with as simple an architecture as possible.

I plan to remove all of the rust aspects of the web app, so that `trunk` does nothing except serve the pages.

dabreegster:
No more trunk!

PR #45, opened 2022-07-23
Make the test explorer a vanilla JS app

dabreegster:
#32 

I want to try adding a new feature where you can select an area in Leaflet, query Overpass, pipe through osm2streets, and generate output on-the-fly. To do that, I first need to bring in a JS/WASM API to the Rust crates, and the current yew stuff is in the way / making things confusing. So I went ahead and removed it. We still depend on trunk to build the WASM crate and bundle everything, but similar to the osm2lanes editor WIP, we can replace that later.

All behavior of the test explorer remains the same AFAICT.


osm2streets-js/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+// TODO Placeholder
+
+#[derive(Serialize, Deserialize)]
+pub struct Input {
+    drive_on_right: bool,
+}
+
+#[wasm_bindgen]
+pub fn import_osm(val: &JsValue) -> String {
```
dabreegster, line 12: Next step here will be taking a huge osm.xml blob result from Overpass as input, and returning the graphviz and geometry layers as output. Or... maybe the Overpass query should live in Rust-land like osm2lanes, in which case the input will just be a bounding box. Either way, that's my next step

tests-web/assets/js/main.js
```diff
@@ -2,45 +2,30 @@ import { makeDropHandler, makeLinkHandler, handleDragOver } from "./files.js";
 import { makeOpenTest, loadTests } from "./tests.js";
 
 const useMap = (map) => {
-    const container = map.getContainer();
```
dabreegster, line 5: Part of the diff is me formatting with `prettier`

tests-web/assets/js/main.js
```diff
@@ -2,45 +2,30 @@ import { makeDropHandler, makeLinkHandler, handleDragOver } from "./files.js";
...
-    if (q.has('test')) {
-        const test = q.get('test');
-        console.info("Loading test " + test + " from URL.");
-        map.openTest(test);
-    }
-
-    loadTests();
-}
-
-// Smuggle a reference to the created map, so I can work with it in JS land.
```
dabreegster, line 25: But I did remove `window.maps` and the indirection for leaflet settings. The interaction before between the Rust leaflet crate and this was kind of confusing; I think the direct thing below is equivalent. Let me know if the map state should live elsewhere, or if we should structure this differently to wire up the "load tests" logic

tests-web/index.html
```diff
@@ -1,28 +1,53 @@
...
     <script type="module" src="js/leaflet-osm.js"></script>
     <script type="module" src="js/main.js"></script>
 
     <link data-trunk rel="copy-dir" href="../tests/src" />
-</head>
-
-</html>
\ No newline at end of file
+  </head>
+  <body>
```
dabreegster, line 45: Mostly formatting changes above. Here's the real change -- just taking what the yew app did.

osm2streets-js/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "osm2streets-npm"
```
BudgieInWA, line 2: Is there a reason the package name is different from the directory name?

tests-web/index.html
```diff
@@ -1,28 +1,53 @@
...
+    <script
+      src="https://unpkg.com/@hpcc-js/wasm@1.12.8/dist/index.min.js"
+      crossorigin
+    ></script>
+    <script
+      src="https://unpkg.com/d3-graphviz@4.1.1/build/d3-graphviz.js"
+      crossorigin
+    ></script>
 
+    <link data-trunk rel="rust" href="../osm2streets-js/Cargo.toml" />
```
BudgieInWA, line 38: Is this the "magic" that brings in the JS API to the compiled rust/WASM bits? I'm guessing `osm2streets-js/Cargo.toml` configures the build for the `wasm-bindgen` target, then `trunk` uses this `link` tag to trigger that build, and pick up the correct `.js` output to include in the compiled `.html`?

tests-web/assets/js/main.js
BudgieInWA: Yes, this is appropriate. I had that indirection because I didn't know the extent of the stuff being done in `yew`.

BudgieInWA (approved):
Thanks for making this change! It was tricky for me because I don't know the distinction between `yew` and `trunk` and `cargo`. This PR helps make that clearer in my mind too!

My only request is that you update the first couple of paragraphs in the README Architecture section to make the role of the different pieces clear for a first time reader. Calling out `trunk` as the "bundler" and "dev server" is the main improvement that comes to mind.

That reminds me, does the static build that gets deployed to github.io still work? If so, then `trunk` is doing a a good job bundling :)

osm2streets-js/src/lib.rs
BudgieInWA: My initial impression is that the fetching should be done in JS, because `fetch` and `await` make it very erganomic, and I would want the osm2lanes JS API to accept an osm.xml blob/string as an input for flexibility.

That being said, if there was another function exposed by the JS API that lets me specify a bounding box and osm2lanes does the loading behind the scenes, that would be cool too!

dabreegster: That's the approach I went with in #46. The argument for putting it in Rust would be to share more code if/when we have Java, Python, etc callers. But it's definitely much easier to write this in JS than pulling in `reqwest` dependencies.

osm2streets-js/Cargo.toml
dabreegster: Oops, fixed!

tests-web/index.html
dabreegster: Correct. This causes trunk to call `wasm-pack` on `osm2streets-js`. The methods in that crate labelled `#[wasm_bindgen]` wind up with JS bindings. But as you'll see in the next PR, the way trunk exposes those bindings is not ideal.

dabreegster:
Thanks for the review! Merging, then I'll rebase the next PR.

> My only request is that you update the first couple of paragraphs in the README Architecture section to make the role of the different pieces clear for a first time reader

Done. I also happened to glance at the main project README and noticed it's now out of date too; I'll attempt a pass on that later.

> That reminds me, does the static build that gets deployed to github.io still work? If so, then trunk is doing a a good job bundling :)

I haven't tested end-to-end (we don't have github actions configured to run except on `main`), but I don't expect anything to change. At the end of the day, `trunk` does just emit all the static assets to serve!

osm2streets-js/src/lib.rs
BudgieInWA: Yeah, using overpass servers for data loading sounds more like a decision that the user of the lib should be making (even if we make it super easy).

dabreegster: Right, Overpass isn't the only option, and it won't be appropriate for dynamically loading large areas. Sounds like it's good to keep this at the JS layer.

PR #46, opened 2022-07-24
Import from Overpass on-the-fly

dabreegster:
This is diffbased on #45 -- I'll merge this one second.

The dramatic demo:

https://user-images.githubusercontent.com/1664407/180641266-f8076d9e-c04e-4119-8717-9dbed0a1980c.mp4

It's hard to see the "progress bar" in the button, but basically all of the time is spent waiting on Overpass; the conversion is practically instantaneous!

There are many next steps, but this gets things started. One idea is producing unit test input/output files and making it dead-simple to add a new test case to this repo.


import_streets/src/lib.rs
```diff
@@ -89,9 +89,9 @@ pub enum PrivateOffstreetParking {
     // TODO Based on the number of residents?
 }
 
-/// Create a `StreetNetwork` from an `.osm.xml` file
+/// Create a `StreetNetwork` from the contents of an `.osm.xml` file
 pub fn osm_to_street_network(
-    osm_input_path: String,
+    osm_xml_input: &str,
```
dabreegster, line 94: This layer shouldn't be IO, because that might happen from a filesystem or network

osm2streets-js/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+abstutil = { git = "https://github.com/a-b-street/abstreet" }
+console_error_panic_hook = { version = "0.1.6", optional = true }
+import_streets = { path = "../import_streets" }
+# TODO Upstream this in abstutil crate. WASM is missing some runtime dep otherwise.
```
dabreegster, line 19: Without this, I was hitting https://github.com/rustwasm/wasm-bindgen/issues/2215. I followed the `wasm2wat` workflow in there and figured out this was the problem.

osm2streets-js/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+use serde::{Deserialize, Serialize};
+use wasm_bindgen::prelude::*;
+
+#[derive(Serialize, Deserialize)]
+pub struct Input {
+    driving_side: street_network::DrivingSide,
+}
+
+// TODO Just take a string and bool, maybe remove serde dependency?
```
dabreegster, line 9: This is nowhere near a final API ready for NPM-land, just enough to get the prototype going

tests-web/assets/js/main.js
```diff
@@ -1,46 +1,76 @@
 import { makeDropHandler, makeLinkHandler, handleDragOver } from "./files.js";
 import { makeOpenTest, loadTests } from "./tests.js";
 
+// TODO The hash changes every time the Rust code does, this is very brittle.
```
dabreegster, line 4: This will get annoying very quickly. :P  Recommend any other JS build/bundler tool that can do the same sort of asset copying, call `wasm-pack` (until we have an NPM package and/or for local development)?

tests-web/assets/js/main.js
```diff
@@ -1,46 +1,76 @@
...
+  attribution: "© OpenStreetMap",
+}).addTo(map);
+useMap(map);
+
+// TODO Should we split this into another .js file? How should it get the map if so?
+// TODO Is it OK to just assume the button exists by the time this script runs,
+// or should we create the button here?
+const btn = document.getElementById("import-view");
+btn.onclick = async function importCurrentView() {
+  if (map.getZoom() < 15) {
```
dabreegster, line 42: It'd be nicer to use a real selection widget and have some feedback about when the area selected is too big.

tests-web/assets/js/main.js
```diff
@@ -1,46 +1,76 @@
...
+
+  // Grab OSM XML from Overpass
+  // (Sadly toBBoxString doesn't seem to match the order for Overpass)
+  const b = map.getBounds();
+  const bbox = `${b.getSouth()},${b.getWest()},${b.getNorth()},${b.getEast()}`;
+  const query = `(nwr(${bbox}); node(w)->.x; <;); out meta;`;
+  const url = `https://overpass-api.de/api/interpreter?data=${query}`;
+  console.log(`Fetching from overpass: ${url}`);
+
+  btn.innerText = "Downloading from Overpass...";
```
dabreegster, line 54: Cheap way to do progress bars!

BudgieInWA: This feels like the right place to surface this state in the UI too, because it's async.

tests-web/assets/js/main.js
```diff
@@ -1,46 +1,76 @@
...
+L.tileLayer("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
+  maxNativeZoom: 18,
+  maxZoom: 21,
+  attribution: "© OpenStreetMap",
+}).addTo(map);
+useMap(map);
+
+// TODO Should we split this into another .js file? How should it get the map if so?
+// TODO Is it OK to just assume the button exists by the time this script runs,
+// or should we create the button here?
```
BudgieInWA, line 39: Yes, it should go into it's own file, and have some initialisation function that accepts the map that it should attach to, which is what I've done with the `makeX` functions. That pattern is probably not the best fit, I am thinking of replacing each of them with a class instead (because the "class" metaphor is exactly what we are trying to do).

The button can either be part of the initial HTML, in which case we can assume it exists already. Alternatively, the initialisation of this feature could create the button (and append it to a "toolbar" element), which encapsulates the feature nicely.

Don't worry about the overall structure yet, that is something that I have been working through in my head, that I intend to tackle soon.

BudgieInWA (changes requested):
What a great prototype!

The linking to the wasm is too fragile to merge this in just yet I think. I suggest trying to stop using `trunk` entirely, try `wasm-pack` plus symlinks or relative `../` links directly in the HTML source, so that `tests-web` can be served directly (with `python3 -m http.server` or your favourite alternative).

tests-web/assets/js/main.js
dabreegster: Done

osm2streets-js/src/lib.rs
dabreegster: Fixed the casing, and good point -- the options struct will come in handy.

tests-web/assets/js/main.js
dabreegster: I'll look into this next. There's got to be a good workflow someone's come up with at this point that doesn't bring in an entire framework.

The temporary hack might be a small shell script that calls `wasm-pack` and copies some assets around. This'd be called manually during local dev, and also in the github workflow. Would that be an OK temporary workaround? Or maybe the symlinks are cleaner... I'll see if `wasm-pack` produces consistently named output files.

> That might be worth while in the future, in order to gain access to npm dependencies (without using unpkg), typescript and .less or .scss support, etc.

I actually already hit a real use case for that. This morning, `unpkg` was slow for a bit. And yesterday I didn't have internet on a train, and couldn't develop. Having local caches of all external dependencies and hosting everything ourselves on github pages feels preferable to me.

BudgieInWA: I think a shell script that calls `wasm-pack` and copies files around is a perfect solution. And having a local copy of the JS deps is a good idea too., so let's make a `vendor` dir and store a copy of the JS assets we need in there.

dabreegster: I guess you've done that for leaflet and it's been working nicely. Literally should I just `wget` things manually, or does npm/yarn/something help do things like this?

dabreegster: OK @BudgieInWA, PTAL. No more trunk. The local dev script literally just calls `wasm-pack` to build the Rust crates to JS/WASM, then runs a Python server. The rest is handled by symlinks. It's nice and simple, works great locally for me. The github action may need a bit more work (I'm unclear if the publish action chases symlinks). I can iterate on that after merging this if needed.

We can vendor external JS deps into this repo as a followup for more performance + less points of failure on other CDNs.

BudgieInWA (approved):
Looks good and runs well on my end!

I have pushed a commit that renames some directories, so check that you are happy with that.

dabreegster:
Awesome, thanks for the review and the `www` move. I'll merge this and poke at github actions in case the symlinks cause problems.

PR #47, opened 2022-07-28
Add "IntersectionComplexity" classification to intersections

BudgieInWA:
This PR renames `intersection_type` to `control` and introduces `complexity` to represent the kind of traffic interactions that occur at the intersection. The primary reason is to specialise the implementation of intersection geometry algorithms, rendering algorithms, and other transforms.

`IntersectionComplexity` is based on `streets::road_functions::IntersectionType` but renamed. Like every rename, I feel like this iteration is an improvement, but what do you think of the names and docstrings, now that this is in the context of the A/B Streets code? 

- [ ] Make use of the new classification in `streets::io`.
- [ ] Use `.complexity` `MapEdge` instead of `.control` `Border`, remove `ControlType::Border`.

I added a quick-and-dirty colouring of intersections based on complexity:

image


street_network/src/types.rs
```diff
@@ -124,12 +124,64 @@ pub enum DrivingSide {
...
-    TrafficSignal,
-    Border,
-    Construction,
+pub enum InterruptionType {
+    Uninterrupted,
+    Yield,
+    Stop,
+    Signal,
+    DeadEnd,
+}
```
BudgieInWA, line 135: This is not used yet, so I will probably delete it for this PR. Both end of every road should be classified with one of these, I think.

street_network/src/transform/classify_intersections.rs
```diff
@@ -0,0 +1,46 @@
+use crate::osm::NodeID;
+use crate::IntersectionComplexity::*;
+use crate::{IntersectionComplexity, OriginalRoad, RawIntersection, StreetNetwork};
+
+/// Determines the initial complexity of all intersections. Intersections marked "Crossing" are
+/// considered "unclassified" and will be updated with a guess, others will be left unchanged.
+pub fn classify_intersections(streets: &mut StreetNetwork) {
+    let mut changes: Vec<(NodeID, IntersectionComplexity)> = Vec::new();
+    for (id, inter) in streets.intersections.iter() {
```
dabreegster, line 9: Style nit
```suggestion
    for (id, inter) in &streets.intersections {
```

The separate loops to guess and then assign are unfortunately necessary due to `streets.roads_per_intersection`, as I'm sure the borrow checker let you notice. ;)  Even more motivation for me to finish refactoring and storing the set of roads on `RawIntersection` directly.

street_network/src/transform/classify_intersections.rs
```diff
@@ -0,0 +1,46 @@
...
+        if let Crossing = inter.complexity {
+            changes.push((
+                *id,
+                guess_complexity(inter, streets.roads_per_intersection(*id)),
+            ));
+        }
+    }
+
+    for (id, complexity) in changes.into_iter() {
+        if let Some(inter) = streets.intersections.get_mut(&id) {
```
dabreegster, line 19:
```suggestion
        streets.intersections.get_mut(&id).unwrap().complexity = complexity;
```
Slightly more succinct, and the unwrap is definitely safe because no IDs suddenly disappeared

street_network/src/transform/classify_intersections.rs
```diff
@@ -0,0 +1,46 @@
...
+    if roads.len() == 1 {
+        return Terminus;
+    }
+
+    // A Connection is characterised by exactly two connected roads.
+    if roads.len() == 2 {
+        return Connection;
+    }
+
+    // A MultiConnection is characterised by exactly two connected dividing lines (lines that
```
dabreegster, line 39: I think I understand the distinction between all of these finally :)

street_network/src/types.rs
dabreegster: I like this. End of every _lane_ though, right?

street_network/src/types.rs
```diff
@@ -124,12 +124,64 @@ pub enum DrivingSide {
...
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub enum IntersectionComplexity {
+    /// The edge of the data that we have.
```
dabreegster, line 139:
```suggestion
    /// The edge of the data that we have.
    ///
    /// The road may continue past here in reality and do anything, but the current dataset is clipped here.
```

street_network/src/types.rs
```diff
@@ -124,12 +124,64 @@ pub enum DrivingSide {
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub enum ControlType {
+    Uncontrolled, // Pretty sure this is a term that implies right of way rules somewhere.
+    //TODO YieldSign,
+    StopSign,      // Signed is a good standard of safety
+    TrafficSignal, // Signalled is better.
+    Border,        //TODO move to using IntersectionComplexity::MapEdge
+    Construction,  // Are these treated as "closed"?
```
dabreegster, line 184: Yes. Sometimes in-progress or existing roads are marked as under construction in OSM, so we want to close them down to all traffic, but still represent them

street_network/src/types.rs
```diff
@@ -124,12 +124,64 @@ pub enum DrivingSide {
...
+
+    /// The end of the line.
+    ///
+    /// Turning circles, road end signs, train terminus thingos, ...
+    Terminus,
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub enum ControlType {
+    Uncontrolled, // Pretty sure this is a term that implies right of way rules somewhere.
```
dabreegster, line 179: This is a good thing to introduce. I've been abusing `StopSign` to imply this, by disabling the stop sign on every incoming road, and coding in rules at the traffic sim layer to mimic some kind of yielding.

tests-web/www/js/layers.js
```diff
@@ -1,7 +1,27 @@
...
+    MultiConnection: '#669',
+    Merge: '#969',
+    Crossing: '#999',
+    Terminus: '#966',
+}
+
+const styleGeoJson = (feature) => {
+    if (feature.geometry.type === 'Polygon') {
+        return {
+            color: intersectionColours[feature.properties?.complexity] || '#666',
```
dabreegster, line 17: You can do `?` in the middle there and have `||` at the end? O_O   The `[]` returns null, so the map lookup returns null. Very elegant, much more readable than Rust-style `features.properties.get("complexity").and_then(|c| intersectionColors.get(c)).unwrap_or_else("#666")`

dabreegster (approved):
This is an excellent step forward! Few style nitpicks, otherwise ready to merge. The two TODOs you marked could be done here or separately, up to you.

> what do you think of the names and docstrings, now that this is in the context of the A/B Streets code?

All sound good. I'm understanding more now some strategies for iterating more quickly here and not needing to make sure everything existing remains working. I'll take care of any updates in the other code-base asynchronously, by only upgrading the crate dependencies in batches.

street_network/src/transform/classify_intersections.rs
BudgieInWA: And our borrow of `streets` guarantees that! I had it that way first, and changed it because i forgot that the borrow checked has my back here.

tests-web/www/js/layers.js
BudgieInWA: `?.` is actually the operator, which returns `undefined` if the LHS is `null` or `undefined` (instead of crashing), otherwise acts like `.`. It is guarding against `properties` not being there (thought it probably always will)

`x[undefined]` is coerced to `x['undefined']` (because JS) which will return `undefined` in our case, so we can use `||` to chain to the default value. Chaining with `||`, `&&`, and now `??` and `???` can be quite elegant.

street_network/src/transform/classify_intersections.rs
BudgieInWA: It certainly did!

I did a lot of back and forth writing my street network impl on where to store the connections between roads and intersections in rust. Making references work out seems not worth it (due to lifetimes and the doubly linked nature of roads and intersections). The other option was using ids and getting roads and intersections out of a map every time, which actually meshes pretty well with some "data oriented" ideas that I'm vaguely familiar with, especially if the ids could be an index.

Working with ids, my favourite solution was to use a proper graph impl to store the source of truth for connectivity. It is the right abstraction and gives us all the `roads_per_intersection` type functionality that we could need. Its a bit wrinkly, but an undirected multigraph with edges around a node ordered in CW order, should cover us i think.

I'm interested to see where we go with this.

street_network/src/types.rs
BudgieInWA: Yes. Every *manoeuvre* I suppose. Lanes divvy themselves up, by tags and defaults.

I mean, any time you're turning across traffic, you pull up and give way at some point. Every driveway is a give way for any traffic doing that turn manoeuvre (but private driveways tend to be ignored, unless you see something). So the same lane is both uninterrupted and sometimes give way. (if the person in front of you is giving way, so are you, as long as you stay in the lane). Traffic lights are also uninterrupted and sometimes stop, I guess.

Here in Aus, we have double white lines, for no overtaking on long highways. You can drive straight ahead as long as you like, without and conflict, or turn into a parking bay or side road. Because you can't cross double white lines, every driveway they you're allowed to use across traffic has a small gap in the doubles, to show that you're allowed across. I'll see if I can find a picture somewhere.

BudgieInWA: [Proper gap at a Yielding road](https://www.mapillary.com/app/?lat=-34.970524391611&lng=116.94807567397&z=17&pKey=3894113797324472&focus=photo) then a [tiny one at a driveway](https://www.mapillary.com/app/?lat=-34.9700375617&lng=116.94865674417&z=17&pKey=688989171843342&focus=photo) just up the road.

street_network/src/transform/classify_intersections.rs
```diff
@@ -0,0 +1,46 @@
...
+                *id,
+                guess_complexity(inter, streets.roads_per_intersection(*id)),
+            ));
+        }
+    }
+
+    for (id, complexity) in changes.into_iter() {
+        if let Some(inter) = streets.intersections.get_mut(&id) {
+            inter.complexity = complexity;
+        }
```
BudgieInWA, line 21:
```suggestion
        streets.intersections.get_mut(&id).unwrap().complexity = complexity;
```
PS @dabreegster, you can drag the blue <kbd>+</kbd> down over a few lines, for suggesting to replace multiple lines. Also helps include context in the conversation tab.

tests-web/www/js/layers.js
```diff
@@ -1,7 +1,28 @@
...
 
+const intersectionColours = {
+    undefined: '#666', // for default tarmac
+    Connection: '#666',
+    MultiConnection: '#669',
+    Merge: '#969',
+    Crossing: '#966',
+    Terminus: '#999',
+    MapEdge: '#696',
+}
```
BudgieInWA, line 13: image

Here, non-intersections get the value for the string key `'undefined'`, because they have no `complexity`

BudgieInWA:
Thanks for the style nitpicks, they are really helping me figure out *how* to think about some technical aspects of the language.

I'll leave the TODOs for later. Remove `ControlType::Border` is probably easier as part of the work integrating this into A/B Street.

Issue #22, opened 2022-06-08
Create a slippy map "Tests Browser"


BudgieInWA:
This repo contains an independent test suite that tests`RoadNetwork` and abstreet's `RawMap`. Each test case is a real world example with a lot of geographic data.

We should build a slippy map that shows the various formats of data for a test case, layering the .osm with the .geojson and linking the .dot. A simple menu would let you choose a test case to view, and highlight tests that need approving.

It could then be extended to run tests on demand, downloading .osm on the fly.

BudgieInWA:
[osm2lanes](https://github.com/a-b-street/osm2lanes) probably has exactly the stack that would achieve this. It would also benefit from the same tool. Maybe osm2lanes tests end up here too?

BudgieInWA:
- GeoJSON is built in: https://leafletjs.com/reference.html#geojson
- openstreetmap/leaflet-osm has [a straightforward OSM `DataLayer`](https://github.com/openstreetmap/leaflet-osm/blob/master/leaflet-osm.js#L73)
- DOT with something like https://github.com/magjac/d3-graphviz#d3-graphviz ?

BudgieInWA:
Before calling this issue done, just need to

- [ ] present a list of available tests to click on, and
- [ ] link from the README

dabreegster:
Hijacking this slightly to write down some ideas I'm working on implementing. Here are some desired features of the StreetExplorer:

- click a test case and see its 3 renderings / layers quickly
- click a button to reimport a test case's built-in .osm. This makes a convenient dev workflow to work on transformation / importing code and use the browser for interactive debugging.
- for any imported area, see a plain geometric GeoJSON view, or a more detailed rendering with the start of lane/intersection markings
- import a custom area from Overpass, then save it as .osm for a new test case input
- when importing a new area or reimporting an existing one, modify the transformation options. Optionally enable a debug mode, where we get back multiple copies of the `StreetNetwork` after each transformation step. Be able to browse forwards/backwards through the results as layers.

I'm imagining the UI state is organized around a "current loaded thing." That has a test name, or it's a freshly imported bit. There's a list of leaflet layers associated with it (so we can clean them up when we switch.) Then there are extra UI controls, depending on what's loaded. For a test case, there's a "reimport" button with options for transformation options. For a newly imported bit, there's a button to save the `.osm` for a new test case.

BudgieInWA:
[...]

PR #49, opened 2022-07-30
Start rendering more detail and add a "reimport" feature

dabreegster:
This PR is a bit all over the place -- it does a few things:

- make the WASM interface nicer. After importing a `StreetNetwork`, we get back a javascript object and can then call different methods on it.
- Start a "more detailed" render mode for #24. Right now it just adds dashed lines between every lane as an example. We can explore various visual styles and output formats (geojson, SVG, whether to do the "hard work" of dashing lines or placing icons in Rust or leveraging SVG, etc).
- Add a "reimport" button to each test case, to pipe the `.osm` through the WASM API and generate geometry again.

A demo of some of this:

https://user-images.githubusercontent.com/1664407/181916860-477566b0-7dd4-4a0c-b5ca-688088fe5238.mp4

What I'm trying to build towards is a list of features: https://github.com/a-b-street/osm2streets/issues/22#issuecomment-1200152718
I think my next step will be to organize all of the app state in a class -- the leaflet map, the layer control, the list of layers, and what's currently happening (a test or newly imported area). That'll let us sanely manage layers and keep the layer widget up to date, I think.

There's lots in this PR and it leaves the app a bit more messy; feel free to pushback and I can tease this into smaller patches.


tests-web/www/js/layers.js
```diff
@@ -1,27 +1,63 @@
...
-                if (!svg) console.error('no svg element came about from the render')
-                resolve(new L.svgOverlay(svg, bounds, {
-                    opacity: 0.3,
-                    interactive: true
-                }));
-            })
-            .dot(text).render();
-    })
-}
+  return new Promise((resolve, reject) => {
```
dabreegster, line 38: Some of the diff here is `prettier`

tests-web/www/js/tests.js
```diff
@@ -1,58 +1,100 @@
-import { makeOsmLayer, makeJsonLayer, makeDotLayer  } from './layers.js';
+import {
+  makeOsmLayer,
+  makePlainGeoJsonLayer,
+  makeDetailedGeoJsonLayer,
+  makeDotLayer,
+} from "./layers.js";
+import { JsStreetNetwork } from "./osm2streets-js/osm2streets_js.js";
+
+var currentLayers = [];
```
dabreegster, line 9: I started a little bit of layer management here, but it doesn't work across everything. For example, we only create the layers widget after clicking on a test case.

tests-web/www/js/tests.js
```diff
@@ -1,58 +1,100 @@
...
+  // Add all of the test cases to the list.
+  const listNode = window.document.getElementById("test-list");
+  for (const t of testNames) {
+    const li = listNode.appendChild(window.document.createElement("li"));
+    const a = li.appendChild(window.document.createElement("a"));
+    // Here we encode the test name in the URL to be read elsewhere.
+    a.href = "?test=" + t;
+    a.innerHTML = t;
+
+    const reimport = li.appendChild(window.document.createElement("button"));
```
dabreegster, line 75: The button style is pretty bad. I think the better UI is probably clicking a test to open it, then getting a context-sensitive set of actions that knows a test is open and has a button to reimport.

dabreegster:
The merge conflict with #47 should be minimal -- feel free to merge that one first when it's ready

tests-web/www/js/tests.js
BudgieInWA: I agree, that is how I am imagining the UI evolving too.

tests-web/www/js/import.js
```diff
@@ -27,13 +28,14 @@ export const makeImportCurrentView = (map, btn) => {
...
-      const output = importOsm(osmXML, {
+      const network = new JsStreetNetwork(osmXML, {
         // TODO Ask overpass
         driving_side: "Right",
       });
 
       // TODO Definitely time to think about cleaning up old layers
-      L.geoJSON(JSON.parse(output), { style: { color: "#f55" } }).addTo(map);
+      makePlainGeoJsonLayer(network.toGeojsonPlain()).addTo(map);
+      makeDetailedGeoJsonLayer(network.toGeojsonDetailed()).addTo(map);
```
BudgieInWA, line 38: We should consider providing mutually exclusive layers that can be used in conjunction, I think. Seems like a useful API going forward, as the different kinds of geom will be used in different ways. We can always combine the layers using a leaflet `LayerGroup` as needed for ease of use within the web app.

I was imagining "road area", "road markings" and "debug" (for whatever ad-hoc geom we want to render in garish colours during dev) as three initial outputs.

BudgieInWA (approved):
I like this PR. While it is a bit all over the place, each change is something that I have already wanted, or otherwise seems natural, so I am able to make sense of it all!

The JS is currently structured awkwardly (as we've discussed), so let's add these changes into the "mess" before we refactor; they will help inform the refactor.

tests-web/www/js/import.js
dabreegster: Yeah, that makes more sense than having two mutually exclusive "renderings", one of which has a bunch of labeled layers anyway. I'll keep it this way for this PR and we can iterate further.

One of the things stopping us from lifting more of A/B Street's lane markings is the lack of turning/movement info. We can discuss that too shortly...

dabreegster:
Alright, I'll merge this in before we start the JS overhaul

Issue #16, opened 2022-05-23
Handle the simplest sausage link case


dabreegster:
One of my LTN collaborations actually has some pesky geometry issues that look simple to deal with. 
![Screenshot from 2022-05-23 15-49-13](https://user-images.githubusercontent.com/1664407/169846397-d83341b0-aee0-478f-a8ef-c64f386f2de5.png)
https://www.openstreetmap.org/way/291394487#map=19/51.45946/-2.55068

The dual carriageway exists for a very small space due to a crossing island. In the short term, simplifying this to a regular bidirectional road with nothing special in the middle is appropriate. (I would love to figure out how to properly model crossing islands like this, but that's a later step.)

These look straightforward to detect and transform. I'll only try to handle these loops, where road1 and road2 share both endpoints. This means there can't be any other roads connecting to one side or the other.

CC @martinfleis

dabreegster:
Part of this will be fixing up the lanes inferred on each piece, because it's definitely wrong right now. It's possible I might block working on this on cutting over to osm2lanes finally, so any fixes can happen in one (better tested) place

BudgieInWA:
My attempt at something like this in the [JOSM Lanes plugin](https://github.com/BjornRasmussen/Lanes/pull/8) was looking at one merged intersection at a time, but not at any graph properties. I am exploring that idea further, so it would be cool to see any ideas that other people have about it. Looking at the graph might be very powerful, for instances there's a little loop of oneways in the graph for each of the sausages.

It will be interesting for me to see *you* tackle a transformation in `RawMap` as a sort of template for how I might do it.

BudgieInWA:
> (I would love to figure out how to properly model crossing islands like this, but that's a later step.)

(I think: the presence of a traffic island implies that a short stretch of the divider has a substantial width. i.e. `traffic_calming=island` or `crossing:island=yes` implies a median buffer (which I don't quite know how to tag yet).)

dabreegster:
> It will be interesting for me to see you tackle a transformation in RawMap as a sort of template for how I might do it.

https://github.com/a-b-street/abstreet/blob/5907a8e3035ff7ae4af43568e5ad8d651d2fe400/raw_map/src/transform/find_short_roads.rs#L116 is one half of an example. (The other half is `merge_short_road`, which doesn't handle all edge cases)

> (I think: the presence of a traffic island implies that a short stretch of the divider has a substantial width. 

[...]

PR #50, opened 2022-07-31
Detect and collapse sausage links. #16

dabreegster:
This is modernizing https://github.com/a-b-street/abstreet/pull/933. I added some more checks to avoid incorrectly detecting sausage links, adding a few regression tests. I'm also inserting a buffer lane in the middle, to show that there's some kind of physical barrier. (We can get fancier with that in the future -- having semantics for concrete dividers, pedestrian islands, getting a more realistic shape, etc.)

I'm reimporting all maps in A/B Street and using some visual diff tools there to check things. If any bugs creep up, I'll add a regression test here and iterate. But I think I've got the logic correct now.

Some before/after:
Screenshot from 2022-07-31 18-07-36
Screenshot from 2022-07-31 18-07-41

Screenshot from 2022-07-31 18-06-36
Screenshot from 2022-07-31 18-06-41

Screenshot from 2022-07-31 18-06-10
Screenshot from 2022-07-31 18-06-15


street_network/src/transform/sausage_links.rs
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        let road2 = &streets.roads[&id2];
+        if road2.oneway_for_driving().is_none()
+            || road1.osm_tags.get(osm::NAME) != road2.osm_tags.get(osm::NAME)
+        {
+            continue;
+        }
+
+        // The two roads must point in a loop. Since they're both one-way, we can just
```
dabreegster, line 51: Restoring @BudgieInWA's comments from https://github.com/a-b-street/abstreet/pull/933...

> Your "bad example" looks like a coherent collapsing of that topology, but in the wrong situation. It should be ok to represent a looping parking aisle as one way like that. The topology is the same as a sausage link, so we need to use more clues to classify it as one.
> The geometry is I think what we need to use. What if we linearly interpolate along both sides of the sausage, and take the distance between the two resulting points. We would expect all the values to be close for a dual carriage way, and get huge for loop roads.

Fortunately we didn't need to look at geometry to fix these cases. Just the graph structure was enough to tell. But the idea that both sides of a dual carriageway should stay roughly the same distance away from each other is a great one; I think it'll be necessary to handle cases like https://dabreegster.github.io/talks/asg_osm2streets/slides.html#/osm2streets-dual-carriageways-1.

dabreegster:
Overall it mostly worked well. I spotted some "false positives" where it collapsed and shouldn't've:
- https://www.openstreetmap.org/node/8141904208#map=19/51.49211/0.00378
- https://www.openstreetmap.org/node/9563439140#map=19/51.49990/0.00457
- https://www.openstreetmap.org/node/90266738#map=19/51.46112/0.03800
- https://www.openstreetmap.org/node/90407230#map=19/51.47857/0.05173

I'm going to merge and follow up with fixing those cases later. There's a bit of a sunk cost once I regenerate all of the A/B Street data, unfortunately. When it mostly improves things, it's hard to justify perfecting things first.

PR #51, opened 2022-08-01
Refactor JS app

dabreegster:
Alright, I finished restoring all the functionality after our pair programming. There are some quirks, but I think this is a good step forwards. Comments/questions will be added momentarily...


tests-web/www/index.html
```diff
@@ -36,13 +36,17 @@
...
+      import { StreetExplorer } from "./js/main.js";
+      window.app = await StreetExplorer.create(document.getElementById("map"));
+    </script>
   </head>
   <body>
     <menu>
       <h1>StreetExplorer</h1>
       <button type="button" id="import-view">Import current view</button>
       <ul id="test-list" />
+      <ul id="view-controls" />
```
dabreegster, line 49: Screenshot from 2022-08-01 12-32-11
Why isn't this below the `test-list`? And even though I'm adding new child elements to it, it smushes the buttons together. CSS styling problem of some sort?

tests-web/www/js/files.js
```diff
@@ -52,3 +52,20 @@ const forEachFile = (dt, f) => {
     }
   }
 };
+
+export const loadFile = (name) =>
+  fetch(name)
+    .then((body) => body.text())
+    .catch((err) => console.warn(err));
+
+export function downloadGeneratedFile(filename, textInput) {
```
dabreegster, line 61: I lifted this from some stackoverflow page. FWIW, I've seen this not work when passing bigger (50MB?) URL-encoded binary files from Rust. But seems fine for our small xml files for now.

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+  makePlainGeoJsonLayer,
+  makeDetailedGeoJsonLayer,
+  makeDotLayer,
+} from "./layers.js";
+import init, { JsStreetNetwork } from "./osm2streets-js/osm2streets_js.js";
 
+await init();
+
+export class StreetExplorer {
+  constructor(mapContainer) {
```
dabreegster, line 20: Private constructor, have to use `create` so it can do async

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+
+      await app.setCurrentTest((app) =>
+        TestCase.importCurrentView(app, importButton)
+      );
+    };
+
+    return app;
+  }
+
+  addLayer(name, layer) {
```
dabreegster, line 56: I sort of feel like this is a Leaflet API problem. It should be easier to add/remove a named layer to a map and the overlay control simultaneously.

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+
+  removeLayer(layer) {
+    this.map.removeLayer(layer);
+    this.layerControl.removeLayer(layer);
+  }
+
+  async setCurrentTest(testMaker) {
+    this.currentTest?.cleanup();
+    this.currentTest = await testMaker(this);
+    this.map.fitBounds(this.currentTest.bounds, { animate: false });
```
dabreegster, line 70: I'm rethinking the error handling now... maybe `testMaker` should be allowed to throw an exception. We catch it here, just clear out the old layer but do nothing else. Would that be cleaner?

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+      // TODO Ask overpass
+      const drivingSide = "Right";
+
+      const network = new JsStreetNetwork(osmInput, {
+        driving_side: drivingSide,
+      });
+      const rawMapLayer = makePlainGeoJsonLayer(network.toGeojsonPlain());
+      const bounds = rawMapLayer.getBounds();
+
+      var layers = [];
```
dabreegster, line 147: There's some duplicate code with the reimport case below. Maybe we could have a helper that fills out layers given a `JsStreetNetwork`? But there are some differences.. we want to name the layers with "(reimport)" in one case. And in this case, we don't have the `TestCase` created yet, but we do in the reimport case

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+      // Make the button clickable again
+      importButton.innerText = "Import current view";
+      importButton.disabled = false;
+    }
+  }
+
+  renderControls(container) {
+    container.innerHTML = "";
+    if (this.name) {
+      const title = container.appendChild(document.createElement("b"));
```
dabreegster, line 171: The visual output here is weird; I was expecting each child of the `ul` to wind up on its own line.

Also this feels like a clunky API for creating new elements like this. Any suggested improvements without reaching for a huge framework?

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+    attribution: "© OpenStreetMap",
+  }).addTo(map);
+  // Geocoder, satellite layers, etc
+  return map;
+}
+
+// TODO Port stuff
 const useMap = (map) => {
   const container = map.getContainer();
   container.ondrop = makeDropHandler(map);
```
dabreegster, line 230: I haven't fixed this up yet. I don't think I've used this before -- you could drag and drop a `.osm` or `.json` or `.dot` file into the window and have it add a layer, right? How should that work with the `currentTest` paradigm -- maybe we try to create a new test case? But we won't always have the OSM XML as input, and we need that for every `TestCase`. So maybe this just blindly adds a layer to the current test?

tests-web/www/js/main.js
```diff
@@ -1,7 +1,230 @@
...
+      var layers = [];
+      layers.push(app.addLayer("Geometry", rawMapLayer));
+      layers.push(
+        app.addLayer(
+          "Detailed geometry",
+          makeDetailedGeoJsonLayer(network.toGeojsonDetailed())
+        )
+      );
+      layers.push(app.addLayer("OSM", makeOsmLayer(osmInput)));
+      // TODO The dot layer
```
dabreegster, line 156: We can add another method to `osm2streets-js` to get this going

dabreegster:
I probably need to shift focus for a day or two; feel free to take over the PR if you have time, or I'll finish it up Wednesdayish

dabreegster:
Alright, I think this is good enough to merge. Still not working:
- the layouting is funky; there are some big things about CSS I'm not understanding yet
- graphviz reimport is breaking with an error about the `graph` var not being defined, disabled for now
- drag and drop behavior not ported over to the new structure

I'll likely merge tomorrow (and welcome async feedback) and then get the next step out there -- something to emit multiple debug layers from within a transformation, and browse through. I hacked something together in another branch to work on some dual carriageways, but I'd like to clean that up and get another PR out.

BudgieInWA:
>     * the layouting is funky; there are some big things about CSS I'm not understanding yet

Yeah, css tends to end up hyper specific to the exact html, so I just need to look at it and lay it out again.

>     * graphviz reimport is breaking with an error about the `graph` var not being defined, disabled for now

Disabling the graphviz layers is preferable anyway, at the moment.

>     * drag and drop behavior not ported over to the new structure

It was mostly useful for debugging the map, otherwise its just a useful back door. Keep the code that catches the file, then we'll wire it into the map at some point.

> I'll likely merge tomorrow (and welcome async feedback) and then get the next step out there -- something to emit multiple debug layers from within a transformation, and browse through. I hacked something together in another branch to work on some dual carriageways, but I'd like to clean that up and get another PR out.

I'll see what I come up with for CSS and any tweaks I see before you get to the merge button :)

dabreegster:
By the urgency vested in me by #52, I'm going to merge this now and figure out the CSS later. Sorry. :\   Next PR will flesh out an interactive debug mode, making the layer management more interesting.

> It was mostly useful for debugging the map, otherwise its just a useful back door. Keep the code that catches the file, then we'll wire it into the map at some point.

Sounds good!

PR #53, opened 2022-08-08
Optionally show intermediate transformation steps and debug data

dabreegster:
https://user-images.githubusercontent.com/1664407/183407689-c250bbc1-3cff-4f45-8f26-2f95160d7822.mp4

There's now a setting for imports. When enabled, we take a full copy of the `StreetNetwork` in between each transformation, and add it as a layer. Within transformations, we can also label points or lines for more explanation or debugging.

The way it works is a bit clunky, but usable. My next steps:

1) add the first part of a complicated dual carriageway algorithm, where I make extensive use of the debugging
2) work on some controls for managing all the layers. I think we want groups: original / built-in test data; the final reimport; each step of a transformation. For the last, arrow controls to scroll between them will be vital.

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -33,11 +33,13 @@ impl JsStreetNetwork {
             &mut timer,
         )
         .map_err(|err| JsValue::from_str(&err.to_string()))?;
-        street_network.apply_transformations(
-            // TODO Assuming defaults here; probably do take in Input
-            street_network::Transformation::standard_for_clipped_areas(),
-            &mut timer,
-        );
+        // TODO Assuming defaults here; probably do take in Input
+        let transformations = street_network::Transformation::standard_for_clipped_areas();
+        if input.debug_each_step {
+            street_network.apply_transformations_stepwise_debugging(transformations, &mut timer);
+        } else {
+            street_network.apply_transformations(transformations, &mut timer);
+        }
 
         Ok(Self {
             inner: street_network,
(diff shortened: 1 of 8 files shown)


osm2streets-js/src/lib.rs
```diff
@@ -60,4 +63,41 @@ impl JsStreetNetwork {
         let road_network: streets::RoadNetwork = self.inner.clone().into();
         road_network.to_dot()
     }
+
+    #[wasm_bindgen(js_name = getDebugSteps)]
+    pub fn get_debug_steps(&self) -> Vec<JsValue> {
+        // TODO Figure out how to borrow from the RefCell instead of cloning
```
dabreegster, line 69: The perf of all of this is pretty clumsy... cloning everywhere. I need to read up on lifetimes and how borrowing works with WASM. For our small test areas, it doesn't matter at all, even when compiling in debug mode though

osm2streets-js/src/lib.rs
```diff
@@ -60,4 +63,41 @@ impl JsStreetNetwork {
...
+    }
+}
+
+#[wasm_bindgen]
+pub struct JsDebugStreets {
+    inner: street_network::DebugStreets,
+}
+
+#[wasm_bindgen]
+impl JsDebugStreets {
```
dabreegster, line 85: It's also a bit awkward to have to write a "wrapper struct" with all of these wrapper functions. But it might be mostly unavoidable, or unavoidable without some macros or something.

street_network/src/lib.rs
```diff
@@ -3,6 +3,7 @@ extern crate anyhow;
 #[macro_use]
 extern crate log;
 
+use std::cell::RefCell;
```
dabreegster, line 6: https://doc.rust-lang.org/std/cell/index.html is a bit of a Rust trick here. In the middle of a transformation, we might be immutably borrowing `StreetNetwork`, because we're iterating over all roads or something. If we want to write some debug info and record it in the `StreetNetwork`, how can we do that? `RefCell` lets us cheat and do something mutable with an immutable reference. It just does a double-borrow check at runtime. It's always safe & simple in this case; we have some methods to push a labelled point and immediately relinquish the borrow.

The alternative would be passing around some kind of mutable `Debugger` object to the transformations and accumulating debug state there. That might be nicer to switch to eventually, but for the moment, the API would be more awkward -- we'd probably have to do something like `debugger.label_intersection(&streets, "something here")`

street_network/src/lib.rs
```diff
@@ -216,6 +241,44 @@ impl StreetNetwork {
...
+                debug_points: RefCell::new(Vec::new()),
+                debug_polylines: RefCell::new(Vec::new()),
+            },
+            points: self.debug_points.borrow_mut().drain(..).collect(),
+            polylines: self.debug_polylines.borrow_mut().drain(..).collect(),
+        }
+    }
+
+    pub(crate) fn debug_intersection<I: Into<String>>(&self, i: osm::NodeID, label: I) {
+        if self.debug_steps.borrow().is_empty() {
```
dabreegster, line 264: Short-circuit and don't write anything if debugging is disabled. We push a copy of the original `StreetNetwork` on at the beginning of `apply_transformations_stepwise_debugging`, thus indicating we're in a debug mode

tests-web/www/js/main.js
```diff
@@ -219,14 +254,37 @@ class TestCase {
...
+      }
     } catch (err) {
       window.alert(`Reimport failed: ${err}`);
     }
   }
 }
 
 function setupLeafletMap(mapContainer) {
-  const map = L.map(mapContainer, { maxZoom: 21 }).setView([40.0, 10.0], 4);
+  // Make it smoother to zoom farther into the map
```
dabreegster, line 281: These settings help a bit, but they could still use some work

tests-web/www/js/main.js
```diff
@@ -251,3 +309,48 @@ const useMap = (map) => {
   map.openTest = makeOpenTest(map);
   console.info("New map created! File drops enabled.", container);
 };
+
+// TODO A Leaflet control isn't the right abstraction. We want a popup modal
```
dabreegster, line 313: As the TODO says. For now, this is usable, but the style is bad and the code is tedious. I'll look into improvements post-conference

PR #54, opened 2022-08-08
Start detecting a certain pattern of dual carriageways, labelling the

dabreegster:
parts for later transformation.

This is an experiment disabled by default; it doesn't yet even correctly
detect every test case.

https://user-images.githubusercontent.com/1664407/183421293-ce165d54-eb6a-48b4-8955-39354d883b62.mp4

Alright, this PR gets interesting. I added a new test case, `northgate_dual_carriageway`, that nicely exhibits lots of complexities of a dual carriageway (DC). This PR starts to pattern match it and label parts of the structure:

- the two intersections where the DC starts or ends. I believe this matches the concept of `MultiConnection`
- each side of the DC, expressed as a sequence of one-way roads pointing in a chain
- "bridge" roads leading between the two sides
- "branch" roads leading from one side to somewhere else

What I'll attempt to do next is to put all of the intersections on each side, the branches, and the bridges in a "linear" order, "zipping" the two sides together. There'll be some way to figure out which pairs of roads from both sides should get zipped together, with their lanes merged. Not sure how/if it'll work yet!

The detection in this PR isn't robust; it picks up one example in northgate, but doesn't even find all of the examples there. So it's disabled by default. I want to get one example working end-to-end, then make things more robust gradually.

### DIFF
--- a/street_network/src/transform/dual_carriageways.rs
+++ b/street_network/src/transform/dual_carriageways.rs
@@ -1,55 +1,350 @@
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
+
+use geom::Distance;
 
 use crate::{osm, OriginalRoad, StreetNetwork};
 
-/// Does this road go between two divided one-ways? Ideally they're tagged explicitly
-/// (https://wiki.openstreetmap.org/wiki/Tag:dual_carriageway%3Dyes), but we can also apply simple
-/// heuristics to guess this.
-#[allow(unused)]
-pub fn connects_dual_carriageway(streets: &StreetNetwork, id: &OriginalRoad) -> bool {
-    let connectors_angle = streets.roads[id].angle();
-    // There are false positives like https://www.openstreetmap.org/way/4636259 when we're looking
-    // at a segment along a marked dual carriageway. Filter out by requiring the intersecting dual
-    // carriageways to differ by a minimum angle.
-    let within_degrees = 10.0;
-
-    let mut i1_dual_carriageway = false;
-    let mut oneway_names_i1: BTreeSet<String> = BTreeSet::new();
-    for r in streets.roads_per_intersection(id.i1) {
-        let road = &streets.roads[&r];
-        if r == *id || connectors_angle.approx_eq(road.angle(), within_degrees) {
-            continue;
-        }
-        if road.osm_tags.is("dual_carriageway", "yes") {
-            i1_dual_carriageway = true;
-        }
-        if road.oneway_for_driving().is_some() {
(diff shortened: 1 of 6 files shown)


street_network/src/transform/dual_carriageways.rs
```diff
@@ -1,55 +1,350 @@
...
+
+use geom::Distance;
 
 use crate::{osm, OriginalRoad, StreetNetwork};
 
-/// Does this road go between two divided one-ways? Ideally they're tagged explicitly
-/// (https://wiki.openstreetmap.org/wiki/Tag:dual_carriageway%3Dyes), but we can also apply simple
-/// heuristics to guess this.
-#[allow(unused)]
-pub fn connects_dual_carriageway(streets: &StreetNetwork, id: &OriginalRoad) -> bool {
```
dabreegster, line 9: This was old unused code

street_network/src/transform/dual_carriageways.rs
```diff
@@ -1,55 +1,350 @@
...
+                mc.debug(streets);
+            }
+
+            // TODO Just work on one right now
+            break;
+        }
+    }
+}
+
+// TODO We should do this in classify_intersections.rs?
```
dabreegster, line 28: Maybe this should happen in `classify_intersections.rs` upfront. I'm not trying to reason about highway entrances/exits with this transformation, so maybe the pattern matching doesn't apply completely

street_network/src/transform/dual_carriageways.rs
```diff
@@ -1,55 +1,350 @@
...
+        let mut side2 = Self::orient_oneways(side2)?;
+
+        // Which one goes from i1->i2?
+        let i1 = mc.i;
+        let i2 = i2_v1;
+        for swap in [false, true] {
+            if swap {
+                std::mem::swap(&mut side1, &mut side2);
+            }
+            let side1_endpts = (side1[0].i1, side1.last().as_ref().unwrap().i2);
```
dabreegster, line 134: Maybe we need a `struct` to represent a list of roads in order, with methods to walk the intersections in order or get endpoints easily

tests-web/www/js/main.js
```diff
@@ -319,19 +324,34 @@ const SettingsControl = L.Control.extend({
...
     };
 
-    var label = L.DomUtil.create("label");
-    label.for = "debugEachStep";
-    label.innerText = "Debug each transformation step";
+    var label1 = L.DomUtil.create("label");
+    label1.for = "debugEachStep";
+    label1.innerText = "Debug each transformation step\n";
+
+    var checkbox2 = L.DomUtil.create("input");
```
dabreegster, line 341: It's getting very boilerplatey to do this :)

PR #55, opened 2022-08-09
Add two new GeoJSON layers to render more lane detail. #24

dabreegster:
- one with a polygon per lane, so we can use color to distinguish
  parking/bus/bike/driving lanes
- another with dashed lines showing the road center line and separators
  between lanes

Screenshot from 2022-08-09 10-46-33
Screenshot from 2022-08-09 10-48-47

I've chosen to do things in a simple way with GeoJSON and hardcoding some styling about dashed lines in the Rust layer for the moment, but this isn't ideal at all. I'd also like to add arrows for direction and bike/bus/parking icons spaced along some lanes. SVG is looking like an attractive option -- we could ideally return something saying to place some parking icon at a certain position and angle, and let the JS caller or others swap in icons or interpret the styling appropriately.

### DIFF
--- a/street_network/src/lib.rs
+++ b/street_network/src/lib.rs
@@ -501,6 +501,30 @@ impl RawRoad {
             self.osm_tags.get(osm::OSM_WAY_ID).unwrap()
         )
     }
+
+    pub fn total_width(&self) -> Distance {
+        self.lane_specs_ltr.iter().map(|l| l.width).sum()
+    }
+
+    /// Returns one PolyLine representing the center of each lane in this road. Pass in
+    /// `road_center` generated from `InitialMap` (trimmed from the intersection) or from
+    /// `osm_center_points`. The result also faces the same direction as the road.
+    pub fn get_lane_center_lines(&self, road_center: &PolyLine) -> Vec<PolyLine> {
+        let total_width = self.total_width();
+
+        let mut width_so_far = Distance::ZERO;
+        let mut output = Vec::new();
+        for lane in &self.lane_specs_ltr {
+            width_so_far += lane.width / 2.0;
+            output.push(
+                road_center
+                    .shift_from_center(total_width, width_so_far)
+                    .unwrap_or_else(|_| road_center.clone()),
+            );
+            width_so_far += lane.width / 2.0;
+        }
+        output
+    }
 }
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
--- a/street_network/src/render.rs
+++ b/street_network/src/render.rs
@@ -6,7 +6,7 @@ use abstutil::Timer;
 use anyhow::Result;
 use geom::Distance;
 
-use crate::{DebugStreets, StreetNetwork};
+use crate::{DebugStreets, LaneType, StreetNetwork};
 
 impl StreetNetwork {
     /// Saves the plain GeoJSON rendering to a file.
@@ -18,7 +18,7 @@ impl StreetNetwork {
         Ok(())
     }
 
-    /// Generates a plain GeoJSON rendering of roads and intersections.
+    /// Generates a plain GeoJSON rendering with one polygon per road and intersection.
     pub fn to_geojson(&self, timer: &mut Timer) -> Result<String> {
         // TODO InitialMap is going away very soon, but we still need it
         let initial_map = crate::initial::InitialMap::new(self, timer);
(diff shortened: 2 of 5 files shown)


street_network/src/lib.rs
```diff
@@ -501,6 +501,30 @@ impl RawRoad {
...
+    pub fn get_lane_center_lines(&self, road_center: &PolyLine) -> Vec<PolyLine> {
+        let total_width = self.total_width();
+
+        let mut width_so_far = Distance::ZERO;
+        let mut output = Vec::new();
+        for lane in &self.lane_specs_ltr {
+            width_so_far += lane.width / 2.0;
+            output.push(
+                road_center
+                    .shift_from_center(total_width, width_so_far)
```
dabreegster, line 521: https://github.com/a-b-street/abstreet/pull/865 for context. If we repeatedly project a polyline left or right, sharp corners can explode badly. Workaround is to try to always work from the road center and shift just once from there

street_network/src/render.rs
```diff
@@ -65,54 +65,77 @@ impl StreetNetwork {
         Ok(output)
     }
 
-    /// Generates a more detailed GeoJSON rendering of roads and intersections.
-    pub fn to_detailed_geojson(&self, timer: &mut Timer) -> Result<String> {
```
dabreegster, line 69: I got rid of this "detailed" experiment that was repeating some of what the "plain" layer does. I agree it makes sense to just have a bunch of layers representing different things, and let the caller mix them together however.

street_network/src/render.rs
```diff
@@ -65,54 +65,77 @@ impl StreetNetwork {
...
+            let lane_centers =
+                road.get_lane_center_lines(&initial_map.roads[id].trimmed_center_pts);
+
+            for (idx, pair) in road.lane_specs_ltr.windows(2).enumerate() {
+                // Generate a "center line" between lanes of different directions
+                if pair[0].dir != pair[1].dir {
+                    let between = lane_centers[idx].shift_right(pair[0].width / 2.0)?;
+                    // TODO Ideally we would return a full LineString, and the caller would choose
+                    // how to style these as thickened dashed lines.
+                    // TODO We could also at least return a MultiPolygon here
```
dabreegster, line 110: Right, so here I'm returning a polygon for every single piece of the dashed lines, and the parameters are hardcoded in. Maybe we could just return a GeoJSON linestring representing the boundary between lanes, add some properties to say what kind of boundary it is, and then dynamically generate dashing in JS.

Also all of this may change when we switch this layer to osm2lanes, which has a lane type representing semantic boundaries.

tests-web/www/js/layers.js
```diff
@@ -1,49 +1,70 @@
 export const makePlainGeoJsonLayer = (text) => {
-  return new L.geoJSON(JSON.parse(text), { style: styleGeoJson });
-};
+  const intersectionColours = {
```
dabreegster, line 2: It felt a bit cleaner to me to group everything related to one layer in the same function, but happy to reconsider

tests-web/www/js/layers.js
```diff
@@ -1,49 +1,70 @@
...
-  if (feature.geometry.type === "Polygon") {
-    return {
-      color: intersectionColours[feature.properties?.complexity],
-      weight: 1,
-      fillOpacity: 0.7,
-    };
-  }
-  return { color: "#f55" };
+export const makeLanePolygonLayer = (text) => {
+  // These could change per locale
```
dabreegster, line 27: I just lifted some colors from A/B Street's "classic" / original scheme. They're North American centric. It'd be pretty easy at this layer to just look at the locale or country code from Overpass and switch between some styles!

street_network/src/render.rs
```diff
@@ -65,54 +65,77 @@ impl StreetNetwork {
         Ok(output)
     }
 
-    /// Generates a more detailed GeoJSON rendering of roads and intersections.
-    pub fn to_detailed_geojson(&self, timer: &mut Timer) -> Result<String> {
+    /// Generates a polygon per lane, with a property indicating type.
+    pub fn to_lane_polygons_geojson(&self, timer: &mut Timer) -> Result<String> {
         // TODO InitialMap is going away very soon, but we still need it
         let initial_map = crate::initial::InitialMap::new(self, timer);
```
dabreegster, line 71: Oh and this starts to get expensive, since all of the layer functions here call it. I need to finish #2 finally.

PR #56, opened 2022-08-09
Create a new Leaflet layer control widget. It internally manages groups

dabreegster:
of layers, which can be managed together. Use this to establish groups
for the pre-built test case layers, for reimported layers, and for each
step in a debugging sequence.

https://user-images.githubusercontent.com/1664407/183684304-7347720d-1120-46c5-8543-8b0cba6f65b0.mp4

Next up, I'm going to make all the debugging groups bundled together in their own special group and have controls for going back and forth. This'll start to get us to a nicer debugging experience / a way to tell the story of transformations.

### DIFF
--- a/tests-web/www/js/controls.js
+++ b/tests-web/www/js/controls.js
@@ -1,0 +1,197 @@
+// TODO A Leaflet control isn't the right abstraction. We want a popup modal
+// that totally blocks the map and other things. Things like disabling
+// settingsButton to stop multiple of these controls is a total hack.
+const SettingsControl = L.Control.extend({
+  // TODO Centered would be great. https://github.com/Leaflet/Leaflet/issues/8358
+  options: {
+    position: "topleft",
+  },
+  onAdd: function (map) {
+    const checkbox1 = makeCheckbox(
+      "debugEachStep",
+      "Debug each transformation step\n",
+      this.options.app.importSettings.debugEachStep,
+      (checked) => {
+        this.options.app.importSettings.debugEachStep = checked;
+      }
+    );
+    const checkbox2 = makeCheckbox(
+      "dualCarriagewayExperiment",
+      "Enable dual carriageway experiment\n",
+      this.options.app.importSettings.dualCarriagewayExperiment,
+      (checked) => {
+        this.options.app.importSettings.dualCarriagewayExperiment = checked;
+      }
+    );
+
+    const button = L.DomUtil.create("button");
+    button.type = "button";
+    button.innerHTML = "Confirm";
+    button.onclick = () => {
+      this.remove();
+      document.getElementById("settingsButton").disabled = false;
+    };
+
+    var group = makeDiv([checkbox1, checkbox2, button]);
+    group.style = "background: black; padding: 10px;";
+    L.DomEvent.disableClickPropagation(group);
+    return group;
+  },
+});
+
+export const makeSettingsControl = function (app) {
+  return new SettingsControl({ app: app });
+};
+
+export class LayerGroup {
+  constructor(name, map) {
+    this.name = name;
+    this.layers = [];
+    this.enabled = true;
+    this.map = map;
+  }
+
+  addLayer(name, layer, { enabled = true } = {}) {
+    this.layers.push({ name, enabled, layer });
+  }
+
+  // Doesn't re-render
+  setEnabled(enabled) {
+    this.enabled = enabled;
+    for (const layer of this.layers) {
+      layer.enabled = enabled;
+    }
+  }
+
+  renderControls() {
+    var members = [
+      makeCheckbox(
+        this.name,
+        `<u>${this.name}</u>`,
+        this.enabled,
+        (checked) => {
+          this.setEnabled(checked);
+
+          // Rerender
+          document.getElementById(this.name).replaceWith(this.renderControls());
+        }
+      ),
+    ];
+    for (const layer of this.layers) {
+      members.push(
+        makeCheckbox(
+          this.name + "_" + layer.name,
+          layer.name,
+          layer.enabled,
+          (checked) => {
+            layer.enabled = checked;
+            if (checked) {
+              this.map.addLayer(layer.layer);
+            } else {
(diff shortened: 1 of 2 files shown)


tests-web/www/js/controls.js
```diff
@@ -0,0 +1,197 @@
...
+    ];
+    for (const layer of this.layers) {
+      members.push(
+        makeCheckbox(
+          this.name + "_" + layer.name,
+          layer.name,
+          layer.enabled,
+          (checked) => {
+            layer.enabled = checked;
+            if (checked) {
```
dabreegster, line 88: Turns out `map.addLayer` and `removeLayer` are idempotent, so we can be a bit sloppy here and just set the state to whatever want.

tests-web/www/js/controls.js
```diff
@@ -0,0 +1,197 @@
...
+  },
+
+  renderControls: function () {
+    var members = [];
+    for (const group of this.options.groups) {
+      members.push(group.renderControls());
+    }
+    var group = makeDiv(members);
+    group.style = "background: black; padding: 10px;";
+    L.DomEvent.disableClickPropagation(group);
```
dabreegster, line 126: Otherwise clicking on checkboxes can act like double clicking to zoom

tests-web/www/js/controls.js
```diff
@@ -0,0 +1,197 @@
...
+};
+
+// Helpers
+
+function makeCheckbox(id, label, enabled, callback) {
+  var checkbox = L.DomUtil.create("input");
+  checkbox.id = id;
+  checkbox.type = "checkbox";
+  checkbox.checked = enabled;
+  checkbox.onclick = () => {
```
dabreegster, line 180: Leaflet docs say to use these `DomEvent` helpers and do all this cleanup. I'm not sure why. We might be leaking DOM elements or something right now, but the end-user behavior seems correct to me

tests-web/www/js/main.js
```diff
@@ -14,20 +14,24 @@ import {
...
-    const [map, layerControl] = setupLeafletMap(mapContainer);
-    this.map = map;
-    this.layerControl = layerControl;
+    this.map = setupLeafletMap(mapContainer);
     this.currentTest = null;
     this.importSettings = {
       debugEachStep: false,
       dualCarriagewayExperiment: false,
     };
+    this.layers = makeLayerControl(this).addTo(this.map);
```
dabreegster, line 34: Previously we had a `TestCase` own a list of layers, but the `LayerControl` live at the app level. That got confusing here. Now all of the layers in their groups are "owned" by the new layer control widget. That lives at the app level.

tests-web/www/js/main.js
```diff
@@ -82,42 +77,34 @@ export class StreetExplorer {
...
-    var layers = [];
-    layers.push(app.addLayer("Geometry", rawMapLayer));
-    layers.push(app.addLayer("OSM", makeOsmLayer(osmInput)));
-    layers.push(
-      app.addLayer("Network", await makeDotLayer(await network, { bounds }))
-    );
+    var group = new LayerGroup("built-in test case", app.map);
+    group.addLayer("Geometry", rawMapLayer);
+    group.addLayer("OSM", makeOsmLayer(osmInput), { enabled: false });
+    group.addLayer("Network", await makeDotLayer(network, { bounds }));
```
dabreegster, line 100: I want to disable this one by default, but if I try that, then for some reason the SVG shows up covering the entire screen and breaks everything. There's some funniness with `makeDotLayer` that I don't understand yet

tests-web/www/js/main.js
```diff
@@ -143,56 +130,12 @@ class TestCase {
...
-        if (debugGeojson) {
-          layers.push(
-            app.addLayer(
-              `${step.getLabel()}: debug`,
-              makeDebugLayer(debugGeojson)
-            )
-          );
-        }
-      }
+      importOSM("Imported area", app, osmInput, drivingSide, true);
```
dabreegster, line 133: I finally refactored this code between importing a new area and reimporting!

tests-web/www/js/main.js
```diff
@@ -143,56 +130,12 @@ class TestCase {
...
+      const bounds = app.layers
```
dabreegster, line 134: One consequence of that is needing to go look for a certain named layer

tests-web/www/js/main.js
```diff
@@ -240,59 +181,60 @@ class TestCase {
...
-          );
-        }
+
+      const debugGeojson = step.toDebugGeojson();
+      if (debugGeojson) {
+        group.addLayer("Debug", makeDebugLayer(debugGeojson));
       }
-    } catch (err) {
-      window.alert(`Reimport failed: ${err}`);
+      group.setEnabled(i == numDebugSteps);
```
dabreegster, line 233: Only show the last step by default. This is temporary behavior until I make some kind of "scroll through this sequence of groups" control.

tests-web/www/js/main.js
```diff
@@ -328,11 +270,11 @@ function setupLeafletMap(mapContainer) {
     autoClose: true,
   }).addTo(map);
 
-  const layerControl = L.control
-    .layers({ OpenStreetMap: osm, ArcGIS: arcgis }, {})
+  L.control
```
dabreegster, line 273: Still using the built-in layer control widget just for choosing the basemap

tests-web/www/js/main.js
```diff
@@ -345,63 +287,3 @@ const useMap = (map) => {
   map.openTest = makeOpenTest(map);
   console.info("New map created! File drops enabled.", container);
 };
-
-// TODO A Leaflet control isn't the right abstraction. We want a popup modal
```
dabreegster, line 349: Moved to `controls.js`

tests-web/www/js/main.js
BudgieInWA: Oh... Adding it to the map is what moves it out of the DOM i guess... 

The map doesn't make much sense anyway, until it is layed out on an appropriately shaped canvas, and with a sensibly positioned nodes.

PR #58, opened 2022-08-10
Group the debug steps and make a new widget to scroll through them in…

dabreegster:
… order

https://user-images.githubusercontent.com/1664407/183846770-ddaf330b-68c2-470e-a975-cd8afddf89b4.mp4

A followup to yesterday. Pretty straightforward to debug transformations step-by-step now!

### DIFF
--- a/tests-web/www/js/controls.js
+++ b/tests-web/www/js/controls.js
@@ -26,9 +26,7 @@ const SettingsControl = L.Control.extend({
       }
     );
 
-    const button = L.DomUtil.create("button");
-    button.type = "button";
-    button.innerHTML = "Confirm";
+    const button = makeButton("Confirm");
     button.onclick = () => {
       this.remove();
       document.getElementById("settingsButton").disabled = false;
@@ -45,6 +43,8 @@ export const makeSettingsControl = function (app) {
   return new SettingsControl({ app: app });
 };
 
+// Manages a bunch of layers that can be independently toggled. The entire
+// group has a toggle to enable or disable everything.
 export class LayerGroup {
   constructor(name, map) {
     this.name = name;
@@ -55,13 +55,25 @@ export class LayerGroup {
 
   addLayer(name, layer, { enabled = true } = {}) {
     this.layers.push({ name, enabled, layer });
+
+    if (enabled) {
+      this.map.addLayer(layer);
+    } else {
+      this.map.removeLayer(layer);
+    }
   }
 
-  // Doesn't re-render
+  // Updates the map, but doesn't re-render any controls
   setEnabled(enabled) {
     this.enabled = enabled;
     for (const layer of this.layers) {
       layer.enabled = enabled;
+
+      if (enabled) {
+        this.map.addLayer(layer.layer);
+      } else {
+        this.map.removeLayer(layer.layer);
+      }
     }
   }
 
@@ -94,9 +106,7 @@ export class LayerGroup {
         }
       );
 
-      const download = L.DomUtil.create("button");
-      download.type = "button";
-      download.innerHTML = "Download";
+      const download = makeButton("Download");
       download.onclick = () => {
         downloadGeneratedFile(
           `${layer.name}.geojson`,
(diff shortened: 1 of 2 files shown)


tests-web/www/js/controls.js
```diff
@@ -105,21 +115,80 @@ export class LayerGroup {
...
+    next.onclick = () => {
+      this.groups[this.current].setEnabled(false);
+      this.current += 1;
+      this.groups[this.current].setEnabled(true);
+
+      // Rerender
+      document.getElementById(this.name).replaceWith(this.renderControls());
+    };
+
+    var row1 = L.DomUtil.create("u");
```
dabreegster, line 172: I was looking for a way to do something like `const row1 = document.makeDOMElement(`<u>${this.name}</u>`)` to create elements more succinctly. But I couldn't spot a simple answer for this. https://stackoverflow.com/questions/494143/creating-a-new-dom-element-from-an-html-string-using-built-in-dom-methods-or-pro has a bunch of ways, and maybe `new DOMParser().parseFromString(html, 'text/html').body.childNodes[0]` is the very verbose one that's most modern?

tests-web/www/js/controls.js
```diff
@@ -105,21 +115,80 @@ export class LayerGroup {
...
+  }
+
+  remove() {
+    this.groups[this.current].setEnabled(false);
+  }
+}
+
+// Manages a list of LayerGroups or SequentialLayerGroups.
+//
+// The interface required of each member: renderControls(), remove(), name
```
dabreegster, line 191: Maybe I do want TypeScript ^^;

tests-web/www/js/controls.js
BudgieInWA: I think the easiest way is to assign to innerHTML. This sort of thing:

```js
document.getElementById(blah).innerHTML = `<${tag} class="${classnames(...)}">Do it</${tag}>`;
```

You can create elements incrementally using the [low level API](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement#basic_example):

```js
function addElement () {
  const newDiv = document.createElement("div");
  const newContent = document.createTextNode("Hi there and greetings!");
  newDiv.appendChild(newContent);
  const currentDiv = document.getElementById("div1");
  document.body.insertBefore(newDiv, currentDiv);
}
```

tests-web/www/js/controls.js
```diff
@@ -105,21 +115,80 @@ export class LayerGroup {
...
+      this.groups[this.current].setEnabled(true);
+
+      // Rerender
+      document.getElementById(this.name).replaceWith(this.renderControls());
+    };
+
+    var row1 = L.DomUtil.create("u");
+    row1.innerText = this.name;
+    const row2 = makeDiv([prev, label, next]);
+    const column = makeDiv([
```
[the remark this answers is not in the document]
BudgieInWA: There's a CSS breakpoint in there based on ratio. The layout for portrait is not right at all, I know that :)

tests-web/www/js/controls.js
```diff
@@ -105,21 +115,80 @@ export class LayerGroup {
...
+    const prev = makeButton("Previous");
+    prev.disabled = this.current == 0;
+    prev.onclick = () => {
+      this.groups[this.current].setEnabled(false);
+      this.current -= 1;
+      this.groups[this.current].setEnabled(true);
+
+      // Rerender
+      document.getElementById(this.name).replaceWith(this.renderControls());
+    };
```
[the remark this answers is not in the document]
BudgieInWA: It raises an interesting question about what the controls should be. Maybe the widget takes keyboard focus and eats the events. Maybe we choose other keys.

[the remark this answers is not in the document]
dabreegster: Yeah, I'm not sure what the UX should even be. Arrow keys were just a first instinct. Maybe browsing through the debug steps should even automatically pan+zoom the map to focus on things that've changed.

tests-web/www/js/controls.js
dabreegster: There has to be an element out there somewhere to do `innerHTML`. So in these cases where we're building up new things to insert somewhere, we could insert a blank element first, then overwrite `innerHTML` for it?

PR #60, opened 2022-08-11
Draw arrows showing the direction of lanes. #24

dabreegster:
Screenshot from 2022-08-11 08-59-51

Like the rest of the lane marking, it's likely too low-level to prescribe specific arrow polygons in the geojson, but this moves us forward quickly. I need this to make sure a WIP transformation is preserving directions correctly!

### DIFF
--- a/street_network/src/render.rs
+++ b/street_network/src/render.rs
@@ -4,9 +4,9 @@ use std::path::Path;
 
 use abstutil::Timer;
 use anyhow::Result;
-use geom::Distance;
+use geom::{ArrowCap, Distance, PolyLine};
 
-use crate::{DebugStreets, LaneType, StreetNetwork};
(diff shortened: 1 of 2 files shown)


street_network/src/render.rs
```diff
@@ -136,6 +136,35 @@ impl StreetNetwork {
...
+            // Draw arrows along any travel lane
+            for (lane, mut center) in road.lane_specs_ltr.iter().zip(lane_centers.into_iter()) {
+                if !lane.lt.is_for_moving_vehicles() {
+                    continue;
+                }
+                if lane.dir == Direction::Back {
+                    center = center.reversed();
+                }
+
+                let step_size = Distance::meters(20.0);
```
dabreegster, line 149: All arbitrary, quickly tuned. Thickness ought to probably relate to lane width

Issue #59, opened 2020-09-15
Include separate cyclepaths and trails


dabreegster:
Splitting from a-b-street/abstreet#139. The goals here are just:
1) connectivity -- can bikes successfully route over things like the Burke Gilman as expected
2) geometry + rendering -- does it look reasonable
3) bike-only -- just model as one or two bike lanes, ignore foot access for the moment. (Once this bug is solved, a-b-street/abstreet#139 can add a special lane type that has mixed access)

dabreegster:
Moved to the osm2streets repo, where I'm renewing efforts on this. No promises, but State of the Map next week is the much-needed deadline...

dabreegster:
The default in osm2streets is now using separate cycletracks and footpaths as they are in OSM. Still have ongoing efforts to improve the geometry and optionally zip/snap things together

PR #61, opened 2022-08-11
Start a new transformation that merges a class of parallel cycleways to

dabreegster:
the main road, preserving lanes. #59

Disabled by default.

I have been puzzling through how to simplify cases like this for a long time, but the last progress was almost two years ago: https://github.com/a-b-street/abstreet/pull/348. I took a fresh approach by

1) looking for patterns in the graph structure, instead of the [previous approach](https://github.com/a-b-street/osm2streets/blob/main/street_network/src/transform/snappy.rs) of placing a bunch of perpendicular test lines and seeing what roads are hit
2) Only attempting cases where the cycleway is completely parallel to main roads. Not yet handling the trickier edge cases where a cycleway is parallel for some percentage of the polyline, but then it curves away.

The result is delightful. Along Roosevelt in Seattle, before:
Screenshot from 2022-08-11 11-23-11
and after:
Screenshot from 2022-08-11 11-23-24
The red lane is the buffer lane inserted. With additional tagging in OSM, we can know if these are flex posts, concrete bollards, a Jersey barrier, striping, etc.

And much more extravagantly, St George's just past the Elephant & Castle roundabout. Before:
Screenshot from 2022-08-11 11-25-42
After:
Screenshot from 2022-08-11 11-26-30

As you can see to the right, the algorithm is far from finished. It gets confused by the "triangular" shape of the cycleway graph there and just leaves it alone. I'm leaving this transformation disabled until more dramatic problems are fixed, like in Taipei... before:
Screenshot from 2022-08-11 11-28-40
Note this paradise of parallel cycletrack is actually mostly-pedestrian space in reality, but that's another issue. After:
Screenshot from 2022-08-11 11-29-20

Clearly something went very wrong. I'll improve the incremental interactive debugging and figure out what. Right now the debugging is shown for all cycleway segments at the same time; we can split this into smaller steps.
Screenshot from 2022-08-11 11-30-47

### DIFF
--- a/street_network/src/transform/separate_cycletracks.rs
+++ b/street_network/src/transform/separate_cycletracks.rs
@@ -1,0 +1,215 @@
+use geom::{Distance, PolyLine};
+
+use crate::{osm, BufferType, Direction, LaneSpec, LaneType, OriginalRoad, StreetNetwork};
+
+/// Find cycleway segments that exist as separate objects, parallel to a main road. Merge (or
+/// "snap") them into the main road, inserting a buffer lane to represent the physical division.
+pub fn snap_cycleways(streets: &mut StreetNetwork) {
+    for cycleway in find_cycleways(streets) {
+        cycleway.debug(streets);
+        snap(streets, cycleway);
+    }
+}
+
+// We're only pattern matching on one type of separate cycleway right now. This represents a single
+// RawRoad that's parallel to one or more main_roads.
+//
+// X--X
+// C  M
+// C  M
+// C  M
+// C  X
+// C  M
+// C  M
+// X--X
+//
+// C is the cycleway segment. X are intersections. M are main roads -- note there are two matching
+// up to this one cycleway. The '-'s are short connector roads between the two.
+struct Cycleway {
+    cycleway: OriginalRoad,
+    cycleway_center: PolyLine,
+    // Just to distinguish different cycleways when debugging
+    debug_idx: usize,
+    main_road_i1: osm::NodeID,
+    main_road_i2: osm::NodeID,
+    main_roads: Vec<OriginalRoad>,
+}
+
+impl Cycleway {
+    fn debug(&self, streets: &StreetNetwork) {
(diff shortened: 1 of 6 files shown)


street_network/src/transform/separate_cycletracks.rs
```diff
@@ -0,0 +1,215 @@
...
+                    main_road_endpoints.push(candidates[0].other_side(i));
+                }
+            }
+
+            if main_road_endpoints.len() == 2 {
+                // Often the main road parallel to this cycleway segment is just one road, but it
+                // might be more.
+                let main_road_i1 = main_road_endpoints[0];
+                let main_road_i2 = main_road_endpoints[1];
+                // Find all main road segments "parallel to" this cycleway, by pathfinding between
```
dabreegster, line 75: We could perhaps be more paranoid here by checking that the cycleway center line and all of the main road center lines are nice and parallel. If they curve away from each other a bit, then even if the connection roads are under our threshold of 10 meters, we may lose some geometric information by snapping.

street_network/src/transform/separate_cycletracks.rs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                main_road.lane_specs_ltr.len() - 1
+            } else {
+                main_road.lane_specs_ltr.len()
+            }
+        };
+
+        streets.debug_road(r, format!("snap_to_left = {snap_to_left}, oriented_same_way = {oriented_same_way}, insert_idx = {insert_idx}"));
+
+        // This logic thankfully doesn't depend on driving side at all!
```
dabreegster, line 181: I struggled to write this without scribbling diagrams on paper and using temporary debug visualization. It was a similar experience to [sausage links](https://github.com/a-b-street/osm2streets/blob/606cf03f46e79f69ea91a1cd9d62b9db140f2c07/street_network/src/transform/sausage_links.rs#L96). I hope there will be some better patterns to reason through this kind of thing. Maybe #1 is relevant.

PR #66, opened 2022-08-16
Highlight lane and intersection polygons when hovering, click to show…

dabreegster:
… info.

Since the plain geometry layer has different properties exposed, the
unit test data all changes too.

Makes things a bit more alive, emphasizes that we're creating lane-level objects, not just some rendering trick

https://user-images.githubusercontent.com/1664407/184862071-7d46bef0-c056-4a25-8846-aa14f8fb6b18.mp4


street_network/src/render.rs
```diff
@@ -25,23 +25,18 @@ impl StreetNetwork {
 
         let mut pairs = Vec::new();
 
-        // Add a line-string and polygon per road
+        // Add a polygon per road
         for (id, road) in &initial_map.roads {
-            let properties = make_props(&[
```
dabreegster, line 30: I took out the road center line. This doesn't have much use in rendering probably. As we refine the JS/anything API to osm2streets, we can expose this if it's useful

street_network/src/render.rs
```diff
@@ -80,7 +76,11 @@ impl StreetNetwork {
                 pairs.push((
                     pl.make_polygons(lane.width)
                         .to_geojson(Some(&self.gps_bounds)),
-                    make_props(&[("type", format!("{:?}", lane.lt).into())]),
+                    make_props(&[
```
dabreegster, line 79: Adding more properties to show in the popup

tests-web/www/js/layers.js
```diff
@@ -11,15 +11,34 @@ export const makePlainGeoJsonLayer = (text) => {
...
         return {
-          color: intersectionColours[feature.properties?.complexity],
+          color: intersectionColours[feature.properties.complexity],
           weight: 1,
           fillOpacity: 0.7,
         };
       }
       return { color: "#f55" };
     },
+    onEachFeature: function (feature, layer) {
```
dabreegster, line 23: Thanks to https://leafletjs.com/examples/choropleth/

PR #73, opened 2022-08-16
Lookup driving side from Overpass

dabreegster:
This isn't working yet... I swear this query was correct for a few times, but it's blank in a Paris example: `https://overpass-api.de/api/interpreter?data=[out:csv(driving_side)];is_in(2.3454737663269047, 48.82723414504365)->.enclosing; area.enclosing["driving_side"]; out tags;`

@droogmic, I based this off your code in osm2lanes. Any idea what I've done wrong?

### DIFF
--- a/tests-web/www/js/controls.js
+++ b/tests-web/www/js/controls.js
@@ -34,49 +34,7 @@ const SettingsControl = L.Control.extend({
       }
     );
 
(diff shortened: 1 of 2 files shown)


dabreegster:
Abandoning in favor of #108!

PR #84, opened 2022-09-04
Group and order the things in the sidebar a bit better

BudgieInWA:
The test list has become a dropdown that doubles as the "current test case" title, and the test specific controls have been put in a box.

The import options could go in their own (collapsible) box in the sidebar, next.

No test case

custom area

test case loaded

### DIFF
--- a/tests-web/www/js/main.js
+++ b/tests-web/www/js/main.js
@@ -55,16 +55,18 @@ export class StreetExplorer {
 
     // Wire up the import button
     const importButton = document.getElementById("import-view");
-    importButton.onclick = async () => {
-      if (app.map.getZoom() < 15) {
-        window.alert("Zoom in more to import");
-        return;
-      }
+    if (importButton) {
+      importButton.onclick = async () => {
+        if (app.map.getZoom() < 15) {
+          window.alert("Zoom in more to import");
+          return;
+        }
 
-      await app.setCurrentTest((app) =>
-        TestCase.importCurrentView(app, importButton)
-      );
-    };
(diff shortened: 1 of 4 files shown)


tests-web/www/js/main.js
```diff
@@ -55,16 +55,18 @@ export class StreetExplorer {
...
-      if (app.map.getZoom() < 15) {
-        window.alert("Zoom in more to import");
-        return;
-      }
-
-      await app.setCurrentTest((app) =>
-        TestCase.importCurrentView(app, importButton)
-      );
-    };
+    if (importButton) {
```
dabreegster, line 58: Could it ever be missing? I need to read up about initialization order on the web. This code is triggered in a `<script type="module">` in `<head>`. The button is defined in `<body>`. Do we need to specify something to make sure the body is created first?

dabreegster (approved):
Thanks! This is a much more clear UX

> The import options could go in their own (collapsible) box in the sidebar, next.

That makes _way_ more sense than a Leaflet control, yeah. I don't know why I was trying to make a modal / popup.

tests-web/www/js/main.js
BudgieInWA: When loading the page with a test in the URL, the code just above awaits the loading of the test, which replaces the button.

dabreegster:
I'll merge this.

FYI, I'm going to next be working on `street_network` code to start cutting over to osm2lanes. Only UI change should be plumbing a setting to opt into the new osm2lanes code

PR #85, opened 2022-09-04
Use osm2lanes, guarded by experiment flag

dabreegster:
For https://github.com/a-b-street/osm2lanes/issues/248

https://user-images.githubusercontent.com/1664407/188329575-1cb68cd2-c61f-44cf-9d1e-d0f7716eeba3.mp4

The web app now has a setting to opt into osm2lanes, and the unit tests here can now also check osm2lanes + translation. This sets us up for iterating rapidly on osm2lanes and its integration here.

CC @droogmic and @BudgieInWA. Feel free to review async, I'll likely merge in the morning, but address feedback whenever is good


street_network/src/lanes/osm2lanes.rs
```diff
@@ -0,0 +1,355 @@
+use anyhow::Result;
```
dabreegster, line 1: This is adapted from https://github.com/a-b-street/abstreet/pull/890. I wrote it in April and have mostly lost context on some of the subtler choices. There's lot in here that's weird/wrong or is stuff that we should fix in osm2lanes proper. But it's a start!

street_network/src/lanes/tests.rs
```diff
@@ -0,0 +1,211 @@
+use abstutil::Tags;
+use geom::Distance;
+
+use crate::{get_lane_specs_ltr, Direction, DrivingSide, MapConfig};
+
+// osm2lanes has a more extensive unit test suite, so why does this one exist? This also checks the
```
dabreegster, line 6: I decided to keep these unit tests long term. They serve a different purpose than the more detailed ones in osm2lanes

tests/src/service_road_loop/geometry.json
```diff
@@ -5,86 +5,36 @@
         "coordinates": [
```
dabreegster, line 5: The diff here is due to updating `geom`.
before:
Screenshot from 2022-09-04 20-00-24
That road polygon is expressed as a multipolygon with two triangles, which is quite silly.
after:
Screenshot from 2022-09-04 20-00-49
The fix comes from https://github.com/a-b-street/abstreet/issues/951, which is an epic saga to make `geom::Polygon` always represent a "good" polygon composed of an exterior ring and optional interior rings.

PR #86, opened 2022-09-05
Move the import settings to the sidebar

BudgieInWA:
image


tests-web/www/js/main.js
```diff
@@ -68,28 +60,28 @@ export class StreetExplorer {
...
-      }
-    };
-
     return app;
   }
 
+  getImportSettings() {
+    try {
+      const data = new FormData(document.getElementById("import-settings"));
+      return Object.fromEntries(data);
```
BudgieInWA, line 69: Returning the values here exactly as they come out of the form is not the most clean; the consumer has to type cast and all that. But it is less overhead to add or change settings, so I think it's worth it here.

tests-web/www/js/main.js
```diff
@@ -197,13 +189,14 @@ class TestCase {
 
 function importOSM(groupName, app, osmXML, drivingSide, addOSMLayer) {
   try {
+    const importSettings = app.getImportSettings();
     const network = new JsStreetNetwork(osmXML, {
       driving_side: drivingSide,
-      debug_each_step: app.importSettings.debugEachStep,
-      dual_carriageway_experiment: app.importSettings.dualCarriagewayExperiment,
+      debug_each_step: !!importSettings.debugEachStep,
```
BudgieInWA, line 195: The "double bang" is a JS idiom for "cast to float". It's simply a double "not".

tests-web/www/js/main.js
```diff
@@ -146,7 +138,7 @@ class TestCase {
 
       importButton.innerText = "Importing OSM data...";
 
-      const drivingSide = app.importSettings.drivingSideForNewImports;
+      const drivingSide = app.getImportSettings().drivingSideForNewImports || "Right";
```
BudgieInWA, line 141: Ideally, the choice of driving side would be presented alongside the "import" button that uses it, instead of it being part of "import settings".

dabreegster (approved):
This is awesome, thanks!

Merging, so I can fix up #85 and also merge that

tests-web/www/css/main.css
```diff
@@ -39,6 +40,31 @@ body > aside {
   }
 }
 
+details.fieldset {
+  border: 2px groove silver;
+  padding: 0.75em;
+  margin-top: 0.5em;
+}
+details.fieldset > summary:first-of-type {
```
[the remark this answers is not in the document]
BudgieInWA: To be honest, I'm still learning a lot of the newer CSS features too! Mostly new properties, like `column-gap` from my last PR. The CSS selector "language" is worth putting a bit of time into I think, because a small set of [combinators](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Combinators), plus some [Pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) (wow, that list is 3 times as long as I expected!) give you a lot of power in selecting elements in the right context, without having to add anything to the markup.

My super-power though, is using the browser dev-tools to play with CSS. Right click an element, choose "Inspect", then you can see what rules are being applied to each element, visualise the margin/padding/etc., and edit the CSS in place (with appropriate autocomplete). Most CSS changes will first be implemented directly in the inspector, to get my property names right, see which approach will work, tweak lengths with "up" and "down" keys, etc. before copying the changes into the source file.

I just learnt about "[workspaces](https://developer.chrome.com/docs/devtools/workspaces/)" in Chrome DevTools, which lets the inspector modify the files directly on disk! It also means chrome is always using the most up to date version of each file, which avoids a problem I was having where my changes weren't visible due to the browser cache.

Issue #87, opened 2022-09-14
Start representing footpaths and shared-use paths


dabreegster:
Currently osm2streets and A/B Street don't import off-road paths primarily intended for pedestrians. One example of many is https://www.openstreetmap.org/way/362057587, a `highway=footway` in a park.

Also currently many spaces that're shared between pedestrians and cyclists are either not imported, or transformed into this amusing [tiny shoulder for walking, cycle lane one direction, cycle lane other direction, shoulder] representation. This was originally done to workaround traffic simulation assumptions about supporting bidirectional movement on one lane. Now that A/B Street and osm2streets are detangled, though, we can make some progress here independently and do better. So for starters, we can do something else with things like https://www.openstreetmap.org/way/553752370 that're primarily pedestrian spaces that allow bikes.

This is related to #59 and #81. The issue of parallel ways to the main road is still present, but not the focus here. There are plenty of non-parallel examples we could import better today.

dabreegster:
This is now done! There's room for improvement, but no specific followup tasks yet, so closing

PR #88, opened 2022-09-14
Stop inserting fake OSM tags early during importing

dabreegster:
This PR cleans up the import process a bit. Previously, when we make the initial pass over raw OSM data, we'd look at the sidewalk and parking tagging on every road. When tags were missing, based on different bits of config, we'd insert fake tags (including a made-up `abst:sidewalks_inferred = true` to mark the assumption). Now, we leave the OSM tags alone, and instead do this inference right before going through the tags -> lane specification process -- both the current one and the experimental osm2lanes one.

This cleanup helps with the osm2lanes cutover (https://github.com/a-b-street/osm2lanes/issues/248) by making the old and new code-path do the same kind of last-minute tag transformation. This also helps with #87, by grouping strange old assumptions about footway and shared-use paths in less places in the code.

The big behavioral change with this PR is removing some roads we were previously importing as shoulders+bike lanes. All of those cases weren't being represented very well, so removing them before the next step seems fine. Examples are https://www.openstreetmap.org/way/923062091 (a `highway=track`) and https://www.openstreetmap.org/way/539534605 / https://www.openstreetmap.org/way/573276662 (`highway=pedestrian` that definitely shouldn't be a bike lane-focused representation).

I'm regenerating all A/B Street data with this change. I don't expect any new problems after some initial testing, but I'll hold off on merging until things are done there.


import_streets/src/extract.rs
```diff
@@ -125,24 +119,17 @@ impl OsmExtract {
...
-            return false;
-        }
-
-        #[allow(clippy::collapsible_if)] // better readability
-        if (highway == "footway" || highway == "path" || highway == "steps")
-            && opts.map_config.inferred_sidewalks
-        {
-            if !tags.is_any("bicycle", vec!["designated", "yes", "dismount"]) {
-                return false;
+        // If we're only handling sidewalks tagged on roads, skip a bunch of ways
```
dabreegster, line 122: I expect this to go away in my next planned step, which'll hopefully be introducing lane types for footpaths (anything that isn't parallel to a "normal road") and shared-use things (foot+cycle shared spaces)

import_streets/src/extract.rs
```diff
@@ -159,77 +146,7 @@ impl OsmExtract {
             return false;
         }
 
-        // Not sure what this means, found in Seoul.
-        if tags.is("lanes", "0") {
```
dabreegster, line 163: I think I added this to avoid some crash. Just going to remove it as an old hack and if we hit whatever this was again, we can deal with it more properly hopefully

import_streets/src/extract.rs
```diff
@@ -159,77 +146,7 @@ impl OsmExtract {
...
-            return false;
-        }
-
-        if opts.skip_local_roads && osm::RoadRank::from_highway(highway) == osm::RoadRank::Local {
```
dabreegster, line 167: Also cleaning up an old hack that one user in A/B Street needed once. It could be useful to pretend a bunch of roads don't exist, but maybe doing it as some kind of later transformation on `StreetNetwork` would make more sense

street_network/src/lanes/classic.rs
```diff
@@ -384,3 +388,46 @@ fn osm_separation_type(x: &String) -> Option<BufferType> {
         _ => None,
     }
 }
+
+// If sidewalks aren't explicitly tagged on a road, fill them in. This only happens when the map is
+// configured to have this inference.
+pub(crate) fn infer_sidewalk_tags(tags: &mut Tags, cfg: &MapConfig) {
```
dabreegster, line 394: This is the logic removed from `extract.rs`

street_network/src/lanes/osm2lanes.rs
```diff
@@ -175,53 +175,7 @@ fn transform_tags(tags: &Tags, cfg: &MapConfig) -> osm_tags::Tags {
...
-            if tags.is_any(osm::HIGHWAY, vec!["residential", "living_street"])
-                && !tags.is("dual_carriageway", "yes")
-            {
-                tags.insert(osm::SIDEWALK, "both");
-            }
-        } else {
-            tags.insert(osm::SIDEWALK, "both");
-        }
-    }
+    super::classic::infer_sidewalk_tags(&mut tags, cfg);
```
dabreegster, line 178: Code now de-duplicated!

tests/src/borough_sausage_links/geometry.json
```diff
@@ -1005,12 +1005,12 @@
               51.501094561991984
             ],
             [
-              -0.09205544955937478,
```
dabreegster, line 1008: I used the web tool to visually diff. The examples I listed in the PR description explain the changes -- some roads that were imported very incorrectly are now removed. Some will likely return as footpath types soon

BudgieInWA:
Lots of mini cleanups is cool!

dabreegster:
No issues in A/B Street land, so merging this!

Issue #77, opened 2022-08-18
Bicycle road/street rendering


larsschwarz:
This bicycle lane rendering always seems weird to me. Despite the visual representation there also seems an issue with detecting/marking the actual infrastructure itself. 

A few examples:

This is an actual bicycle road with proper tags in OSM: https://www.openstreetmap.org/way/10575838
In A/B street it renders like a usual street/highway. Since it's dedicated use is cycling I'd expect it to be rendered like that (by its primary use).

This path here https://www.openstreetmap.org/way/32916454 on the other hand is just a regular path (no street) where bicycles and pedestrians are allowed, but it renders as bicycle lane in A/B street. For shared uses I'd render this as footpath.

A/B street seems to show everything with bike alloance as type "protected bike lane", which doesn't seem right. 

Nodes with "oneway:bicycle" tagged render the green bicycle lane in A/B street in both directions. Example: https://www.openstreetmap.org/way/102442321

There are also some minor issues like the rendering of streets in a pedestrian zone (highway=pedestrian), these also render as bicycle lanes. Example: https://www.openstreetmap.org/way/17071901

Generally speaking everything bikeable seems to be rendered as "protected bike lane". This might be ok where the bicycle is the only dedicated use, but for shared uses (pedestrians) this seems wrong.

joostschouppe:
Just a minor comment : if a special treatment of bicycle_road=yes were to be implemented, apply this as well to cyclestreet=yes. This tag bundles extremely similar concepts, that do vary from country to country.

Robinlovelace:
>  The goal is to use all the values in https://wiki.openstreetmap.org/wiki/Proposed_features/separation#Typical_separation_values and have a good 2D overhead view for these

That would be amazing and I can see big potential for community contribution and crowd-sourcing designs for these:

- [ ] no – no physical separation/demarcation/protection features (but usually with markings, that can also be specified – see below).

- [ ] bollard – fixed bollards that prevent vehicles from running over them.

- [ ] flex_post – flexible bollards/posts/delineators, often made of plastic, that buckle when vehicles run over them.

- [ ] vertical_panel – larger or smaller traffic panels, sometimes only for temporary separation.

- [ ] dots – small safety devices like [Botts' dots](https://en.wikipedia.org/wiki/Botts%27_dots)/road studs/raised pavement markers that can be easily passed over but give a tactile or auditory feedback.

[...]

PR #90, opened 2022-09-15
Start importing all footways and shared-use paths! #87, #77

dabreegster:
This introduces two new lane types and starts importing from OSM. The experiment for separate sidewalk handling is refined to only deal with parallel sidewalks and crossings. All the footways are imported always.

I'm immensely excited by this PR. Previously, most footways were either not imported at all or mangled into bike lanes. This is a huge step forwards that doesn't get stuck on the problem of parallel sidewalks. Some demos:

Screenshot from 2022-09-15 10-59-05
Screenshot from 2022-09-15 10-58-38
Screenshot from 2022-09-15 10-58-08

I've validated with the test explorer, clicking through to open the OSM ways and spot-check things. I'm next going to update A/B Street with these changes and work through integration issues there.


import_streets/src/extract.rs
```diff
@@ -119,15 +119,14 @@ impl OsmExtract {
             return false;
         }
 
-        // If we're only handling sidewalks tagged on roads, skip a bunch of ways
+        // If we're only handling sidewalks tagged on roads, skip crossings and separate sidewalks
+        // Note we have to do this here -- get_lane_specs_ltr doesn't support decisions like
+        // "actually, let's pretend this road doesn't exist at all"
         if opts.map_config.inferred_sidewalks {
```
dabreegster, line 125: So this experiment now just means "bring in `highway=footway, footway={sidewalk,crossing}` ways". All of the new stuff in this PR works both with and without this experiment

street_network/src/lanes/classic.rs
```diff
@@ -31,33 +34,20 @@ pub fn get_lane_specs_ltr(tags: &Tags, cfg: &MapConfig) -> Vec<LaneSpec> {
...
-        // allowed.
-        if tags.is("bicycle", "no")
-            || (tags.is(osm::HIGHWAY, "footway")
-                && !tags.is_any("bicycle", vec!["designated", "yes", "dismount"]))
-        {
-            return vec![fwd(LaneType::Sidewalk)];
-        }
-        // Otherwise, there'll always be a bike lane.
 
+    // If it's a primarily cycleway, have directional bike lanes and add a shoulder for walking
```
dabreegster, line 38: It's still odd to use two directional lanes for this case. I think `street_network::Direction` and everything downstream needs to have a both-ways case. It's a very complex change downstream, affecting routing, rendering, traffic simulation... so not attempting it yet.

street_network/src/lanes/mod.rs
```diff
@@ -32,6 +32,11 @@ pub enum LaneType {
     Construction,
     LightRail,
     Buffer(BufferType),
+    /// Some kind of pedestrian-only path unassociated with a road
```
dabreegster, line 35: Very related to #77. I'm not happy with this classification as "only pedestrians" and "both pedestrians and cyclists." Things like living streets or alleyways, where motor vehicles are also allowed, aren't captured yet. Maybe we just need one shared use lane type, with a way to specify the legal or cultural priority between vehicles. For instance, I know many streets in the US without pavement, where it's effectively shared use and motor vehicles get priority.

street_network/src/lanes/mod.rs
```diff
@@ -76,17 +83,22 @@ impl LaneType {
...
+        matches!(
+            self,
+            LaneType::Sidewalk | LaneType::Shoulder | LaneType::Footway | LaneType::SharedUse
+        )
     }
 
     pub fn describe(self) -> &'static str {
         match self {
             LaneType::Driving => "a general-purpose driving lane",
-            LaneType::Biking => "a protected bike lane",
```
dabreegster, line 89: "protected" never ever made sense, I erroneously wrote this long ago.

Also this is a localization issue... "cycling lane" or "cycle lane" would fit better in the UK. But anyway.

street_network/src/lanes/mod.rs
```diff
@@ -154,6 +172,8 @@ impl LaneType {
             LaneType::Construction => 'x',
             LaneType::LightRail => 'l',
             LaneType::Buffer(_) => '|',
+            LaneType::Footway => 'f',
+            LaneType::SharedUse => 'V',
```
dabreegster, line 176: There was not an obvious choice here :(

tests-web/www/js/layers.js
```diff
@@ -66,6 +66,25 @@ export const makeLanePolygonLayer = (text) => {
 
   return new L.geoJSON(JSON.parse(text), {
     style: function (feature) {
+      if (feature.properties.type == "Footway") {
```
dabreegster, line 69: The styling is not great. Any ideas for distinguishing the two?

street_network/src/lanes/mod.rs
BudgieInWA: I think when it comes down to it, every lane is a list of specific designations and restrictions. There are also classifications, informed by the locale, I suppose, like "only pedestrians", "bike lane" or "bus lane (bikes allowed)" that inform markup decisions.

street_network/src/lanes/mod.rs
BudgieInWA: Large '`F`'? It is appropriate that is suggests footway, I think.

street_network/src/lanes/classic.rs
BudgieInWA: Yes, I agree. An "alternates" case too.

Most reads of the value will boil down to "is travel supported in this direction", right? Routing could implement "alternating" and use that for "both ways" too.

I called it [`TravelDirection` in my types](https://github.com/a-b-street/osm2streets/blob/a19ab6ac8ef1dc4b812830fc1eae11687e28356c/osm2streets/src/units.rs#L22-L33).

tests-web/www/js/layers.js
BudgieInWA: Maybe it's time to start copying design from the [strassenraumkarte-neukoelln](https://github.com/SupaplexOSM/strassenraumkarte-neukoelln) :P

street_network/src/lanes/mod.rs
dabreegster: I like this way of thinking about it. Going to start a fresh issue on this soon to consolidate all of the thoughts around this...

tests-web/www/js/layers.js
dabreegster: I already lifted the dashed outline idea and path color from there ;)

dabreegster:
https://github.com/a-b-street/abstreet/pull/994 has more rendering demo. Mostly smooth sailing there, I expect to finish ironing out some integration problems and merge both in a few hours!

Issue #67, opened 2022-08-16
Turns / movements


dabreegster:
It feels like we're at a point to start figuring out how to represent turns / movements through intersections in osm2streets. (I've gone through this exercise before for A/B Street and made some questionable choices there, based on assumptions about how lane-changing should work in a traffic simulation. I'm viewing this as an opportunity to try again, with fresh eyes + experience + more discussion!)

## How A/B Street does it: basic schema

A [turn](https://github.com/a-b-street/abstreet/blob/master/map_model/src/objects/turn.rs) is uniquely identified as a (source lane, destination lane, intersection). The intersection seems redundant, but it's needed to talk about pedestrian movements. https://a-b-street.github.io/docs/tech/dev/formats/traffic_signals.html explains why.

Each turn has a type. For vehicles, it's straight / right / left / u-turn. There can be more nuance than that (5-way intersections, slight right). Distinguishing these types was partly helpful in writing heuristics to generate traffic signal configurations -- things like saying "at a 4-way intersection, the straight turns for the north & south roads can all go. The left turns have to yield." But defining these based on relative road angles was always brittle and hit edge cases, and I remember trying to come up with an alternate way to think about it... in a left-handed driving place, a left turn is one that conflicts with less turns than a right turn.

Turns conflict very "coarsely." A turn has a line-string of the movement through the intersection, and if those intersect, then they conflict. I think I've only needed the concept of conflicting turns in traffic signal configurations and traffic sim, but it may be a concept more widely useful.

## Consequences

Defining turns at the lane level gets kind of weird.
![Screenshot from 2022-08-16 12-26-54](https://user-images.githubusercontent.com/1664407/184868752-73520838-c153-4304-9ced-00dff8bdf71e.png)

Changing lanes in the middle of an intersection is illegal (I think in the US, possibly everywhere, whatever). But from this left lane, A/B Street has a straight turn going to either lane. I did it this way to avoid the absolute mess of simulating lane-changing, but it led to other problems in the simulation anyway. There has to be a way to penalize changing lanes like this, so by default vehicles only do it if necessary.

Sometimes 3 lanes lead to 2 and someone has to merge though. Or vice versa, one turn lane opens up onto 3 lanes, and there's a choice. Something has to say what turns are legal, and maybe prioritize them.

## Lane types

Things get weirder with roads that have bus or bike lanes. A/B Street looks like this:
[...]

PR #94, opened 2022-09-19
Detect a few MultiConnections

BudgieInWA:
This PR detects `MultiConnection`s made up of oneways (like a divided highway entrance or exit). It is a tiny improvement, itself, but mostly I want to talk about the approach.

The idea is to look at the entrances and exits of the intersection to see to what extent paths through the intersections conflict. Counting the number of roads pointing in and out seemed like an OK starting point, with additional heuristics added on top. But, it turns out classifying intersections first, in order to understand the intersection better in later transformations, falls down quickly. 

I am now thinking that the calculating the valid "movements" (as A/B Street calls them) should come first. Then the conflicts could be calculated (at the graph level, without geometry, just ordering), then the types of markings that must be present.

### DIFF
--- a/street_network/src/transform/classify_intersections.rs
+++ b/street_network/src/transform/classify_intersections.rs
@@ -1,6 +1,6 @@
 use crate::osm::NodeID;
 use crate::IntersectionComplexity::*;
-use crate::{Intersection, IntersectionComplexity, OriginalRoad, StreetNetwork};
+use crate::{Direction, Intersection, IntersectionComplexity, OriginalRoad, StreetNetwork};
 
 /// Determines the initial complexity of all intersections. Intersections marked "Crossing" are
 /// considered "unclassified" and will be updated with a guess, others will be left unchanged.
@@ -8,10 +8,7 @@ pub fn classify_intersections(streets: &mut StreetNetwork) {
     let mut changes: Vec<(NodeID, IntersectionComplexity)> = Vec::new();
     for (id, inter) in &streets.intersections {
         if let Crossing = inter.complexity {
-            changes.push((
-                *id,
-                guess_complexity(inter, streets.roads_per_intersection(*id)),
-            ));
+            changes.push((*id, guess_complexity(streets, id)));
         }
     }
 
@@ -23,7 +20,9 @@ pub fn classify_intersections(streets: &mut StreetNetwork) {
 /// Guesses the complexity of the intersection based on the connecting roads and their lanes.
 ///
 /// The existing complexity field is ignored, so be careful how you use the guessed value.
-fn guess_complexity(_inter: &Intersection, roads: Vec<OriginalRoad>) -> IntersectionComplexity {
+fn guess_complexity(streets: &StreetNetwork, id: &NodeID) -> IntersectionComplexity {
+    let roads = streets.roads_per_intersection(*id);
+
     // A terminus is characterised by a single connected road.
     if roads.len() == 1 {
         return Terminus;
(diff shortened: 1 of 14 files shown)


street_network/src/transform/classify_intersections.rs
```diff
@@ -34,8 +33,50 @@ fn guess_complexity(_inter: &Intersection, roads: Vec<OriginalRoad>) -> Intersec
         return Connection;
     }
 
-    // A MultiConnection is characterised by exactly two connected dividing lines (lines that
-    // separate traffic in different directions) and no traffic that crosses them.
+    // A MultiConnection is characterised by exactly one dividing line traveling through it (the
+    // line that separates traffic in different directions) and no traffic that crosses it.
+    if roads.len() == 3 {
```
dabreegster, line 38: Check out https://www.openstreetmap.org/node/608494028 in the `arizona_highways` test. This feels like a case too -- one main uninterrupted road, two ramps merging in on either side

street_network/src/transform/classify_intersections.rs
```diff
@@ -34,8 +33,50 @@ fn guess_complexity(_inter: &Intersection, roads: Vec<OriginalRoad>) -> Intersec
...
+        let mut num_roads_in = 0;
+        let mut num_roads_out = 0;
+        let mut num_roads_inout = 0;
+        for or in roads {
```
dabreegster, line 42: Code style: from @michaelkirk I picked up the good habit of trying to spell out longer names, unless the abbreviation was very standard or used frequently in the codebase. Could we try that here? Is this `outward_road`? (Which would be a weird name, because we determine below if it's even outward or not

street_network/src/transform/classify_intersections.rs
```diff
@@ -34,8 +33,50 @@ fn guess_complexity(_inter: &Intersection, roads: Vec<OriginalRoad>) -> Intersec
...
+        }
+
+        if num_roads_inout == 0 {
+            // The simplest MultiConnect
+            return MultiConnection;
+        }
+
+        // detect the simple case of a single road splitting
+        if num_roads_inout == 1 {
+            // TODO Determine if it is possible to turn from the 'in' road to the 'out' road.
```
dabreegster, line 72: Screenshot from 2022-09-20 11-46-44
`aurora_sausage_link` is maybe a good simple example to look at...

pxl_20220920_035003961_720
These are the movements possible, just from the cartesian product of every inbound road going to every outbound. In the T intersection case, two of the movements conflict, because they have the same destination (the south piece of road). Don't need any geometry to decide that.

In the dual carriageway split case, there's just a northbound or southbound movement. There is a U-turn, but it's explicitly banned in OSM, so this should show up in the `turn_restrictions: Vec<(RestrictionType, OriginalRoad)>` list.

dabreegster (commented):
> I am now thinking that the calculating the valid "movements" (as A/B Street calls them) should come first. Then the conflicts could be calculated (at the graph level, without geometry, just ordering), then the types of markings that must be present.

I'm in 100% agreement here, after working through the aurora example. Just counting in/outbound roads doesn't seem like it can distinguish those examples

street_network/src/transform/classify_intersections.rs
dabreegster: pxl_20220920_035821523_720
Hmm, the three movements conflict in the sense that they all have the same destination. So maybe we do need geometry to distinguish this from the T intersection case?

... I'm pretty sure you mentioned an idea about looking at roads in clockwise order and seeing what adjacent pairs look like. Does that help us at all maybe? Or do we need to generate the polyline of the movement and look for intersections?

dabreegster: https://github.com/a-b-street/osm2streets/issues/67#issuecomment-1251012734
Ah, you're way ahead of me

street_network/src/transform/classify_intersections.rs
BudgieInWA: I have a habit of using short names for variables that are short lived over only a few lines, but I appreciate the push towards less obtuse naming. This case is bad no matter how you look at it though, I will change it to `road_id`.

dabreegster:
Also if we want to merge this as a small step forwards, please rebase and fix a few of the unused import warnings please

street_network/src/transform/classify_intersections.rs
BudgieInWA: Yes, I had found this example too, which is great for having both cases right next to each other!

I was going to wait until Movements were represented in some consistent way before progressing on this, but I will keep going using just `turn_restrictions` as a stand in for the full movements that can be fleshed out later.

PR #96, opened 2022-09-22
Start calculating Movements and Conflicts

BudgieInWA:
still faulty due to road sort order and missing travel direction checks.

See #67. We should be able to see on and off ramps, sausage links, etc. pretty clearly in the result when its working.


street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
 pub fn classify_intersections(streets: &mut StreetNetwork) {
-    let mut changes: Vec<(NodeID, IntersectionComplexity)> = Vec::new();
+    let mut changes: Vec<(NodeID, (IntersectionComplexity, ConflictType))> = Vec::new();
     for (id, inter) in &streets.intersections {
         if let Crossing = inter.complexity {
             changes.push((*id, guess_complexity(streets, id)));
         }
     }
 
     for (id, complexity) in changes.into_iter() {
```
dabreegster, line 19:
```suggestion
    for (id, (complexity, conflict_level)) in changes {
```

`into_iter` happens automatically for `Vec`s, and we can do more tuple matching/destructuring here

street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
-                        _num_roads_in += 1
-                    } else {
-                        _num_roads_out += 1
-                    }
-                }
-                None => num_roads_inout += 1,
+    // Calculate all the possible movements, except (U-turns).
+    //FIXME assert!(roads is sorted clockwise), which it isn't
+    let mut connections = Vec::new();
+    // Consider turns pairs of roads, from s to d, using their position as index, so we can them later.
```
dabreegster, line 53:
```suggestion
    // Consider turns pairs of roads, from s to d, using their position as index, so we can reference them later.
```

street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
-    let roads = streets.roads_per_intersection(*intersection_id);
+fn guess_complexity(
+    streets: &StreetNetwork,
+    intersection_id: &NodeID,
+) -> (IntersectionComplexity, ConflictType) {
+    use ConflictType::*;
+    let road_ids = streets.roads_per_intersection(*intersection_id);
+    let roads: Vec<&Road> = road_ids
+        .iter()
+        .map(|id| streets.roads.get(id).unwrap())
```
dabreegster, line 37: For a btreemap, we can just do `&streets.roads[id]` equivalently, which'll crash if the key isn't there

street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
-                    } else {
-                        _num_roads_out += 1
-                    }
-                }
-                None => num_roads_inout += 1,
+    // Calculate all the possible movements, except (U-turns).
+    //FIXME assert!(roads is sorted clockwise), which it isn't
+    let mut connections = Vec::new();
+    // Consider turns pairs of roads, from s to d, using their position as index, so we can them later.
+    for s in 0..road_ids.len() {
```
dabreegster, line 54: It might be easier to just store `&Road` references instead of the indexing and separate `road_ids` / `roads`. This whole method is `&StreetNetwork` read-only, so it's impossible for a pointer to change

street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
             }
         }
     }
+    !has_exclusive_allows
+}
 
-    // A Merge is characterised by an uninterrupted road (that would qualify as a Connection or
-    // MultiConnection), with additional connected roads that yield.
-    // TODO Combine roads into corridors and count them.
+fn calc_conflict(a: &(usize, usize), b: &(usize, usize)) -> ConflictType {
```
dabreegster, line 107: The types are getting a little weird here, so I think passing around `&Road` pointers would indeed more clear. These're connections between two roads, and we're just looking to see if identity of the roads is the same

dabreegster: Edit: never mind, `is_between` is using the ordering of the roads around each other. I understand now

street_network/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,128 @@
...
+    }
+    if is_between(a.0, b) ^ is_between(a.1, b) {
+        return Cross;
+    }
+    return Uncontested;
+}
 
-    return Crossing;
+fn is_between(num: usize, range: &(usize, usize)) -> bool {
+    let bot = min(range.0, range.1);
```
dabreegster, line 125: We're reasoning about a cyclic group here, since it's roads sorted clockwise and there's an arbitrary start/end. IIUC, the simple case is range.0 < num < range.1... if range.0 is < range.1. Argh actually I just need to draw this to understand. Are you confident the logic here is correct? :P  Maybe we can add unit tests if we're unsure

street_network/src/types.rs
```diff
@@ -107,6 +107,14 @@ pub enum InterruptionType {
     DeadEnd,
 }
 
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum ConflictType {
```
dabreegster, line 111: How about a docstring describing what this does? I almost read it as describing how two movements interact or not, but I think it's used to talk about the intersection as a whole

street_network/src/transform/classify_intersections.rs
dabreegster: (Never mind, ordering around the intersection is why the indices are important)

street_network/src/transform/classify_intersections.rs
BudgieInWA:
```suggestion
    // Consider turns pairs of roads, from s to d, using their position as index, so we can compare them later.
```

street_network/src/transform/classify_intersections.rs
BudgieInWA: Yes, the positions are important. The road_ids / road distinction is because a `Road` doesn't know it's own id.

street_network/src/transform/classify_intersections.rs
BudgieInWA:
```rust
type Position = usize;
type ConnectedPositions = (usize, usize);
```
?

street_network/src/transform/classify_intersections.rs
BudgieInWA: I've taken the cyclic group and one connection. The group is like a circle with the connection running through the middle joining a pairs of points. Cut the circle somewhere to arbitrarily define the start/end of our reference list, i.e. where zero is. Note that any point on the circle is either between the pair of points, or between one point and the start/end. For an additional connection to cross the existing connection, one end would have to be between the pair, and one outside. Hence `if is_between(a.0, b) ^ is_between(a.1, b)`. `is_between` is over ordinary numbers.

I don't know if I followed it through accurately, and seeing it all come together under tests will be one way to find out, but that was my idea.

BudgieInWA: Zero at `O`, connection `A` to `B`. Consider starting a comnection from `X` or `Y`

street_network/src/initial.rs
```diff
@@ -56,6 +58,7 @@ pub struct Intersection {
     pub polygon: Polygon,
     pub roads: BTreeSet<OriginalRoad>,
     pub complexity: IntersectionComplexity,
+    pub conflict_level: ConflictType,
```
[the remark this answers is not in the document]
BudgieInWA: I only add it because i need to reference it when "rendering" to geojson. Maybe i should be more creative about chasing down the value for that use case

street_network/src/render.rs
```diff
@@ -49,7 +50,15 @@ impl StreetNetwork {
                     ("osm_node_id", id.0.into()),
                     (
                         "complexity",
-                        format!("{:?}", intersection.complexity).into(),
+                        if intersection.complexity == MultiConnection {
```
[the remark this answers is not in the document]
BudgieInWA: Mostly, but so far `Terminus | Connection => Uncontested` and `Crossing => Cross`, so this keeps the changes to the test output relevant for now. (I have not committed the tests, which are failing, so the diff will remain enlightened as I go.)

A new JSON field is likely the best option in the end.

street_network/src/types.rs
BudgieInWA: Yep :P 

It does describe how two particular movements interact, but `Intersection` `conflict_level` is using that distinction to describe what types of conflict one should expect to find.

street_network/src/render.rs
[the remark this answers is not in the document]
dabreegster: Another option is being more precise with the possibilities..

```
enum Complexity {
  Terminus,
  Connection,
  Crossing(ConflictType),
}
```

since the conflict is unambiguous for some cases. But then if some piece of code needs to reason about the special cases, this might be more awkward

street_network/src/transform/classify_intersections.rs
dabreegster: Eh, if you want. If the `usize`s leaked outside this file, a typedef at least would be nice. I think a single comment near the top of this method could kind of explain what we're working with here, and that would suffice

street_network/src/transform/classify_intersections.rs
dabreegster: Thank you! I understand the logic better now and think this is correct... but yeah, end-to-end tests will be the easiest to flush out problems anywhere

street_network/src/render.rs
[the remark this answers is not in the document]
BudgieInWA: I had both those thoughts too. I think always having it around, and probably also always calculating it, is simpler. Our assumptions are bound to be wrong in some ways, too.

osm2streets/src/render.rs
```diff
@@ -252,6 +261,45 @@ impl StreetNetwork {
...
+                    } else {
+                        pl.last_pt()
+                    };
+                    pt
+                })
+                .collect();
+            for (a, b) in intersection.movements.iter() {
+                if *a != *b {
+                    pairs.push((
+                        Line::must_new(road_points[*a], road_points[*b])
```
BudgieInWA, line 289: Oh, this can fail if the points are too close together... I find it surprising that a line type can't even represent lines segments of length zero... I guess that would mean that it has no direction, and so a whole host of operations on it would need to be fallible...

I'll use the constructor that returns a result instead.

BudgieInWA:
@dabreegster I finally found the spare energy to get this functioning at a draft level. Movements are drawn crudely as arrows between the Road endpoints, and some merges vs diverges are detected.

(screenshot incoming...)

Still to do:

- [ ] actually look into the results and new test output to see if it's working right
- [ ] add a heuristic to identify movements that are U-turns (for the suasage-link case). Tagging convention assumes they are easily identified, so we should be able to find prior work.
- [ ] filter out roads that don't have car traffic at all (they should be considered in a separate pass)
- [ ] recalculate movements when roads are added/removed
- [ ] offset the arrow endpoints for bidi roads closer to the center of the indicated direction
- [ ] document `ConflictType`
- [ ] address a TODO or two
- [ ] tidy up some `use` statements

dabreegster:
Awesome! I'm having a look now. Do you want high- or low-level feedback now, or after those TODOs done?

BudgieInWA:
Any high level thoughts now. I'll request a review when I'm ready for it to go in.

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,205 @@
...
+    let mut connections = Vec::new();
+    // Consider all pairs of roads, from s to d. Identify them using their index in the list - which
+    // is sorted in clockwise order - so that we can compare their position later.
+    for s in 0..road_ids.len() {
+        for d in 0..road_ids.len() {
+            if s == d {
+                continue; // Ignore U-turns.
+            }
+
+            // Calculate if it is possible to emerge from s into the intersection.
```
dabreegster, line 62: Feels related to #1 and the idea of road halves. In A/B Street, we have a `DirectedRoadID`, which is a road ID and a forwards/backwards direction. I wonder if that'd be useful here too? I started thinking through a method for a road to see which directions point incoming or outgoing to an intersection and return the results in a clockwise order, but... didn't come up with anything clear

dabreegster (commented):
I haven't dove into the logic yet, but just reporting a bug in `arizona_highways`:
Screenshot from 2022-11-07 17-02-19

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: `DirectedRoadID` is a concept that would be useful here, I agree. It feels like it would be more ergonomic if an intersection stored `DirectedRoadID` for all connected roads in a consistent direction, so that we don't have to do the `.i1 == id` dance.

I wonder if it would be even more ergonomic to have a type that is a "view" of a road that supports inverting the direction. That is, if I want to get the list of roads connected to an intersection, it would be nice to treat them all as if they were all incoming roads that "end" at the intersection. The idea is to "hoist" one of the boolean attributes up out of the scope, so the algorithm has half as many cases to contend with: all roads are incoming, so I don't have to branch every time I want to get the "end of the road that meets the intersection" or anything like that.

For example, it would be cool if I could do something like the psuedo code that follows. `roads_clockwise` would return an iterator of `RoadView` instances that - throughout its API - reverses the meaning of "start"/"end" and "forward"/"backward" relative to the underlying `Road`.

```rust
for road in intersection.roads_clockwise() {
  print("road {} meets the intersection at {}", road.directed_id, road.get_end_point());
}
```

---

The "half road" idea leads me to a different API that would be even more ergonomic in this case. A `HalfRoad` could be a view into an underlying `Road` that is always oriented in the direction of travel.

```rust
for (s, half_road_s) in intersection.half_roads_clockwise().enumerate() {
  if let Incoming(src) = half_road_s {
    for (d, half_road_d) in intersection.half_roads_clockwise().enumerate() {
      if let Outgoing(dst) = half_road_d {
        if turn_is_allowed(src, dst) {
          movements.add(s, d);
        }
      }
    }
  }
}
```

The logic for determining crossing/merging/diverging becomes a lot simpler too, because every position around the intersection boundary (`s` and `d`) is just one direction, so we wouldn't have to factor in the driving side. (The driving side is the boolean attribute that has been hoisted in this case).

--- 

I suspect we will need to discuss this synchronously to get to a full understanding...

dabreegster: > That is, if I want to get the list of roads connected to an intersection, it would be nice to treat them all as if they were all incoming roads that "end" at the intersection. 

The intersection geometry algorithm does exactly that too! It's a really useful process, so it definitely ought to be easier.

> A HalfRoad could be a view into an underlying Road that is always oriented in the direction of travel.

I'm really liking this idea! Depending what fields/methods this view has, maybe this is straightforward to implement, or maybe we attempt something with lifetimes. (A `HalfRoad` internally stores a pointer to a `RawRoad`, so we write out explicit lifetimes to keep everything safe.

Happy to chat about this soon! I have plenty of time this week

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
+    // Therefore, if `a` is drawn pointing upwards from low .0 to high .1,
+    // then LHT would be crossed by movements joining from the "inside".
+    //     ______          ______
+    //    /  ^:  \        /  :|  \
+    //   |  a|:   n      |   :|   n
+    //   |   |:   0      |   :|a  0
+    //    \__|:__/        \__:V__/
+
+    // This equation (hopefully) works. Once it does, just trust it:
+    // TODO unit test these three equations.
```
BudgieInWA, line 201: The results look pretty good, but I found at least one error, which I think comes from these equations. northgate_dual_carriageway [intersection 95600603](https://a-b-street.github.io/osm2streets/?test=northgate_dual_carriageway#20.97/47.70805/-122.32864) should be a Merge, but is classified as a Cross.

Considering that I'd prefer to work with `HalfRoad`s instead, I might delay unit testing and figuring this out properly.

osm2streets/src/render.rs
```diff
@@ -240,6 +251,65 @@ impl StreetNetwork {
...
+            Distance::meters(-1.3)
+        } else {
+            Distance::meters(1.3)
+        };
+
+        let mut pairs = Vec::new();
+
+        for (i, intersection) in &self.intersections {
+            // Find the points where the arrows should (leave, enter) the roads.
+            let road_points: BTreeMap<_, _> = intersection
```
BudgieInWA, line 267: I wanted to implement a method on `Road` that could calculate these offset endpoints (for a given end and direction). Because a `Road` doesn't know it's own `DrivingSide` or dividing line (!!), it's non-trivial, so I'm happy with this currently.

I want to figure out how a `Road` should figure out it's own driving side, dividing line and `HalfRoad`s pretty soon. Maybe annotating contraflow lanes within `LaneSpec` would make it easy enough to calculate on demand from `lanes_lrt`, or maybe the driving side and dividing lane should be stored on the `Road` as part of the OSM parsing process (where we have the most context).

osm2streets/src/lib.rs
```diff
@@ -86,21 +87,36 @@ impl StreetNetwork {
         for i in [id.i1, id.i2] {
             self.intersections.get_mut(&i).unwrap().roads.push(id);
             self.sort_roads(i);
+            // Recalculate movements and complexity.
+            let (complexity, conflict_level, movements) =
+                crate::transform::classify_intersections::guess_complexity(self, &i);
```
BudgieInWA, line 92: This sort of thing challenges my understanding of what should be a "transformation step" and what is an inherent part of the data types. Or at least which transformations are responsible for calculating data (that therefore need to be rerun when later transformations want to change things, such as merging intersections).

The "calculate change first, then commit them all" pattern that is forced upon us by the borrow checker in our current transformation implementations might actually be a better idea than calling this directly whenever a road is added. That is, maybe the responsibility for recalculating the movements etc. should be higher up than `insert_road`. At the level of `merge_intersections(a, b)` or higher.

Anyway, I've got a lot to understand yet about how the transformations fit together...

BudgieInWA (commented):
@dabreegster I'm ready for you to take a look at the implementation, and consider merging this in.

The `complexity` and `conflict_level` fields themselves might end up not actually being used in any calculations / rendering decisions, but they are useful now for debugging and for the tests. It's more likely to be the Major/Minor movement distinction, which I'm starting to work through in my head atm.

osm2streets/src/lib.rs
```diff
@@ -86,21 +87,36 @@ impl StreetNetwork {
         for i in [id.i1, id.i2] {
             self.intersections.get_mut(&i).unwrap().roads.push(id);
             self.sort_roads(i);
+            // Recalculate movements and complexity.
```
dabreegster, line 90: Since we do this twice and it's non-trivial, could we split out a helper method similar to `sort_roads`? Maybe `fn recalculate_movements`

osm2streets/src/lib.rs
```diff
@@ -86,21 +87,36 @@ impl StreetNetwork {
...
+            let (complexity, conflict_level, movements) =
+                crate::transform::classify_intersections::guess_complexity(self, &i);
+            let int = self.intersections.get_mut(&i).unwrap();
+            if int.complexity != IntersectionComplexity::MapEdge {
```
dabreegster, line 94: This feels weird until I read the `guess_complexity` docstring. Feels like it's worth a comment here too (which we can write once if this bit happens in a helper

osm2streets/src/lib.rs
dabreegster: > This sort of thing challenges my understanding of what should be a "transformation step" and what is an inherent part of the data types

I don't have clear ideas here either. :(  I'm going to try to draw the flow chart of how everything gets created / set currently. The big hammer is to treat this as a big dataflow-type problem and consider automatically recalculating derived state when something changes. Or not do that automatically, but at least have a principled way to organize / reason about this. The problem is that changes percolate to related objects. If you do anything to a road, you need to recalculate geometry for two intersections. That calculation might trim back other roads one hop away. If there are transformations that look at trimmed road length, would we want to always go trigger them again? It gets confusing very quickly...

> maybe the responsibility for recalculating the movements etc. should be higher up than insert_road. At the level of merge_intersections(a, b) or higher.

I like this way of thinking. Sometimes a very low-level, "surgical" set of mutations can happen. Or we use this "medium level" of abstraction. But another transformation may just want to say "merge short road" and automatically recalculate a bunch of derived state.

osm2streets/src/render.rs
```diff
@@ -240,6 +251,65 @@ impl StreetNetwork {
...
+                                    .pt1(),
+                                first_road_segment
+                                    .shift_either_direction(-arrow_shift_dist)
+                                    .pt1(),
+                            )
+                        },
+                    )
+                })
+                .collect();
+            for (a, b) in intersection.movements.iter() {
```
dabreegster, line 295:
```suggestion
            for (a, b) in &intersection.movements {
```

osm2streets/src/render.rs
```diff
@@ -240,6 +251,65 @@ impl StreetNetwork {
         let output = serde_json::to_string_pretty(&obj)?;
         Ok(output)
     }
+
+    /// For an intersection, show all the movements.
+    pub fn debug_movements_geojson(&self) -> Result<String> {
+        let arrow_shift_dist = if self.config.driving_side == DrivingSide::Right {
```
dabreegster, line 257:
```suggestion
        // Multiple arrows with endpoints on one road would overlap and look cluttered. Offset them to help.
        let arrow_shift_dist = if self.config.driving_side == DrivingSide::Right {
```

I had to play around with this param to understand the purpose

osm2streets/src/render.rs
dabreegster: I agree it's annoying how often `DrivingSide` has to be plumbed around. Just so I understand, by diving line, you mean the point where the direction of travel changes (or the left/right edge on one-way roads)? Looking at pairs of `lanes_ltr` and direction should work for that, right? And that process exposes cases where there isn't one obvious dividing lane (bidi cycletracks, shared center turn lanes)

> Maybe annotating contraflow lanes within LaneSpec

What do you mean by this? They have `Direction` already

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
+use std::cmp::{max, min};
+use std::collections::BTreeMap;
 
 /// Determines the initial complexity of all intersections. Intersections marked "Crossing" are
 /// considered "unclassified" and will be updated with a guess, others will be left unchanged.
 pub fn classify_intersections(streets: &mut StreetNetwork) {
-    let mut changes: Vec<(NodeID, IntersectionComplexity)> = Vec::new();
+    let mut changes: Vec<_> = Vec::new();
     for (id, inter) in &streets.intersections {
         if let Crossing = inter.complexity {
```
dabreegster, line 16:
```suggestion
        if inter.complexity == Crossing {
```

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
 /// The existing complexity field is ignored, so be careful how you use the guessed value.
-fn guess_complexity(streets: &StreetNetwork, intersection_id: &NodeID) -> IntersectionComplexity {
-    let roads = streets.roads_per_intersection(*intersection_id);
+pub fn guess_complexity(
+    streets: &StreetNetwork,
+    intersection_id: &NodeID,
+) -> (IntersectionComplexity, ConflictType, Vec<Movement>) {
+    use ConflictType::*;
+    let roads: Vec<_> = streets
+        .roads_per_intersection(*intersection_id)
```
dabreegster, line 38: I was going to propose this is another cleanup PR... what do you think about `roads_per_intersection` returning `Vec<&Road>`? Many places need to do something like this. Those that just need road IDs can use the `roads` field that's now just stored in the struct

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
+        .iter()
+        .map(|id| &streets.roads[id])
+        .filter(|road| road.is_driveable())
+        .collect();
 
     // A terminus is characterised by a single connected road.
     if roads.len() == 1 {
-        return Terminus;
+        return (Terminus, Uncontested, Vec::new());
```
dabreegster, line 46: I don't remember if we've talked about this before or not, but `Terminus Cross` is impossible, as are many others. Maybe the `IntersectionComplexity` enum cases should only embed `ConflictType` when it's possible? No need to do that in this PR, just an idea for a possible cleanup

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
-                    }
-                }
-                None => num_roads_inout += 1,
+            // Calculate if it is possible to leave the intersection into d.
+            let dst_road = roads[d];
+            if !can_drive_into(dst_road, *intersection_id) {
+                continue;
+            }
+
+            // TODO detect U-Turns that should be assumed forbidden.
```
dabreegster, line 71: These conditions sound necessary + sufficient to me! Happy to leave this for a followup PR

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
-            // The simplest MultiConnect
-            return MultiConnection;
+    // Calculate all the collisions.
+    let mut conflicts = BTreeMap::new();
+    let mut worst_conflict = Uncontested;
+    // Compare every pair of connections. Use the order of the roads around the intersection to
+    // detect if they diverge, merge, or cross.
+    // assert!(connections is sorted) so small_con large_con makes sense.
+    let mut each_con = connections.iter();
+    while let Some(small_con) = each_con.next() {
```
dabreegster, line 91: Just so I understand: conflict is symmetric, so we only need to calculate "half the matrix" of pairs, right? What do you mean by sorted / small and large connections? Literally the tuples of movement indices, ie, (1, 3) < (2, 1)?

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
 /// Guesses the complexity of the intersection based on the connecting roads and their lanes.
 ///
 /// The existing complexity field is ignored, so be careful how you use the guessed value.
-fn guess_complexity(streets: &StreetNetwork, intersection_id: &NodeID) -> IntersectionComplexity {
-    let roads = streets.roads_per_intersection(*intersection_id);
+pub fn guess_complexity(
+    streets: &StreetNetwork,
+    intersection_id: &NodeID,
+) -> (IntersectionComplexity, ConflictType, Vec<Movement>) {
+    use ConflictType::*;
```
dabreegster, line 36: nit: Do this once at the top instead

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
             }
         }
     }
+    !has_exclusive_allows
+}
+
+fn calc_conflict(a: &(usize, usize), b: &(usize, usize), side: DrivingSide) -> ConflictType {
+    use ConflictType::*;
+
+    // If the traffic starts of ends at the same place in the same direction...
```
dabreegster, line 164:
```suggestion
    // If the traffic starts and ends at the same place in the same direction...
```

osm2streets/src/transform/classify_intersections.rs
```diff
@@ -1,85 +1,229 @@
...
+        return Merge;
+    }
 
-    // A Merge is characterised by an uninterrupted road (that would qualify as a Connection or
-    // MultiConnection), with additional connected roads that yield.
-    // TODO Combine roads into corridors and count them.
+    // The intersection has a boundary that we have labelled 0 to n-1 in clockwise order (from an
+    // arbitrary point), like a string laying in a circle. If we represent `a` as an arc from one
+    // point on the string to another, then there is a section of the string between the two points,
+    // connecting them the two points and two ends of string "on the outside". A second arc, `b`,
```
dabreegster, line 178:
```suggestion
    // connecting them and two ends of string "on the outside". A second arc, `b`,
```

(I'm nitpicking here because I'm reading stuff super carefully to make sure I understand! The diagrams make it easy to follow along)

osm2streets/src/transform/classify_intersections.rs
dabreegster: I got kind of lost around here. I'm fine with leaving this alone for now and iterating later, likely with `HalfRoads`.

Also illegal u-turns being generated right now cause some unexpected results in `aurora_sauage_link`, but that'll go away with a fix for that

dabreegster (approved):
I'm fine merging this in and making smaller incremental steps. We have enough ideas for refactors and new abstractions, and rebasing a long-lived PR gets tough. As long as nothing external using osm2streets currently gets affected, I prefer checking in experiments and leaving the results unused or flagging them off

Awesome leap forward!

osm2streets/src/lib.rs
BudgieInWA: Good call. Maybe `guess_complexity` should be the place that respects `MapEdge`... I'll keep that in mind as we learn more about what `complexity` is actually used for and where movement calculations fit into the overall flow.

osm2streets/src/lib.rs
BudgieInWA: > Or not do that automatically, but at least have a principled way to organize / reason about this [big dataflow-type problem].

This seems the likely best approach to me. At least for the code that takes the raw input to the naively "filled out" street representation.

My idea is that when you call `StreetNetwork::new(osm_stuff)` you might expect to get back `Road`s and `Intersection`s that are complete, but blindly representing the raw data. There would be a handful of low-level conditions that are met ("all the geometry connects up properly to each other..."), and the other higher level conditions haven't been addressed yet ("... but some of it overlaps"). These are the bits of code that don't feel like "transformations" to me, more like "hydrating" the abstract raw descriptions ("this centerline with this placement and these lanes") into a useful concrete representation ("these lanes covering these areas"). 

Because the raw data is criminally under-specified we are expected to infer a bunch of stuff; under-represented so we want to guess/fill in a bunch of stuff; and imprecise so we want to correct a bunch of stuff. For example, if all intersection-internal ways were tagged `junction=intersection` (or whatever the tag is) then we wouldn't need a transformation to merge intersections together, we'd simply collect all the connected intersection nodes together before creating the `Intersection`.

These inferences, guesses and corrections are easier to make with the *hydrated* representation, so it makes sense to *hydrate* the raw data, calculate any inferences, propagate them back into the raw data then *hydrate* again. This is how I see "transformations" fitting in conceptually.

Actually modifying the osm data as the result of transformations sounds like a painful idea, but it feels like there's something there: `StreetNetwork`/`Road`/`Intersection` are like primitives, with some source-of-truth data (placement adjusted centerlines, connectivity, stop line positions) and some straightforwardly derived data (geometry, etc.). If `StreetNetwork` provided a rich enough set of operations (like collapsing a road (merging intersections), collapsing an intersection (merging roads), zipping/snapping adjacent roads, adding turn restrictions, moving centerlines/stop line positions, that sort of thing) then it could be used to manually draw beautifully accurate street networks (or a horrible mess). The problem of getting a beautiful street network out of raw osm data then becomes one of naively translating the osm spec into `StreetNetwork` primitives, then doing bunch of tweaking operations on them using osm-specific inferences/assumptions.

The `StreetNetwork` primitive operations could be efficiently implemented, maybe even in a way that allows batching (perform a bunch of operations and keep track of what is invalidated, then recalculate all the invalidated data). Then we could reason about the order of transformations, if they need repeating, even of multiple transformations can be committed at the same time.

I'm gonna let that marinate in my brain for a little while and look to talk with you about it later.

osm2streets/src/render.rs
BudgieInWA: Yes, that is what I mean by "dividing line". If contraflow lanes were annotated as "contraflow" from the beginning, then we wouldn't get confused when trying to find the dividing line. Knowing the driving side might be enough to come up with a consistent definition / detection algorithm though.

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: I was fighting with rust syntax a while a go and had convinced myself that this wasn't valid for enums. I'm glad it is!

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: Are there ever times when you would find it appropriate to scope these kinds of `use` statements? I take your point that it makes sense at the file scope in this case though.

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: I'm for it!

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: `IntersectionComplexity` might become `IntersectionType { MapEdge, Terminus, Connection, Fork, Intersection(ConflictType) }` or something. I've added a comment talking about the awkwardness to `IntersectionComplexity`.

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: Yes, spot on. I'll clarify this in the "every pair of connections" comment.

BudgieInWA: In fact, the logic doesn't depend on the sort order of `small_con`/`large_con` (it may have at one point). We also aren't doing anything with the full set of conflicts, so I'll omit storing them all here.

osm2streets/src/render.rs
dabreegster: I'm open to trying either way. This is sounding like a good thing to unit test individually -- it's probably just a few weird edge cases I'm worrying about

osm2streets/src/lib.rs
dabreegster: I just finished writing up #117 and this sounds so much better. The dataflow "framework" doesn't have to make the entire codebase complex; this batching executor could isolate it. I really agree at least two of the current transformations are something much more fundamental / less optional

osm2streets/src/transform/classify_intersections.rs
dabreegster: If the enum derives or implements `PartialEq`, then you can just do regular `==`. `if let` is nice for extracting something from an enum case. Like if we later do... `if let IntersectionComplexity::Crossing(conflict_type) = inter.complexity`

osm2streets/src/transform/classify_intersections.rs
dabreegster: It might be a matter of preference... I think I only do it to avoid unused warnings when doing build flags. Like if you have a web-only `#[cfg(target_arch = "wasm32")]` block and an import that's only used (or available) in there

osm2streets/src/transform/classify_intersections.rs
BudgieInWA: > If the enum derives or implements `PartialEq` ...

This is the details that I didn't realise was relevant! Makes sense now.

PR #97, opened 2022-09-22
Rename the crates for a bit of reorganisation

BudgieInWA:
What do you think of these renames?


Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [workspace]
 members = [
-    "import_streets",
-    "osm2streets",
+    "streets_reader",
```
BudgieInWA, line 3: `streetsio`?

dabreegster: Maybe? Or `osm_reader` to emphasize the source, but maybe that just repeats the main crate name? Not opinionated here

dabreegster (approved):
I'm a fan of the changes! `tests-web` incorrectly emphasized tests. And making the main crate match the project name is great

Cargo.toml
BudgieInWA: I like `streetsio` as an analogy to [osmio](https://crates.io/crates/osmio), but I'll leave it as is until its context in the wider ecosystem becomes more obvious.

Issue #93, opened 2022-09-16
Store roads per intersection


dabreegster:
Store the roads per intersection in [the struct](https://github.com/a-b-street/osm2streets/blob/322044d26785d92f328931b610ae923ec1389161/street_network/src/lib.rs#L539) directly, and remove [roads_per_intersection](https://github.com/a-b-street/osm2streets/blob/322044d26785d92f328931b610ae923ec1389161/street_network/src/lib.rs#L162). Maintain this as we do various transformations.

Then we can make sure this list of roads in clockwise order and always maintain that.

Then that can help classify `IntersectionComplexity`

PR #98, opened 2022-09-26
Cache roads per intersection. #93

dabreegster:
Since inserting or removing a road in the graph now requires maintaining state in intersections, we have to be careful and use the new methods that keep everything correct. I hit a few hard-to-debug problems where the order of operations in some transformations matters... you have to delete the road first, then the intersection. Otherwise, `remove_road` will be upset that one of the intersections doesn't exist.

This is still broken -- two tests still failing for reasons I'm having trouble debugging. The complexity of the code and transformations increases a bit with this change, but I think it's a bit necessary. The previous brute-force `roads_per_intersection` search was really inefficient.

Once I fix the last bug, I plan to add clockwise ordering using the original OSM center line. When we merge short roads, this'll have some occasional brokenness maybe, but will see


osm2streets/src/lib.rs
```diff
@@ -544,6 +573,11 @@ pub struct Intersection {
     pub control: ControlType,
     pub elevation: Distance,
 
+    /// All roads connected to this intersection. They may be incoming or outgoing relative to this
+    /// intersection.
+    /// TODO: Guarantee they're sorted clockwise
+    pub roads: Vec<OriginalRoad>,
```
BudgieInWA, line 579: Is there a good type that has an `insert_at` to take advantage of the fact that roads is already sorted?

BudgieInWA:
Great to see! Sorted roads on all intersections will unlock movements in #96, so this progress is exciting.

BudgieInWA:
This is my approach to sorting roads, if it is decipherable:

osm2streets/src/lib.rs
dabreegster: Like https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert ? When we're adding a new road, we could calculate its position in the list and add it, or just recalculate the ordering for the intersection. It'll depend on the sorting algorithm

BudgieInWA: Right. Performance doesn't really matter, so `Vec` works if we want to do it the "find and insert" way.

dabreegster:
OK @BudgieInWA, this is ready for review! The changes:

1) `Intersection` now stores `Vec<OriginalRoad>`, sorted in clockwise order
2) There are new methods like `insert_road`, `remove_road`, `retain_roads` that maintain invariants. They must be used, and carefully. If you follow through the commits in order, you'll see me making some errors with the order things are called, or forgetting to call things, and the resulting bugs were not the easiest to track down.
3) I rewrote the logic that clips the imported OSM data to the boundary polygon entirely. It was working in a too-crazy way before, because when an intersection is out-of-bounds but connected to multiple roads, we clone the intersection and reassign IDs. The IDs changing was complicated to reason about now that we store derived state in more places. I think the newer logic is easier to follow.
4) I added a debug layer to show the clockwise sorting in the web UI. We can probably remove it after we're confident sorting works.

There are lots of test diffs. Internal IDs change a bit, especially the negative ones that're created by clipping. Some coordinates also change. I visually diffed all of the cases and they all look identical, except for northgate, where a long disconnected piece of light rail gets clipped a bit differently.

The clockwise sorting is probably imperfect, but all the cases I've looked at are fine:
Screenshot from 2022-10-01 13-14-34
Screenshot from 2022-10-01 13-14-07

I wanted to simplify the logic in `geometry/algorithm.rs` to use the existing clockwise sorting, but I hit a bunch of bugs I don't understand yet. I'll try again and follow up with a smaller change. I think this PR should be enough to unblock the movements work though!

osm2streets/src/lib.rs
```diff
@@ -286,6 +303,60 @@ impl StreetNetwork {
                 .push((self.roads[&r].untrimmed_road_geometry().0, label.into()));
         }
     }
+
+    // Restore the invariant that an intersection's roads are ordered clockwise
+    //
+    // TODO This doesn't handle trim_roads_for_merging
```
dabreegster, line 309: In cases where we remove a short road and merge intersections, sorting around a "center" point gets kind of funny. `geometry/algorithm.rs` takes care to try and handle that. I didn't port the same logic here yet.

osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -157,11 +158,11 @@ pub fn collapse_intersection(streets: &mut StreetNetwork, i: NodeID) {
...
 
     info!("Collapsing degenerate {}", i);
-    streets.intersections.remove(&i).unwrap();
     // We could be more careful merging percent_incline and osm_tags, but in practice, it doesn't
     // matter for the short segments we're merging.
-    let mut new_road = streets.roads.remove(&r1).unwrap();
-    let mut road2 = streets.roads.remove(&r2).unwrap();
+    let mut new_road = streets.remove_road(&r1);
+    let mut road2 = streets.remove_road(&r2);
+    streets.intersections.remove(&i).unwrap();
```
dabreegster, line 165: Order matters! `remove_road` cleans up the two intersections, so they need to both exist until the end

osm2streets/src/lib.rs
BudgieInWA: I think the best way to determine the order of roads in those multi-node intersections from scratch is to use the OSM graph structure. First sorting the nodes (that are the Road endpoints) then by bearing of the first OSM segment.  https://github.com/a-b-street/osm2streets/pull/98#issuecomment-1259035988

This would require keeping the OSM graph around in some capacity, at least within multi-node intersections.

BudgieInWA (approved):
The results look good. I like the idea of merging this, even if the order isn't correct all of the time. Definitely good enough to start experimenting with.

I've only had a cursory look over the code, and wont have much time for movement calculations this week, so I'm happy to see this go ahead!

osm2streets/src/lib.rs
dabreegster: I agree with keeping the graph structure around somehow. Preserving turn restrictions when we merge gets very complex, and a possibly simpler approach is to use the original graph and just ask questions about it. Keeping the graph also means we can map back to objects in OSM more directly.

I don't know how this would work, so shelving the idea for the moment, but you've definitely got me thinking...

dabreegster:
Cool, then I'll merge! Based on approaching timelines for some work stuff, I'm about to start an issue and some experimentation with pedestrian/bike crossings. Not sure yet how things will fit into the movement PR that's started, but we'll find out I guess

osm2streets/src/lib.rs
BudgieInWA: I like "Asking questions about it". In fact, load and leave the osm graph completely in place. Every road should get it's own copy of the geom (like when you merge similar ways currently) which can be promptly offset to account for placement. Then further tweaked if needed (like sausage links). (Think of an editor, where the graph is literally drawn on the map. One could point at a lane to edit the way.)

(Roads and intersections need to have IDs independant of the OSM ids they reference, [and they need to know their own ID, so we don't need to pass pairs eveywhere])

Asking questions about roads is how I'm thinking about movements too. It's like I want to iterate flows of traffic clockwise around the intersection. A way could yield a single (`oneway`) flow, or commonly two. An intersection provides an `iter_cw_flows() -> Flow`, where Flow is a view onto a subset of lanes on a Road (that all go in the same direction).

I think that's how the type system helps me out here: There exists a whole lot of data about a whole lot of lanes, laid out in memory in one way or another, I want to slice it up differently, with differently typed structs, for different tasks. It's like the [HashMap Entry type](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html).

Keeping an immutable copy of the OSM graph and making the Street graph independent will no doubt become GH Issues to discuss and target soon.

PR #103, opened 2022-10-15
Fix subtract underflow crash when calculating backwards lane counts

### LINKED ISSUE #102: Subtraction underflow crash calculating backwards lane count
Hi, I ran into a crash when trying to load Edinburgh, Scotland into A/B Street:
`thread 'main' panicked at 'attempt to subtract with overflow', /home/david/.cargo/git/checkouts/osm2streets-fbbbdf753e406749/a55aef7/osm2streets/src/lanes/classic.rs:110:20`

Did some debugging and saw [here](https://github.com/a-b-street/osm2streets/blob/main/osm2streets/src/lanes/classic.rs#L110), `n = 2` and `num_driving_fwd=3`, which doesn't make much sense, but was causing a crash. 

I added a workaround to check for situations where that subtraction would underflow and default to 0 backward lanes, diff here: https://github.com/a-b-street/osm2streets/compare/main...dmfutcher:osm2streets:fix-lanes-calculation-crash. This works and I successfully loaded Edinburgh, but I'm not sure if that's the "correct" way to fix it. I'd be happy to hack on this and submit a PR if you could give me some guidance on the correct way to handle this situation (I found this software looking for something to hack on for Hacktoberfest).

You can repro using this GeoJSON: https://gist.github.com/dmfutcher/2b906f1b2b953a62ea1da8b361ab2fed

dmfutcher:
Fixes crash when value of `lanes:forward` is greater than value of `lanes`. This _should_ not happen but has been spotted on live OSM data. This will at least let us handle the situation more gracefully.

Fixes #102

### DIFF
--- a/osm2streets/src/lanes/classic.rs
+++ b/osm2streets/src/lanes/classic.rs
@@ -107,7 +107,12 @@ pub fn get_lane_specs_ltr(tags: &Tags, cfg: &MapConfig) -> Vec<LaneSpec> {
     {
         n
     } else if let Some(n) = tags.get("lanes").and_then(|num| num.parse::<usize>().ok()) {
-        let base = n - num_driving_fwd;
+        let base = if n > num_driving_fwd {
+            n - num_driving_fwd
+        } else {
+            0
+        };
+
         if oneway {
             base
         } else {


dabreegster (approved):
Thanks for the fix!

Issue #80, opened 2022-08-21
Lookup driving side automatically


Robinlovelace:
Not sure if this is an OSM data issue but, this ain't right:

![image](https://user-images.githubusercontent.com/1825120/185782479-a79441bb-61d1-4197-ba9a-c72f32f7bf1f.png)

dabreegster:
You need to use Settings to change driving side to left. #73 is in progress to figure this out automatically

Robinlovelace:
Also: I need to update OSM to say sidewalk=no

dabreegster:
I've never gotten the Overpass query for country to work. What if we do something... more brute-force? Per https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic, only 54 countries use left-handed driving. "admin 0 countries" from http://geojson.xyz/ is only about 600KB. Once stripping out 140 right-handed countries, removing all the properties, and maybe further simplifying the geometry, I bet we can get this file pretty small and quick to load. Would it be so crazy to just do a point-in-polygon check?

dabreegster:
This isn't a crazy idea at all. https://github.com/westnordost/countryboundaries#speed is doing it already, reporting fantastic performance. Maybe the equivalent for us is to build a quadtree once and serialize it too, if that helps

dabreegster:
https://github.com/a-b-street/osm2lanes/issues/103 is related.

https://github.com/streetcomplete/countrymetadata/blob/master/data/isLeftHandTraffic.yml
https://crates.io/crates/reverse_geocoder

I'm thinking:
- a small standalone Rust crate to do the query
- a quick python script to filter that geojson file using the YAML list

dabreegster:
We might also be able to get rid of the double dependency in osm2lanes of `celes` and `locale-codes`

Robinlovelace:
That would be another co-benefit :tada: watching this one with interest

Robinlovelace:
In terms of the actual issue, realise I wasn't clear in the original post. The issue is with the sidewalks being on the inside. Reproducible example: https://a-b-street.github.io/osm2streets/#19.5/53.81977/-1.54114

Robinlovelace:
Another issue with that view, this important shared use footway is not imported by osm2streets it seems: https://www.openstreetmap.org/way/1087921367#map=18/53.81506/-1.53894

[...]

PR #23, opened 2022-06-09
Start 'Street Explorer' test browser for viewing tests

BudgieInWA:
Works towards #22, closes #13 by introducing a `trunk serve` slippy map with controls, that (will hopefully soon) shows the test data from `tests/`.

TODO
- [x] Allow trunk to read the tests dir to list test cases and load test case files
- [x] Add GeoJSON from a test file to the map (as proof of concept)
- ~~Generate the test list at build time, link to the `.json` files~~

CC @droogmic @dabreegster

PR #108, opened 2022-11-08
Calculate driving side automatically

dabreegster:
#80 
This is the followup to https://github.com/a-b-street/osm2lanes/pull/261. It deletes a bunch of test and UI code, because it's simpler to just calculate from a point!

### DIFF
--- a/osm2streets/src/types.rs
+++ b/osm2streets/src/types.rs
@@ -54,6 +54,9 @@ impl NamePerLanguage {
 pub struct MapConfig {
     /// If true, driving happens on the right side of the road (USA). If false, on the left
     /// (Australia).
+    ///
+    /// Note this is calculated by osm2streets! The value passed in is ignored; don't do any work
+    /// to set it.
     pub driving_side: DrivingSide,
     pub bikes_can_use_bus_lanes: bool,
     /// If true, roads without explicitly tagged sidewalks may be assigned sidewalks or shoulders.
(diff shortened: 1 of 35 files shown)


osm2streets/src/types.rs
```diff
@@ -54,6 +54,9 @@ impl NamePerLanguage {
 pub struct MapConfig {
     /// If true, driving happens on the right side of the road (USA). If false, on the left
     /// (Australia).
+    ///
+    /// Note this is calculated by osm2streets! The value passed in is ignored; don't do any work
```
dabreegster, line 58: It's a bit weird to let the caller be able to specify this at all. It'd be a bigger refactor to pull this field out elsewhere, and it kind of fits here otherwise. So for now, seems OK

rust-toolchain.toml
```diff
@@ -1,4 +0,0 @@
-[toolchain]
```
dabreegster, line 1: @BudgieInWA, you added this in #23. I'm getting rid of it because:
1) Pulling in `driving-side` created duplicate geo and geojson dependencies
2) I tried `cargo update -p geom` to grab a newer version of the A/B Street crate
3) But hit https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#workspace-inheritance, since I switched to workspace-level dependencies over there.
I realized my Rust beta install was still on 1.62. I've tested everything in osm2streets with 1.64, the latest stable, and it all works. I don't remember what beta features you were making use of a few months ago, but looks like they're in stable now!

dabreegster:
Again to avoid diff-basing over in the abstreet repo, I'm going to merge now, but happy to fix anything async!

rust-toolchain.toml
BudgieInWA: I think it was some part of the `trunk` that needed beta, so I'm happy to see a return to stable.

PR #109, opened 2022-11-08
Fix bus lane order in left-sided driving

### LINKED ISSUE #83: Bug: incorrect order of bus:lanes when the driving side is left
If the driving side is set to "left", the values in `bus:lanes` are unexpectedly flipped. 

Example: [Go here](https://a-b-street.github.io/osm2streets/#20.5/-36.85183/174.76605), change the driving mode to left. Note how the bus lane is on the very right, when it should be on the very left.

the value of `bus:lanes=designated|yes|yes` should not be flipped in left-hand mode

image

---

[Another example, without dual carriageway](https://a-b-street.github.io/osm2streets/#20.5/-36.79778/174.76255): The bus lane should be on the southern side, not in the middle

```
bus:lanes:backward=designated|yes
lanes=3
lanes:backward=2
lanes:forward=1
```

image

dabreegster:
Closes #83

The commits are easier to view individually. The fix itself is pretty simple. And then since I feel like we'll be living with the classic osm2lanes algorithm for a while, I went ahead and split it into smaller parts, inspired by the current osm2lanes structure. Tests show everything is still fine.

The two reported cases are fixed:
Screenshot from 2022-11-08 13-14-05
Screenshot from 2022-11-08 13-12-49

### DIFF
--- a/osm2streets/src/lanes/classic.rs
+++ b/osm2streets/src/lanes/classic.rs
@@ -152,7 +152,6 @@ pub fn get_lane_specs_ltr(tags: &Tags, cfg: &MapConfig) -> Vec<LaneSpec> {
     let mut back_side: Vec<LaneSpec> = iter::repeat_with(|| back(driving_lane))
         .take(num_driving_back)
         .collect();
-    // TODO Fix upstream. https://wiki.openstreetmap.org/wiki/Key:centre_turn_lane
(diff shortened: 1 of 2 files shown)


dabreegster:
@k-yle I can't tag you as a reviewer, but if you're interested at all in ever working on this codebase, this is an example PR

PR #110, opened 2022-11-09
Use SmoothWheelZoom Leaflet lib for smooth zooming

jakecoppinger:
I found the zoom steps of 0.5 frustrating so I implemented smooth zooming with a library. Is this a desired feature? 😄 

Library is under MIT license from
https://github.com/mutsuyuki/Leaflet.SmoothWheelZoom

### DIFF
--- a/street-explorer/www/index.html
+++ b/street-explorer/www/index.html
@@ -15,6 +15,7 @@
     <link rel="stylesheet" href="css/deps/geosearch.css" />
 
     <script src="js/deps/leaflet.js"></script>
+    <script src="js/deps/SmoothWheelZoom.js"></script>
     <script src="js/deps/d3.min.js"></script>
     <script src="js/deps/graphviz.min.js"></script>
     <script src="js/deps/d3-graphviz.js"></script>
--- a/street-explorer/www/js/deps/SmoothWheelZoom.js
+++ b/street-explorer/www/js/deps/SmoothWheelZoom.js
@@ -1,0 +1,109 @@
+
+L.Map.mergeOptions({
+    // @section Mousewheel options
+    // @option smoothWheelZoom: Boolean|String = true
+    // Whether the map can be zoomed by using the mouse wheel. If passed `'center'`,
+    // it will zoom to the center of the view regardless of where the mouse was.
+    smoothWheelZoom: true,
+
+    // @option smoothWheelZoom: number = 1
+    // setting zoom speed
+    smoothSensitivity:1
+
+});
+
+
+L.Map.SmoothWheelZoom = L.Handler.extend({
+
+    addHooks: function () {
+        L.DomEvent.on(this._map._container, 'wheel', this._onWheelScroll, this);
+    },
+
+    removeHooks: function () {
+        L.DomEvent.off(this._map._container, 'wheel', this._onWheelScroll, this);
+    },
+
+    _onWheelScroll: function (e) {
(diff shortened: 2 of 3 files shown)


dabreegster (approved):
This is a great UX improvement! I was tempted to switch to MapLibre GL a while back because Leaflet's choppy scroll was so annoying. Thanks for the PR!

jakecoppinger:
Great, thanks!

BudgieInWA:
If I am not mistaken, the desired result can be achieved by simply setting `zoomSnap` to 0 and a using smaller `zoomDelta`. (I remember setting `zoomSnap` to 0 in the past, but that must have been in one of the other apps.) The only difference being, that `zoomDelta` applies to all methods of incrementing the zoom (keyboard, the overlay buttons, double click, and the scroll wheel) so we can't change it independently for different inputs, but I don't know if we actually want to anyway. @jakecoppinger do these settings allow the feel you are after?

This PR removes the `zoomSnap` and `zoomDelta` settings, setting them back to their defaults of 1. This is a regression, because we are now snapping to whole number zoom levels after `fitBounds` and after pinch zoom! We should use `zoomDelta` 0 either way. (leaflet has the best support for touch screen panning/zooming that I've seen on desktop - better than Google Maps, too.)

Zooming with this change feels janky/wobbly to me. I don't think it's a performance thing (it's still hitting 60fps), but something to do with the animation easing, or the way it pans the map as a result of scrolling while my mouse pointer is moving. It feels like the animation zooms in on where the pointer started, then when the animation finishes that point jumps to the new pointer position, or something like that. Now that I play with the original settings I can see how the long animation for each zoom step, and how it seems to eat subsequent scroll inputs during animation is annoying, but it doesn't feel as janky, and it is improved by using a smaller `zoomDelta`.

@dabreegster what is the "choppyness" that you want to avoid?

dabreegster:
> what is the "choppyness" that you want to avoid?

Previously, a single gentle flick of my scroll wheel made a big jump. This PR feels like it lowers the step size to a more comfortable amount. I haven't played with just lowering the previous `0.5` settings yet.

I'm not terribly opinionated about what we do here, but felt like this PR improved things a bit.

michaelkirk:
It might be that different platforms have different experiences here. With scroll changes it's helpful to test things on both linux scroll wheel and a mac touchpad since they can behave quite differently.

Other configurations to try if you want a really thorough test:
- linux touchpad
- All the above on linux X11 vs linux wayland
- I think the common apple "magic mouse" works just like the touchpad, but a scroll wheel mouse + macos would be interesting
- And whatever the confounding factors on windows might be...

BudgieInWA:
Thanks for clarifying.

I just found `wheelPxPerZoomLevel` (which we had in there already) and `wheelDebounceTime`. With the removal of any zoom snapping at all, I am confident we can achieve exactly what we want with the built in options alone. I will open a new PR with my take on it.

BudgieInWA:
I've been playing around with those settings. I can see now that leaflet will only ever zoom in discrete steps, though those settings do let us zoom in small steps.

Here is an example of the jank that the plugin introduces:

https://user-images.githubusercontent.com/3784591/201303563-7ab133c3-1969-4476-8187-850ce9b49482.mp4

BudgieInWA:
Turns out it's a bug that was easy enough to fix! See #116 and https://github.com/mutsuyuki/Leaflet.SmoothWheelZoom/pull/8

PR #114, opened 2022-11-11
Store IDs in roads and intersections

dabreegster:
This unblocks simpler code in the movements PR by storing the current IDs in each structure. I'm doing this separate from changing IDs to opaque, because that's going to be a much more complex refactor. This smaller step also forces us to find the places where currently we modify the ID of an existing road.

### DIFF
--- a/osm2streets/src/ids.rs
+++ b/osm2streets/src/ids.rs
@@ -1,0 +1,78 @@
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+
+use crate::osm::{NodeID, WayID};
+
+/// Refers to a road segment between two nodes, using OSM IDs. Note OSM IDs are not stable over
+/// time.
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct OriginalRoad {
+    pub osm_way_id: WayID,
+    pub i1: NodeID,
+    pub i2: NodeID,
+}
+
+impl fmt::Display for OriginalRoad {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "OriginalRoad({} from {} to {}",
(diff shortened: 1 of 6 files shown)


osm2streets/src/ids.rs
```diff
@@ -0,0 +1,78 @@
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+
+use crate::osm::{NodeID, WayID};
+
```
dabreegster, line 6: Planning to define the opaque IDs here too. `lib.rs` is getting big

osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -17,7 +17,7 @@ pub fn collapse(streets: &mut StreetNetwork) {
         if roads.len() != 2 {
             continue;
         }
-        match should_collapse(roads[0], roads[1], streets) {
+        match should_collapse(&streets.roads[&roads[0]], &streets.roads[&roads[1]]) {
```
dabreegster, line 20: Example case that maybe gets a bit simpler

dabreegster:
Going to merge now and get into an A/B Street release, feedback anytime welcome. Still planning to try opaque IDs

osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -51,7 +48,7 @@ fn should_collapse(r1: OriginalRoad, r2: OriginalRoad, streets: &StreetNetwork)
     // a bizarre example. These are actually blackholed, some problem with service roads.
     if road1.oneway_for_driving().is_some()
         && road2.oneway_for_driving().is_some()
-        && r1.i2 == r2.i2
+        && road1.id.i2 == road2.id.i2
```
BudgieInWA, line 51: Maybe introducing `start_intersection()` and `end_intersection()` (or whetever the naming should be) would be a worthwhile stepping stone towards opaque ids.

PR #115, opened 2022-11-11
Remove InitialMap, folding the fields in StreetNetwork directly. #2

### LINKED ISSUE #2: Reconsider RawMap and InitialMap
`RawMap` has an awkward API today.

1) `roads_per_intersection` is a method that does a linear search for basic graph functionality. Annoying and slow, but not the most urgent problem.
2) `RawRoad` has `center_points: Vec<Pt2D>`, representing the raw geometry from OSM. There are `untrimmed_road_geometry` and `trimmed_road_geometry` methods that run through the intersection geometry algorithm to trim the road on both ends.

If this is going to be a nice graph+geometry representation, the 2nd especially will become annoying quickly. It's actually already a problem trying use the map_editor to debug particularly weird geometry. You would hope that what's shown there matches the fully imported map, but there are discrepancies sometimes -- https://github.com/a-b-street/osm2streets/issues/62. I haven't looked into this in quite a while, but I think it might have to do with ordering. We trim every road twice, for both intersections. When we do a full import, the order of trimming doesn't match what happens in the map_editor UI, especially when we're modifying things and recalculating only some intersections.

And then there's [InitialMap](https://github.com/a-b-street/abstreet/blob/osm2lanes/raw_map/src/initial.rs). It sort of solves both problems above.

So my task: keep the OSM center_points on `RawRoad`, but add `trimmed_center_pts` there and see if we can totally do away with `InitialMap`. It _should_ simplify code and make the map_editor match the imported map in those niche cases. It'll also increase `RawMap` file size a bit. If it's much, we can avoid serializing and fill it out from center_points when we load. (The only typical place saving a `RawMap` is `convert_osm`, which creates a sort of "unsimplified / untransformed" `RawMap`)

CC @BudgieInWA

dabreegster:
At last, this closes #2! I think the code is now simpler (a bit more internally stateful, but there's not a second struct with partly overlapping responsibility). And the web app gets faster, because each layer doesn't need to keep calculating all the geometry.

The test diffs look spurious to me. `leeds_cycleway` has geometry changes, but they're invisible to me. The `.dot` files look like some IDs have jumped around -- not entirely sure why.


osm2streets/src/lib.rs
```diff
@@ -200,40 +199,18 @@ impl StreetNetwork {
         self.intersections[&i].roads.clone()
     }
 
-    /// (Intersection polygon, polygons for roads, list of labeled polygons to debug)
-    #[allow(clippy::type_complexity)]
```
dabreegster, line 204: Used in the abst rawmap editor. I'll do something different there -- probably running the transformation first and using geometry directly. Might have to come back later and add a method to just focus that transform on a few intersections.

osm2streets/src/lib.rs
```diff
@@ -200,40 +199,18 @@ impl StreetNetwork {
...
+    /// This calculates a road's `trimmed_center_line` early, before
+    /// `Transformation::GenerateIntersectionGeometry` has run. Use sparingly.
+    pub(crate) fn estimate_trimmed_geometry(&self, road_id: OriginalRoad) -> Result<PolyLine> {
         // First trim at one of the endpoints
         let trimmed_center_pts = {
             let mut input_roads = Vec::new();
             for r in self.roads_per_intersection(road_id.i1) {
-                input_roads.push(initial::Road::new(self, r).to_input_road());
+                let road = &self.roads[&r];
+                // trimmed_center_line hasn't been initialized yet, so override this
```
dabreegster, line 210: I thought about setting `trimmed_center_line = untrimmed_center_line` in `Road::new`, but all the places that modify `untrimmed_center_line` would need to keep things in sync... This is the only place where we need to worry about this

osm2streets/src/lib.rs
```diff
@@ -602,6 +587,16 @@ impl Road {
         let right = center.shift_from_center(total_width, total_width / 2.0)?;
         Ok((left, right))
     }
+
+    // TODO Don't take our own ID
```
dabreegster, line 591: The other PR moving IDs into the struct would be helpful here and in a few other places!

osm2streets/src/pathfinding.rs
```diff
@@ -57,8 +56,7 @@ impl StreetNetwork {
             &graph,
             from,
             |i| i == to,
-            // TODO Expensive!
```
dabreegster, line 60: This stopped being expensive with the last PR that stores a polyline instead of vec<pt2d>

osm2streets/src/transform/intersection_geometry.rs
```diff
@@ -0,0 +1,124 @@
+use abstutil::Timer;
+use geom::{Circle, Distance};
+
+use crate::{ControlType, StreetNetwork};
+
+pub fn generate(streets: &mut StreetNetwork, timer: &mut Timer) {
```
dabreegster, line 6: git didn't detect this as a move from `initial.rs`, but the logic comes from there

dabreegster:
Rebased against the ID embedding, and going to merge now for an A/B Street release. Feedback later welcome

CONTRIBUTING.md
```diff
@@ -12,3 +12,7 @@ Come hang out at https://github.com/a-b-street/osm2streets to ask questions and
 See the `tests` crate [(docs)](tests/README.md) for a collection of interesting example scenarios.
 
 See `street-explorer` crate [(docs)](street-explorer/README.md) for a *Street Explorer* test browsing interface.
+
+## Design decisions
+
+The `Road` and `Intersection` structs are rather stateful. `trimmed_road_line` and `polygon` are derived state, produced by a transformation step. Derived state isn't updated automatically, so you have to be internally careful about knowing what transformations have run. Some earlier ones call `estimate_trimmed_geometry` for this reason.
```
BudgieInWA, line 18: If we wanted to make it safer to use `trimmed_road_line` and `polygon` without paying the cost of potentially unnecessary recalculations, we could go for a cached getter pattern: Make the fields private `Option`s that cache the result of the calculation, add public `trimmed_road_line()` and `polygon()` methods that calculate on demand, and set the fields to `None` any time we change data that invalidates them.

I don't yet know if that would be an improvement, as we still need to have a clear understanding of the dependencies between the different transformations, so maybe it's something we look into more as a performance improvement later.

osm2streets/src/lib.rs
BudgieInWA: Alternative to the cached getter idea: Keep the data stored within a single `Road` or `Intersection` internally consistent and safe to use, but allow the resulting network to have inconsistent (overlapping etc.) geometry until the appropriate transformations have run.

I imaging setting `trimmed_center_line = untrimmed_center_line` and `{start,end)_trim_distance = 0` in `Road::new` like you suggest, and `polygon = <some little square/circle at the center>` in `Intersection::new`. That way it is always safe to use those fields in code, and the ordering of transformation only matters for interpreting the values.

I imagine instantiating all of the `Roads` and `Intersections` from the osm data and being able to visualise it immediately (using the normal debug vis mechanisms). First a transformation accounts for placement. Then a transformation looks at all of the intersections and trims roads back just enough to generate sensible intersection geom. Then a transformation collapses roads with a negative trimmed distance (therefore merging neighbouring intersections that have overlapped). Then movements can be calculated and intersections classified. Then all manor of heuristic transformations can take place, like parallel way snapping, like collapsing short `Road`s between two `Cross`ing `Intersection`s (because `Cross` implies roads markings are interrupted, so if the road is short it probably doesn't have it's own road markings, so it's probably internal to the intersections), or like increasing the trim distance based on the kind of intersection or the position of pedestrian crossings/stop lines.

I haven't got a good picture of how the existing transformations fit together yet, I will wrap my head around it as I work on placement tags and on trying to tweak intersections based on `complexity`/`conflic_level`.

dabreegster: I like this idea. It'd also mean the system is totally self-explaining through the debug steps -- huzzah for slide decks for free! Explain how things work by stepping through.

> Then a transformation looks at all of the intersections and trims roads back just enough to generate sensible intersection geom.

I've been picturing the geometry calculation as always happening at the very end based on how the other transformations work currently: (almost) none of them use trimmed distances. But that's just a historic assumption.

> Then a transformation collapses roads with a negative trimmed distance (therefore merging neighbouring intersections that have overlapped).

Hmm, nothing produces negative trim distances today, but that's probably an assumption baked into the intersection geometry algo. This sounds like a nicer way to detect the case of two intersections overlapping than actually checking the polygons themselves.

> like collapsing short Roads between two Crossing Intersections (because Cross implies roads markings are interrupted, so if the road is short it probably doesn't have it's own road markings, so it's probably internal to the intersections)

!! This is the great fresh new ideas I'm super hyped for. If you dig around, you'll probably find 2 or 3 incorrect attempts to auto-detect short roads for merging. None of those attempts have had anything resembling intersection complexity as an input.

> I haven't got a good picture of how the existing transformations fit together yet

I will try to start some more substantial docs about how everything works currently. Please do ask questions in the meantime.

Lots of new ideas / actions in this comment, will keep starred in my own notes for now

dabreegster:
Ahh wait, it's because even the basic geometry layer uses trimmed center lines. OK,

>  imaging setting trimmed_center_line = untrimmed_center_line

I'll look into doing this now to unbreak the debug mode. Alternatively, we need to run this final transformation step before rendering each debug layer

dabreegster:
I restored the behavior with a quick hack. Thinking about how to do this properly still

PR #113, opened 2022-11-11
Remove the zoom plugin, but disable zoom snapping instead

BudgieInWA:
This PR is an alternative to #110 using built in options to achieve fine grained zooming.

Disabling zoom snapping is a must for fine grained scrolling and is an improvement to the ping-zooming UX via touch screen.

With zoom snapping disabled, we are able to actually achieve smaller zoom increments using `wheelPxPerZoomLevel` . The value of 90 feels like a nice balance for me, using the default mouse settings on Windows 10 (which scrolls about 6 lines), a mouse on KDE Plasma (which scrolls 3 lines), and a trackpad on KDE Plasma.

PR #116, opened 2022-11-12
Fix the jank in the SmoothWheelZoom plugin

BudgieInWA:
This PR fixes the jankiness in the plugin that was introduced in #110, and re-introduces the zoomSnap and zoomDelta settings which are still needed for other zooming methods. It is an alternative to #113.

`zoomSnap` is set to 0 to disable snapping altogether, which is a great win for touch-screen pinch zooming.

Zooming with a scroll wheel and track pad scrolling both feel great to me!


dabreegster (approved):
The result also feels great to me! Nice job for fixing the upstream problem; this is the great thing about open source

BudgieInWA:
> Nice job for fixing the upstream problem; this is the great thing about open source

It feels great when a hunch fixes the bug first go, but even better when it fixes someone unrelated's "hell naw" reaction from a [2 year old Stack Overflow question](https://stackoverflow.com/questions/41759536/truly-smooth-zooming-in-leaflet) :D

image

Issue #95, opened 2022-09-20
Add "end cap angle" capability to "thick linestring" and Roads


BudgieInWA:
From a discussion on [intersection turns/movements](https://github.com/a-b-street/osm2streets/issues/67#issuecomment-1248685070), we talked about the idea of allowing roads to end with an angle other than 90 degrees.

The ability to generate thick linestring and "parallel" geometry with an arbitrary angle at the end will allow us to:

- have roads join intersections at differing angles,
- seamlessly join roads together at `Connection` type intersections.

BudgieInWA:
The desired road/lane end angle (where the stop line is often found) is just like the line of reflection that would be found running through a miter, if the road continued along the grey markings. Perhaps these "angled end caps" should be implemented as a capability of our "thick line strings". Seen [`angStart` and `angEnd` arguments in `getParallel`](https://github.com/BjornRasmussen/Lanes/blob/eed921526aceaad28dd9ea67f85358f446d6f62d/src/org/openstreetmap/josm/plugins/lanes/UtilsSpatial.java#L283)

![image](https://user-images.githubusercontent.com/3784591/190846284-6d422f20-3a80-403c-a550-8ce687d28030.png)

dabreegster:
The point projection does 90 degrees in the start/end, look at generalizing

PR #117, opened 2022-11-15
Brain-dump how osm2streets currently works

dabreegster:
This is quite messy, but more or less covers everything

### DIFF
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -1,0 +1,132 @@
+# How it works
+
+As of November 2022, and probably incomplete. This describes how the codebase currently works, not necessarily how we want it to.
+
+## The model
+
+At its heart, a graph of roads and intersections. (Roads lead between exactly two intersections -- "road segments" might be more precise.) Roads have their lanes listed from left-to-right, with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
+
+Intersections have a `ControlType` -- stop signs, traffic signals, uncontrolled, etc. This is orthogonal to `IntersectionComplexity` and `ConflictType`... TODO, narrow down valid combinations and give examples. MultiConnection vs Merge, please!
+
+### IDs
+
+Roads and intersections are currently identified by OSM IDs, but this will soon change to opaque IDs. As we merge, split, or fix things, the model diverges from what's in OSM. Instead, roads and intersections will keep a list of OSM IDs that went into the osm2streets object.
+
+## Import walkthrough
+
+### streets_reader
+
+`osm_to_street_network` is the main function, taking raw input OSM XML, an optional boundary clipping polygon, and some options, and returning a `StreetNetwork`. Some callers (A/B Street) repeat the logic of this method and add in extra bits (for adding in other sources of parking and elevation data).
+
+1.   Extract raw info from OSM
+2.   Split ways
+3.   Clip to the boundary
+4.   Match extra stuff
+
+Extraction is straightforward. Since OSM ways often cross many intersections, they don't form a graph yet, so the split step finds nodes common to multiple ways and declares those intersections. Very small roundabouts also get collapsed to a single point here (a hack!). Raw turn restriction data and traffic signal nodes are also matched to a road. After this step, we have the first cut of a `StreetNetwork`. There are no movements filled out and geometry is almost exactly what OSM has.
+
+But from this point, roads do have their lanes filled out, parsed from OSM tags. That currently uses `osm2streets/src/lanes/classic.rs`, but will use a separate project `osm2lanes` in the future.
+
+Clipping takes the boundary polygon (which should be passed in explicitly, but can also just be the bounding box around the input XML) and removes roads totally out of bounds. Roads crossing the boundary will get clipped to the boundary, and that intersection will be marked as a border.
+
+The final step here takes raw crossing and barrier nodes and matches them to roads. This representation of those features is very early stages and will definitely evolve more.
+
+### Transformations
+
+The `StreetNetwork` is techncially usable at this point, but it's still very close to OSM -- which is both under-specified (lane width is almost never tagged, but we need to render something) and imprecise. The rest of the magic happens by calling `apply_transformations`. This performs the specified steps in order. `apply_transformations_stepwise_debugging` can be used by UIs to preserve the intermediate `StreetNetwork` after each step, for debugging and understanding the transformations.
+
+The caller explicitly lists the transformations they want, in order. `standard_for_clipped_areas` is a good list to start with. Splitting things into explicit steps like this is good:
+
+- Users can opt into experimental steps
+- Some callers may not want to deviate too far from OSM, because they're using osm2streets to edit directly
+- Incremental debugging
+
+But it's also confusing in a few ways:
+
+- Roads and intersections both contain derived state. When we modify something, we may need to re-run some transformations. For example, after collapsing sausage links, a road's trimmed road geometry changes, so we may need to detect and collapse short roads again. Effects from one transformation may need to propagate to adjacent roads and intersections. The dataflow is implicit; the caller must deal with it manually.
+- Some transformations fill out state that's pretty fundamental, like intersection complexity and movements. This maybe shouldn't be expressed as a transformation and should happen more upfront, like how `Road::new` immediately determines lanes from OSM tags.
+
+## Operations on a StreetNetwork
+
+Some of the below transformations do something complicated worth calling out. Likely this code shouldn't be in `transform/`.
+
+### Collapsing an intersection
+
+`collapse_intersection` removes an intersection that has only two roads connected to it. Transformations below describe when this is called. The operation itself mostly just stitches together the geometry of the two roads in a straightforward way. It fixes turn restrictions (which should get simpler with opaque IDs). The caller is responsible for fixing up lanes between the two roads -- or rather, making sure they match up compatibly in the first place.
+
+### Collapsing a short road
+
+`collapse_short_road` removes a road, then combines the two intersections into one.
+
+It first calculates trimmed intersection geometry for the two intersections. On each connected road (besides the short one being collapsed, of course), we store the trimming distance in `trim_roads_for_merging`, so that later the intersection geometry algorithm can follow a special case for the single merged intersection.
+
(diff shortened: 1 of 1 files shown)


dabreegster:
And after writing this (basically a quick code review of everything), I'm extremely convinced switching to opaque IDs from OSM ones will dramatically simplify a bunch of code. Whenever we merge / delete / modify something, we won't need to do this crazy dance of fixing up IDs for turn restrictions and connecting objects

docs/how_it_works.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Collapsing an intersection
+
+`collapse_intersection` removes an intersection that has only two roads connected to it. Transformations below describe when this is called. The operation itself mostly just stitches together the geometry of the two roads in a straightforward way. It fixes turn restrictions (which should get simpler with opaque IDs). The caller is responsible for fixing up lanes between the two roads -- or rather, making sure they match up compatibly in the first place.
+
+### Merging a short road
+
+`merge_short_road` removes a road, then combines the two intersections into one.
+
+It first calculates trimmed intersection geometry for the two intersections. On each connected road (besides the short one being merged, of course), we store the trimming distance in `trim_roads_for_merging`, so that later the intersection geometry algorithm can follow a special case for the single merged intersection.
```
BudgieInWA, line 61: I mentioned briefly in our recent meeting that I think the trim distances and multi-node intersections are fundamental to the data model we want. I hope these ideas can be incorporated into the data types over time.

docs/how_it_works.md
```diff
@@ -0,0 +1,132 @@
...
+
+Most of the complexity here comes from fixing up IDs and turn restrictions. Opaque IDs likely make this a much simpler operation!
+
+## The transformations
+
+TODO: before/after pictures
```
BudgieInWA, line 67: I don't think it's worth getting so carried away as to add screenshots here. I think this is valuable as a jumping off point that should get shorter and shorter as the structure of the code gets clearer, and as the docstrings improve.

BudgieInWA (approved):
Great! The underlying reasoning and intentions listed here are very helpful.

docs/how_it_works.md
dabreegster: Fair enough. At some point for explaining purposes, it'd be nice to have a "gallery" of all the transformations available. Maybe if we plumb more info into the street-explorer URLs, we could eventually express "run with debug steps, and show step 5"

docs/how_it_works.md
dabreegster: `trim_roads_for_merging` is capturing start/end trim distances before collapsing a short road. I agree storing this on every road could make things much more clear. Still not sure how multi-node intersections should work.

If I'm thinking about the hydrated concept, maybe it becomes less weird. As a first pass, we run the intersection geometry algorithm on the original network. Later we remove a short road, but leave the trim distances alone:
consolidate_pt3
Still feels like it's a bit special-cased; that algorithm needs to know to use the trim distances that've been calculated before...

docs/how_it_works.md
```diff
@@ -0,0 +1,132 @@
...
+
+Some of the below transformations do something complicated worth calling out. Likely this code shouldn't be in `transform/`.
+
+### Collapsing an intersection
+
+`collapse_intersection` removes an intersection that has only two roads connected to it. Transformations below describe when this is called. The operation itself mostly just stitches together the geometry of the two roads in a straightforward way. It fixes turn restrictions (which should get simpler with opaque IDs). The caller is responsible for fixing up lanes between the two roads -- or rather, making sure they match up compatibly in the first place.
+
+### Merging a short road
+
+`merge_short_road` removes a road, then combines the two intersections into one.
```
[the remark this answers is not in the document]
dabreegster: aefade1d5cd5fdca88e20ee8d7b2af296d750d87

docs/how_it_works.md
BudgieInWA: > ...that algorithm needs to know to use the trim distances that've been calculated before...

This is the kind of reason I think trim distance needs to be a fundamental part of the `Road` model. Editing a bunch of OSM highways teaches you that there are many scenarios where the geometry near the end of a way is nonsense - usually the physical feature that the way represents has ended - but it is clear why the way continues: because it needs to connect the graph properly. Slip lanes and sausage links are the weirder ones, where you commonly end up making aesthetic decisions about where to put the connecting geometry, but every traffic intersection has the same thing.

I think a fundamental trim distance field should exist on `Road` that represents where the physical feature actually ends, or at least where the `Road` ends and the `Intersection` takes over. I'm not sure quite how it will all shake out, but I feel like a sort of "minimum" trim distance can be calculated very early based on where the roads overlap each other - much like it seems `trim_roads_for_merging` is doing. Maybe stop lines and other context clues can be incorporated right from the start, leading to more accurate `Intersection` area calculations, or maybe the come much later, staying out of the way of the early transformations and being considered near the end for largely aesthetic reasons.

---

The big complication is when a road continues uninterrupted through an intersection (like when a side road with a stop sign joins a major road, or the slip lane case). In these cases, some of the roads *do* physically continue right to the end of their geometry.

I think the best way to think about this for now, is that the `Road`s all get trimmed back to the overlapped area, and the `Intersection` takes responsibility for the whole overlapping area. The `Intersection` can figure out which roads continue through and use its own internal `Road` instances or something to draw the section of the road that continues through the `Intersection` area. The `Intersection` is already responsible for handling the individual lane turn paths, so this is pretty natural.

I think we can consider allowing the uninterrupted ("major") roads to remain untrimmed (with the "minor" roads getting trimmed to avoid any overlap) later in development when we have a better understanding. I think that would unlock some more sophisticated and nuanced outcomes, but we need to get a handle on the simpler cases first.

---

Editing complex intersections in OSM also teaches you that inside an intersection area (where the lane marking have ceased), the position of any ways doesn't really matter, only the connectivity of the ways. So the data is authored with the assumption that the whole intersection is being considered together. Therefore, our `Intersection` model should treat multi-node intersections as first class. Perhaps as the *other case* but not as *special case*.

dabreegster: All of this makes sense to me! An attempt to extract some next steps...

> I think a fundamental trim distance field should exist on Road that represents where the physical feature actually ends

Then as you suggested earlier, a `trim_distance_start` and `trim_distance_end` field would make sense. The more I look at 
```
// true if src_i matches this intersection (or the deleted/consolidated one, whatever)
    pub trim_roads_for_merging: BTreeMap<(osm::WayID, bool), Pt2D>,
```
the more I'm wondering why I chose this convoluted representation originally. It's equivalent to start and end trim distance per road, except it has an indirect way of referring to a road. I think once we have opaque IDs and the process of removing short roads from the graph doesn't have to reassign IDs everywhere, this BTreeMap will be even sillier.

> a sort of "minimum" trim distance can be calculated very early based on where the roads overlap each other

Right, so that'd pretty much mean running `transform/intersection_geometry.rs` upfront, and maybe incrementally recalculating trim distances and intersection polygons as we do later transformations. This makes the hydration idea even more relevant.

> Maybe stop lines and other context clues can be incorporated

I've never even looked at these in OSM in practice, it'd be great to at least start extracting them and displaying as an optional extra layer! We could see how well the current intersection trimming algorithm matches up to the stop lines, as a very basic start

> I think we can consider allowing the uninterrupted ("major") roads to remain untrimmed (with the "minor" roads getting trimmed to avoid any overlap)

One possible consequence of this could be how we define movements then. If the main road is uninterrupted, the movement geometry there may not exist (or be just a point). This would break A/B Street's current definition of turn conflicts for figuring out that a side road turning onto a main road. But the new clockwise-overlapping definition would sort that out.

Also, looks like the case of a side road confused me long ago: https://a-b-street.github.io/docs/tech/map/geometry/index.html#a-radically-simpler-approach. I found this representation problematic, because I wanted the intersection polygon to physically exist and be that entire rectangle. The more nuanced idea of an "intersection" without geometry of its own never occurred!

> Therefore, our Intersection model should treat multi-node intersections as first class. Perhaps as the other case but not as special case.

I'm _nearly_ at the point where `Intersection` will have `original_ids: Vec<osm::NodeID>`, so we can start to directly represent this. We could also store all of the internal `OriginalRoads` if that winds up proving useful.

BudgieInWA: > ... the more I'm wondering why I chose this convoluted [trim distance] representation originally.

Seems appropriate if it were in fact an edge case relating just to this transformation. I had the benefit of seeing the trim distance idea in the JOSM Lanes plugin, so I had it as part of my mental model right from the start.

> Right, so that'd pretty much mean running `transform/intersection_geometry.rs` upfront, and maybe incrementally recalculating trim distances and intersection polygons as we do later transformations. This makes the hydration idea even more relevant.

"Hydration" is a term I came up with while writing that comment to describe how I'm conceptualising some of the properties as derived data which is a more useful representation of the state. (The term is not entirely perfect, but "the hydration idea" is very useful).

It would by awesome if the intersection geometry/trim distance calculation can produce good results at every stage so that we can treat the geom and trim distances as purely derived state and recalculate them in the straight forward way as you describe. There is a chance that the trim distance needs to be actual state that we guess initially but otherwise need to maintain through transformations, but I'd like to avoid that if possible. 

> I've never even looked at [stop lines] in OSM in practice, it'd be great to at least start extracting them and displaying as an optional extra layer! We could see how well the current intersection trimming algorithm matches up to the stop lines, as a very basic start.

Yes, that would be cool. In my experience, they are not commonly present in the data, but I think that's because they aren't used anywhere that I've seen. Making them move visible would be a good outcome in its own right. 

> > I think we can consider allowing the uninterrupted ("major") roads to remain untrimmed (with the "minor" roads getting trimmed to avoid any overlap)
> 
> One possible consequence of this could be how we define movements then. If the main road is uninterrupted, the movement geometry there may not exist (or be just a point). This would break A/B Street's current definition of turn conflicts for figuring out that a side road turning onto a main road. But the new clockwise-overlapping definition would sort that out.

These are exactly the sorts of complications that I was coming up with, and why I think we shouldn't pursue it yet.

> Also, looks like the case of a side road confused me long ago: https://a-b-street.github.io/docs/tech/map/geometry/index.html#a-radically-simpler-approach. I found this representation problematic, because I wanted the intersection polygon to physically exist and be that entire rectangle. The more nuanced idea of an "intersection" without geometry of its own never occurred!

The piece that might be worthwhile earlier on is supporting `Connection` intersections whose geometry is a line with no area (like where the vertical roads meet in the linked diagram, if the side road didn't exist). That would require "angled end caps" #95 and probably a non-intersection solution to drawing the beginning/end of new lanes (like where turn lanes appear from nothing). But for now I think those situations can all be treated as the same kind of thing: movements through large `Intersection` areas, some of which are marked as normal lanes.

PR #118, opened 2022-11-18
A few refactors

dabreegster:
Viewing commits individually may be easier.

1) `roads_per_intersection` returns `Vec<&Road>`. It's more convenient for some callers. The others can just lookup the `Vec<OriginalRoad>` themselves.
2) Move `Road` and its impl into a module, just because there's enough methods there to feel like a bit more organization is nice.
3) Remove some of the methods used by the `RawMap` editor. They're for interactively dragging intersections around. I think it's noisy / irrelevant to have these methods in this repo; they don't yet batch mutations or recalculate derived things in any nice way. If we want to support this kind of operation on a streetnetwork "properly", we can bring the code back and think through it again.

The big refactor coming after this is switching to opaque IDs

### DIFF
--- a/osm2streets/src/lib.rs
+++ b/osm2streets/src/lib.rs
@@ -10,10 +10,11 @@ use anyhow::Result;
 use serde::{Deserialize, Serialize};
 
 use abstutil::{deserialize_btreemap, serialize_btreemap};
-use geom::{Distance, GPSBounds, PolyLine, Polygon, Pt2D};
+use geom::{GPSBounds, PolyLine, Polygon, Pt2D};
 
 pub use self::geometry::{intersection_polygon, InputRoad};
 pub use self::ids::OriginalRoad;
+pub use self::intersection::Intersection;
 pub use self::lanes::{
     get_lane_specs_ltr, BufferType, Direction, LaneSpec, LaneType, NORMAL_LANE_THICKNESS,
     SIDEWALK_THICKNESS,
@@ -28,6 +29,7 @@ pub use self::types::{
 mod edit;
 mod geometry;
 mod ids;
+mod intersection;
 mod lanes;
 pub mod osm;
 mod pathfinding;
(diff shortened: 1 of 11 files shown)


osm2streets/src/lib.rs
```diff
@@ -117,10 +119,20 @@ impl StreetNetwork {
...
-        self.intersections[&i].roads.clone()
+    pub fn remove_intersection(&mut self, id: osm::NodeID) {
+        let i = self.intersections.remove(&id).unwrap();
+        if !i.roads.is_empty() {
+            panic!("Can't remove_intersection({id}), it has roads still connected");
+        }
+    }
+
+    /// Returns roads oriented in clockwise order around the intersection
+    pub fn roads_per_intersection(&self, i: osm::NodeID) -> Vec<&Road> {
```
BudgieInWA, line 130: This is an improvement in my eyes :+1:

It makes me wonder (more academically than practically) is it common for rust APIs to return `Iterator<&Road>` in this situation? (I just read through the rust book chapter on iterators, can you tell?)

BudgieInWA (approved):
Nice!

Can we move Intersection into its own file too? The symmetry will help with grokking they code layout i think.

dabreegster:
> Can we move Intersection into its own file too? The symmetry will help with grokking they code layout i think.

Done! I'd vote for moving `IntersectionComplexity`, `ControlType`, etc into `intersection.rs` too, but I won't do that yet. `types.rs` kind of started as a miscellaneous / catch-all/

osm2streets/src/lib.rs
dabreegster: Good question... I haven't read the book in ages and am not sure about best practices actually! If pretty much all callers are just iterating over the list, that'd probably be a better return type, to make things a bit lazier and save an odd manual call to `.into_iter()`.

I've returned something like `&[Road]` before in some cases, but that's only when a `Vec<Road>` is internally stored.

BudgieInWA: Thanks, I'll keep an eye out for APIs returning `Vec` vs `Iterator` to see if a pattern emerges. I'll keep it in the back of my mind as a micro-optimisation I think, avoiding allocating the `Vec` when it's just going to be iterated.

PR #120, opened 2022-11-19
Add src_i and dst_i fields to Road, as a step towards opaque IDs. Change

dabreegster:
some callers to prefer these to using OriginalRoad.

I started a few attempts to cutover to opaque IDs and store OSM IDs separately, but it's a brutally hard refactor. This is an intermediate step that starts changing some code to figure out road direction by looking at the `Road` struct instead of its ID. Not all callers converted over yet, but it's a decent chunk. Tests pass.

About naming: what I might do just for the refactor is make everything use `src_i` and `dst_i` fields on `Road`, so it's easy to grep code for `i1` and `i2` and see places left to convert. Once the cutover is done, I'm not too opinionated about how we call these, as long as it's consistent.

### DIFF
--- a/osm2streets/src/transform/remove_disconnected.rs
+++ b/osm2streets/src/transform/remove_disconnected.rs
@@ -49,7 +49,7 @@ pub fn remove_disconnected_roads(streets: &mut StreetNetwork) {
     for p in partitions.iter().skip(1) {
         for id in p {
             info!("Removing {} because it's disconnected from most roads", id);
-            streets.remove_road(id);
+            streets.remove_road(*id);
             next_roads.remove(id.i1, *id);
             next_roads.remove(id.i2, *id);
         }
@@ -57,7 +57,7 @@ pub fn remove_disconnected_roads(streets: &mut StreetNetwork) {
 
     // Also remove cul-de-sacs here. TODO Support them properly, but for now, they mess up parking
     // hint matching (loop PolyLine) and pathfinding later.
-    streets.retain_roads(|id, _| id.i1 != id.i2);
+    streets.retain_roads(|r| r.src_i != r.dst_i);
 
     // Remove intersections without any roads
     streets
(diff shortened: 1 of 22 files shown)


dabreegster:
Alright, I finished all the preps. Very minimal use of `OriginalRoad` now, so the next step of swapping it out for an opaque ID should be way easier

osm2streets/src/transform/remove_disconnected.rs
```diff
@@ -1,20 +1,11 @@
...
-use crate::{osm, OriginalRoad, StreetNetwork};
+use crate::{OriginalRoad, StreetNetwork};
 
 /// Some roads might be totally disconnected from the largest clump because of how the map's
 /// bounding polygon was drawn, or bad map data, or which roads are filtered from OSM. Remove them.
 pub fn remove_disconnected_roads(streets: &mut StreetNetwork) {
     // This is a simple floodfill, not Tarjan's. Assumes all roads bidirectional.
-    // All the usizes are indices into the original list of roads
-
-    let mut next_roads: MultiMap<osm::NodeID, OriginalRoad> = MultiMap::new();
```
dabreegster, line 13: I just rewrote this, now that intersections track their own roads

osm2streets/src/transform/separate_cycletracks.rs
```diff
@@ -101,11 +101,7 @@ fn snap(streets: &mut StreetNetwork, input: Cycleway) {
     assert!(streets.roads.contains_key(&input.cycleway));
 
     // Remove the cycleway, but remember the lanes it contained
-    let mut cycleway_lanes = streets
-        .roads
-        .remove(&input.cycleway)
-        .unwrap()
-        .lane_specs_ltr;
+    let mut cycleway_lanes = streets.remove_road(input.cycleway).lane_specs_ltr;
```
dabreegster, line 104: This was a huge bug, because the two intersections' list of roads got out of sync!

BudgieInWA:
Good approach.

And the naming idea makes sense too. I think forward/backward and start/end will be the best terms in the long run. Consistent with OSM and leaves source/dest (or should that be source/sink, origin/destination?) available as distinct terms for graph algorithms and connectivity contexts.

BudgieInWA (approved):

PR #122, opened 2022-11-21
Switch to opaque road and intersection IDs!

dabreegster:
This has several benefits:

1) We can manipulate the endpoints of a road without needing to change its ID everywhere
2) We can correctly track multiple OSM IDs making up one object
3) Lots of code gets simpler

I checked all tests manually. There are some visible changes in bristol_sausage_links, cycleway_rejoin_road, kingsway_junction, leeds_cycleway, and others. They don't look like major problems to me. I think possibly the amount of trimback used for terminus intersections is somehow changed; will look into. northgate_dual_carriageway and perth_stretched_lights have an intersection change for the better, seemingly due to the calculated clockwise ordering. Again, not totally sure why this PR affects them.

# Followups

`OriginalRoad` --a way ID and two OSM node IDs -- is a really bizarre name. Maybe we should rename it `osm::RoadSegment`? Or do we need to track it at all -- some `Road` just remember `Vec<osm::WayID>`?

There may be a few error cases in `transform/collapse_short_road.rs` that might've gone away now that we're not swapping out IDs in scary ways.


osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -130,24 +129,19 @@ fn should_collapse(road1: &Road, road2: &Road) -> Result<()> {
     Ok(())
 }
 
-pub fn collapse_intersection(streets: &mut StreetNetwork, i: NodeID) {
+pub fn collapse_intersection(streets: &mut StreetNetwork, i: IntersectionID) {
```
dabreegster, line 132: Much of the PR was a fairly mechanical search-and-replace. One of the big changes is here. I thought through this carefully and kind of rewrote the logic, to do less fiddling with IDs. Let me know if it's understandable, or if more comments would be helpful

osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -156,66 +150,62 @@ pub fn collapse_intersection(streets: &mut StreetNetwork, i: NodeID) {
...
-        i2: new_i2,
-    };
-    new_road.id = new_r1;
-    new_road.src_i = new_i1;
-    new_road.dst_i = new_i2;
-    streets.insert_road(new_road);
+    // Keep the same ID, but fix the endpoints
+    keep_road.src_i = new_src_i;
+    keep_road.dst_i = new_dst_i;
+    streets.insert_road(keep_road);
```
dabreegster, line 193: Note we remove and re-insert. This updates the derived state in intersections

osm2streets/src/transform/collapse_short_road.rs
```diff
@@ -15,36 +12,32 @@ use crate::{
 
 impl StreetNetwork {
     /// Collapses a road, merging the two intersections together. Returns (the surviving
-    /// intersection, the deleted intersection, deleted roads, new roads)
+    /// intersection, the deleted intersection). The caller can see all roads connected to the
+    /// surviving intersection.
```
dabreegster, line 16: Another massive change here, please check my logic carefully. These two "transforms" are some of the more complicated logic in the library so far

streets_reader/src/clip.rs
```diff
@@ -33,59 +34,48 @@ pub fn clip_map(streets: &mut StreetNetwork, timer: &mut Timer) -> Result<()> {
...
+        let mut old_intersection = streets.intersections.remove(&old_id).unwrap();
+        old_intersection.complexity = IntersectionComplexity::MapEdge;
+        old_intersection.control = ControlType::Border;
+
+        if old_intersection.roads.len() <= 1 {
+            // We don't need to make copies of the intersection; put it back
+            streets.intersections.insert(old_id, old_intersection);
+            continue;
+        }
+        for r in old_intersection.roads.clone() {
```
dabreegster, line 55: I think this bit gets much more clear. No more making up weird fake negative OSM IDs. We can remember the original OSM node ID just fine

streets_reader/src/split_ways.rs
```diff
@@ -24,19 +24,30 @@ pub fn split_up_roads(
 ) -> Output {
     timer.start("splitting up roads");
 
-    let mut roundabout_centers: HashMap<osm::NodeID, Pt2D> = HashMap::new();
-    let mut pt_to_intersection: HashMap<HashablePt2D, osm::NodeID> = HashMap::new();
+    let mut roundabout_centers: Vec<(Pt2D, Vec<osm::NodeID>)> = Vec::new();
+    // Note we iterate over this later and assign IDs based on the order, so HashMap would be
```
dabreegster, line 28: I also rewrote a little logic here. I had a bug at first here using HashMap. Note with BTreeMap, the internal order of our `IntersectionIDs` will follow spatial patterns based on how `Pt2D` is ordered by X and then Y. That shouldn't matter, just kind of worth mentioning

osm2streets/src/transform/collapse_short_road.rs
BudgieInWA: The logic modifying the intersections and roads seems sound, but I haven't attempted to grok the trim distance or turn restriction logic.

osm2streets/src/transform/collapse_intersections.rs
BudgieInWA: It's straighforward enough enough.

BudgieInWA (approved):
I've had a relatively thorough read through of the logic you've changed here and it seems to make sense to me. It's such a relief to see `OriginalRoad` and `osm::NodeID` disappear from places where those concepts have no bearing.

I've understood more about `split_ways.rs` and `clip.rs` which is good. (You keep nerd-sniping with relevant notifications every time I sit down to edit this codebase :stuck_out_tongue:). I've got some thoughts about boundary clipping and about sorting intersection roads that I'll write elsewhere, as they're not relevant to this change.

streets_reader/src/split_ways.rs
```diff
@@ -24,19 +24,30 @@ pub fn split_up_roads(
...
-            let id = input.osm_node_ids[&pts[0].to_hashable()];
-            roundabout_centers.insert(id, Pt2D::center(pts));
+            // TODO Test more carefully after opaque IDs
+            let id = ids[0];
+
             for pt in pts {
                 pt_to_intersection.insert(pt.to_hashable(), id);
             }
 
+            roundabout_centers.push((Pt2D::center(pts), ids));
```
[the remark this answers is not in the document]
dabreegster: I'm honestly not sure. I added a test case to track this, and it looks like it turns out OK:
Screenshot from 2022-11-22 11-43-52

streets_reader/src/split_ways.rs
```diff
@@ -24,19 +24,30 @@ pub fn split_up_roads(
 ) -> Output {
     timer.start("splitting up roads");
 
-    let mut roundabout_centers: HashMap<osm::NodeID, Pt2D> = HashMap::new();
-    let mut pt_to_intersection: HashMap<HashablePt2D, osm::NodeID> = HashMap::new();
+    let mut roundabout_centers: Vec<(Pt2D, Vec<osm::NodeID>)> = Vec::new();
```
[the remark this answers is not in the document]
dabreegster: I agree. I don't remember if there's a good reason it's happening so early here; probably just legacy code at this point. I'll followup later and move it into transformations.

osm2streets/src/transform/collapse_short_road.rs
```diff
@@ -58,18 +51,20 @@ impl StreetNetwork {
         // [X] road we're deleting is the 'via' of a complicated restriction
         // [ ] road we're deleting has turn lanes that wind up orphaning something
 
-        if src_i == dst_i {
-            bail!("Can't collapse {short} -- it's a loop on {src_i}");
+        if keep_i == destroy_i {
+            bail!("Can't collapse {short} -- it's a loop on {keep_i}");
```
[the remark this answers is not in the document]
dabreegster: I think this case is impossible to reach right now... I'll add a TODO to revisit this, but I think we need a test case to make sure nothing weird happens.

osm2streets/src/transform/collapse_short_road.rs
```diff
@@ -83,101 +78,78 @@ impl StreetNetwork {
...
-            let i = self.intersections.remove(&dst_i).unwrap();
-            if i.control == ControlType::TrafficSignal {
-                self.intersections.get_mut(&src_i).unwrap().control = ControlType::TrafficSignal;
+            let road = self.roads.get_mut(&r).unwrap();
+            if road.src_i == destroy_i.id {
+                road.src_i = keep_i;
+            } else {
+                assert_eq!(road.dst_i, destroy_i.id);
+                road.dst_i = keep_i;
             }
```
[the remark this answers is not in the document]
dabreegster: Aha, you're right! I forgot about this case; it's hard for me to picture without drawing it out or using the incremental debug steps. I added a comment for now. Since this PR is complex and you have something stacked on it, I vote for taking another round later at cleaning up some of these cases

osm2streets/src/transform/collapse_short_road.rs
```diff
@@ -15,36 +12,32 @@ use crate::{
...
         &mut self,
-        short: OriginalRoad,
-    ) -> Result<(
-        osm::NodeID,
-        osm::NodeID,
-        Vec<OriginalRoad>,
-        Vec<OriginalRoad>,
-    )> {
-        let (src_i, dst_i) = {
+        short: RoadID,
```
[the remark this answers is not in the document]
dabreegster: Done

dabreegster:
Thanks for the review! As followup later on, I'll look at:
- rethinking the roundabout collapsing logic and making it a transform
- removing paranoia checks in `collapse_short_road` that should be impossible now
- allowing loop roads to be collapsed for the two dual carriageway case

streets_reader/src/split_ways.rs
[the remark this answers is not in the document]
dabreegster: The latter does prune roadless intersections, yes. But it'd be cleaner (and probably cause less weird bugs in the future) if they didn't exist in the first place..

PR #123, opened 2022-11-22
Reshape IntersectionComplexity into IntersectionType

BudgieInWA:
*This PR builds on #122, so I've set the merge base there until that lands.*

After actually playing with intersection `complexity` and `control_level` a bit, thinking deeply about movements, and reading more code in detail during all these refactors, I've chipped away even further at what I think is useful to annotate an intersection with.

The primary purpose of `IntersectionType` is to describe what this node in the graph means to a driver, which coincides with different cases in the code, such as geometry calculations. I think it's the best place to represent `MapEdge`. `Connection` now represents the trivial two-road case, as well as the multi-road sausage-link case that used to be called `MultiConnection`. `Connection` intersections exist because our data model is a graph, not because there is any intersection that drivers have to consider. `Fork` represents a decision-point for the driver, but one where you keep moving, and `Intersection` is what you would actually call an "intersection".

Code-wise, working with a `Connection` and `Fork` is going to be very similar, but the fact that the number or directed-carriageways stays constant through a `Connection` seems significant enough for the distinction. The only further distinction I would like to make is between major-minor intersections (where some "major" road remains unbroken, and has priority) and "true" intersections (where lane marking are interrupted, and there is some area that is not owned by any one lane). The major/minor distinction is primarily a property of Movements, and I'll only think about incorporating it into `IntersectionType` if and when it seems useful.

### DIFF
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -4,9 +4,7 @@ As of November 2022, and probably incomplete. This describes how the codebase cu
 
 ## The model
 
-At its heart, a graph of roads and intersections. (Roads lead between exactly two intersections -- "road segments" might be more precise.) Roads have their lanes listed from left-to-right, with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
-
-Intersections have a `ControlType` -- stop signs, traffic signals, uncontrolled, etc. This is orthogonal to `IntersectionType` and `ConflictType`... TODO, narrow down valid combinations and give examples. MultiConnection vs Merge, please!
+At its heart, a graph of roads and intersections. A `Road` is a segment of road that leads between exactly two `Intersection`s. An `Intersection`'s type tells you if it represents a real-life intersection or some other kind of node in the graph. Roads have their lanes listed from left-to-right, each with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
 
 ### IDs
 
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -4,7 +4,9 @@ As of November 2022, and probably incomplete. This describes how the codebase cu
 
 ## The model
 
-At its heart, a graph of roads and intersections. A `Road` is a segment of road that leads between exactly two `Intersection`s. An `Intersection`'s type tells you if it represents a real-life intersection or some other kind of node in the graph. Roads have their lanes listed from left-to-right, each with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
+At its heart, a graph of roads and intersections. A `Road` is a segment of road that leads between exactly two `Intersection`s. An `Intersection`'s kind tells you if it represents a real-life intersection or some other kind of node in the graph. A `MapEdge` connects a single `Road` the edge of the map, a `Terminus` marks an actual dead end. A `Connection` joins multiple `Road`s together where there is no traffic interaction at all, whereas a `Fork` joins multiple roads that merge or diverge without any stop line. Finally, an `IntersectionKind::Intersection` represents everything that you would actually call an "intersection", where traffic merges with, diverges from or crosses other traffic.
+
+Roads have their lanes listed from left-to-right, each with a type, width, and direction. A lane represents any longitudinal feature of a road: travel lanes on the carriageway, separated bike and footpaths, street-side parking, and buffers, medians and verges. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
 
 ### IDs
 
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -6,7 +6,7 @@ As of November 2022, and probably incomplete. This describes how the codebase cu
 
 At its heart, a graph of roads and intersections. (Roads lead between exactly two intersections -- "road segments" might be more precise.) Roads have their lanes listed from left-to-right, with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
 
-Intersections have a `ControlType` -- stop signs, traffic signals, uncontrolled, etc. This is orthogonal to `IntersectionComplexity` and `ConflictType`... TODO, narrow down valid combinations and give examples. MultiConnection vs Merge, please!
+Intersections have a `ControlType` -- stop signs, traffic signals, uncontrolled, etc. This is orthogonal to `IntersectionType` and `ConflictType`... TODO, narrow down valid combinations and give examples. MultiConnection vs Merge, please!
 
 ### IDs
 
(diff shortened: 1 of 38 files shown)


osm2streets/src/transform/classify_intersections.rs
```diff
@@ -101,18 +100,15 @@ pub fn guess_complexity(
...
-            },
-            c,
-            movements,
-        ),
-    }
+    (
+        match worst_conflict {
+            Uncontested => Connection,
+            Diverge => Fork,
+            Merge => Fork, // TODO check for give way signs or count lanes to detect Intersections.
```
BudgieInWA, line 107: There are some cases that look like the "merge" version of a `Fork` when looking at the half-road/`Movement` level, but should be `Intersection`s when looking at the lane-level, or when considering yield signs. The classic example is a slip lane that merges into the destination road with a *Give Way* sign. [Here's an example](https://www.openstreetmap.org/node/1836742019) (without the give way sign tagged, as is common currently), and [just to the north](https://www.openstreetmap.org/node/31337970) is a sliplane that joins with a "lane added" sign, so should be a `Fork`.

forks and intersections

Another example I've seen is a [T-intersection between two oneway roads](https://www.openstreetmap.org/node/1886045849).

dabreegster:
Base PR is merged now and I haven't gotten any other massive refactors or nerd-swipes up my sleeve :)  Looking at this now

osm2streets/src/intersection.rs
```diff
@@ -21,8 +18,7 @@ pub struct Intersection {
     pub point: Pt2D,
     /// This will be a placeholder until `Transformation::GenerateIntersectionGeometry` runs.
     pub polygon: Polygon,
-    pub complexity: IntersectionComplexity,
-    pub conflict_level: ConflictType,
+    pub t: IntersectionType,
```
dabreegster, line 21: nitpick: Style wise it's constantly frustrating that `type` is a reserved word. We could use the special syntax `r#type` but I'd vote strongly against. Spelling out `intersection_type` is quite annoying. Tiny acronyms like `t` or `it` don't feel super readable. How do you feel about `kind`?

osm2streets/src/types.rs
```diff
@@ -120,52 +120,36 @@ pub enum ConflictType {
...
-    /// which retains its right of way.
+    /// Multiple `Road`s connect but no flow of traffic interacts with any other.
     ///
-    /// You would expect lane markings for the major road to continue (largely)
-    /// uninterrupted through the intersection, with a yield line ending the minor road.
-    Merge,
+    /// Usually one `Road` ends and another begins because the number of lanes has changed or some
+    /// other attribute of the roadway has changed. More than two `Road`s could be involved,
+    /// e.g. when a single carriageway (a bidirectional `Road`) splits into a dual carriageway
+    /// (two oneway `Road`s).
```
dabreegster, line 141: The "no flow of traffic interacts" part means that if we had a DC rejoining in a place where U-turns were explicitly allowed, it'd be an `Intersection` then?

docs/how_it_works.md
```diff
@@ -4,9 +4,7 @@ As of November 2022, and probably incomplete. This describes how the codebase cu
...
-
-Intersections have a `ControlType` -- stop signs, traffic signals, uncontrolled, etc. This is orthogonal to `IntersectionComplexity` and `ConflictType`... TODO, narrow down valid combinations and give examples. MultiConnection vs Merge, please!
+At its heart, a graph of roads and intersections. A `Road` is a segment of road that leads between exactly two `Intersection`s. An `Intersection`'s type tells you if it represents a real-life intersection or some other kind of node in the graph. Roads have their lanes listed from left-to-right, each with a type, width, and direction. Note osm2streets doesn't model bidirectional lanes yet -- sidewalks and shared center turn lanes are either forwards or backwards right now, and something downstream interprets them in a special way. (osm2lanes has more nuance, but isn't used in osm2streets yet.)
```
dabreegster, line 8: I'd still vote for a quick explanation here of `IntersectionType`. I think one picture/diagram with examples of all cases would do the trick. We don't need it now, but if you stumble on a small area with all examples, maybe we could just add labels

dabreegster (approved):
This new model is much simpler to think about!

osm2streets/src/types.rs
```diff
@@ -120,52 +120,36 @@ pub enum ConflictType {
     Cross,
 }
 
-/// What kind of feature an Intersection actually represents. Any connection between roads in the
-/// network graph is represented by an Intersection, but many of them are not "intersections" in
-/// the traditional sense.
-///
-/// This type might be better named IntersectionType and the distinction between MultiConnection,
-/// Merge and Crossing is of dubious value and will probably change in the future.
+/// What kind of feature an `Intersection` actually represents. Any connection between roads in the
```
dabreegster, line 123: Other optional style thing: some of these types could live in `intersection.rs`. I don't think we need loads of organization so far, but `road.rs` and `intersection.rs` have a reasonably clear split for many things

osm2streets/src/intersection.rs
BudgieInWA: I agree, and I really like `kind`.

Do we leave the name of the type itself alone, and work with the rust-ism that "type" is spelled "kind" in identifier contexts, or let rust have the term "type" all to itself, referring to the type system, and go with `IntersectionKind` (and `ConflictKind`)?

dabreegster: I'm not too sure. :\ I think I've seen `Kind` used in structs in other projects. Up to you. For consistency, I'd slightly lean towards `IntersectionKind`

osm2streets/src/types.rs
BudgieInWA: Yes, that's the spirit of it. I would expect some additional markings or something in that situation. In Perth I would never expect that, there would be a dedicated [U-turn lane](https://www.openstreetmap.org/way/669734507) or [dedicated area in the median](https://www.openstreetmap.org/way/29018596) for those manoeuvres.

If it turns out U-turns are allowed in situations that we still want to treat like `Connections`, then maybe U-turns become a special case.

dabreegster:
I was having a look at rolling out the "dog-leg intersection" transform everywhere. In one case, it incorrectly merges some stuff in a roundabout, but if we cutover to using `IntersectionType`, it'd work properly there! So, we should be able to remove https://github.com/a-b-street/osm2streets/blob/38aa927e75164d55cc00eb403423f67f3d081fcc/osm2streets/src/transform/find_short_roads.rs#L162.

Another bit of duplicate code is https://github.com/a-b-street/osm2streets/blob/38aa927e75164d55cc00eb403423f67f3d081fcc/osm2streets/src/transform/dual_carriageways.rs#L30. We could make this a simpler method on `Intersection` that trivially figures out the two one-way pieces and only makes sense for certain types.

osm2streets/src/types.rs
BudgieInWA: I will look at moving some bits around next.

docs/how_it_works.md
BudgieInWA: I have added a bit more explanation. I am weary of simply reproducing good docstrings here; and of turning this doc into yet another place that describes future ideals that haven't been exercised yet, instead of providing a structural introduction that sets the reader up to understand the code as they dive in.

osm2streets/src/intersection.rs
BudgieInWA: I found a bunch of uses of  "kind" used in exactly this way in rust docs, so I'm convinced this is a better term than "type" across the board.

BudgieInWA:
>  if we cutover to using `IntersectionType`, it'd work properly there!
> We could make [`MultiConnection`] a simpler method on `Intersection` that trivially figures out the two one-way pieces and only makes sense for certain types.

That's the idea! I think `IntersectionKind` will be the starting point for a lot of these cases, and other details will need to be interrogated too, but having a consistent classification for these kinds of properties has been my goal from the start. e.g. `if int.kind == IntersectionKind::Connection && int.roads.len() == 2 { .. }`.

BudgieInWA:
I have renamed the `Type` suffix out of a handful of enums. Let me know what you think. I'll also consider `intersection::Kind` and `intersection::Control` in my next change.

BudgieInWA:
I have merged this in order to put up a follow-on PR, but I'm still interested in feedback on the type renames.

docs/how_it_works.md
dabreegster: Thanks for the extra explanation, and yeah, this doc is meant as a code walkthrough prep. At some point a diagram would be nice, but we don't need it now

PR #125, opened 2022-11-23
Remove the ClassifyIntersection transformation in favour of always updated movements

BudgieInWA:
View the diff for each commit independently.

It turns out, because `recalculate_movements` was being called every time a road was added, the movements were almost completely filled in by the time we got to the transformation anyway. The only information that was missing was the turn restrictions, added in `split_ways.rs`. I changed `split_up_roads` a little bit, so that it doesn't use `StreetNetwork::insert_road`, but instead inserts the roads manually, then updates all the intersections once at the end.  (I think `split_up_roads` should become a method of `StreetNetwork` so `update_movements` and `sort_roads` can remain private. But I don't want to do that here, I will write that rational in another issue.)

Then I moved a bunch of functions into `road.rs` and `intersection.rs`. Most of the functions would have been best placed there anyway.

This change does not effect any tests!


osm2streets/src/render.rs
```diff
@@ -307,13 +307,20 @@ impl StreetNetwork {
...
+                                line.to_polyline()
+                                    .make_arrow(Distance::meters(0.5), ArrowCap::Triangle)
+                                    .to_geojson(Some(&self.gps_bounds)),
+                                make_props(&[]),
+                            ))
+                        }
+                    } else {
+                        warn!(
+                            "{} has movement ({}, {}) referring to road that is not connected",
+                            &intersection.id, a, b
```
dabreegster, line 322: Style nit: It should almost never be necessary to `&` or `*` in a `format!` style macro. https://doc.rust-lang.org/std/fmt/trait.Display.html borrows already

BudgieInWA: I swear I've had rust complain that I couldn't move into a `format!` type macro before... I must be misremembering. Maybe I was passing stuff to `make_props` or one of those formatting-adjacent helpers instead.

Good to know, thanks!

dabreegster: In case my comment above got lost, it's relevant here: do we need this check now, or can we assume no movements referring to invalid roads're gone?

dabreegster:
Good question... I think the warning is fine for interactive debugging use case. Maybe we ought to have a "consistency check" at the very end, or that tests run. I would want that to panic, so we can catch problems like this automatically, and then use interactive stepwise debugging to help figure out where things go wrong

BudgieInWA:
I like the idea of a consistency check that can be called whenever. Then including it in the test output and each debug step sounds super useful, especially as we work on data-flow aspects of the process.

I'll remove the check here, because the crash is more visible.

Note to self: add movements themselves to the test output.

BudgieInWA:
Oh, there are still stale movements going on... on `main` too. Something to do with `clip.rs` I think. 

I plan to add movements to the test output in another PR, then rebase this on top, to see where the problem has come in.

dabreegster:
https://github.com/a-b-street/osm2streets/blob/3d64f99d400b654367f40d86bee8678ade037e81/streets_reader/src/clip.rs#L71
I made this assumption in the opaque ID refactor recently. Maybe movements are getting assigned to a `MapEdge` but need to removed now?

BudgieInWA:
> https://github.com/a-b-street/osm2streets/blob/3d64f99d400b654367f40d86bee8678ade037e81/streets_reader/src/clip.rs#L71
> 
> I made this assumption in the opaque ID refactor recently. Maybe movements are getting assigned to a `MapEdge` but need to removed now?

This was exactly it! Because of the lack of movement visibility in the tests I thought I had fixed it. See b8ae9f1 for the solution (which we'll just have to remember). There was a similar problem with dual carriageway collapsing, fixed in 352723d.

Having this kind of problem is definitely the cost of maintaining (semi-)derived data in the intersections. I'm pretty happy with how it is at the moment though, and we should keep an eye out for ways to incorporate this data dependency into the `StreetNetwork` API (like in `insert_road`) in an ergonomic way.

BudgieInWA:
@dabreegster Please change the merge base and merge this is in after `movement_tests`. I'm heading off to help with a house-move over the weekend, so I'll be offline for a bit.

dabreegster (approved):
Rebase done!

> Having this kind of problem is definitely the cost of maintaining (semi-)derived data in the intersections

Good catch on the sausage link bug. Hopefully as we write more transformations, some patterns will emerge for making the mutations using slightly higher-level methods that can handle consistency better

PR #126, opened 2022-11-23
Simplify IntersectionControl, use IntersectionKind::MapEdge instead of Border

BudgieInWA:
`Border` has no place in `IntersectionControl` now that `MapEdge` covers it. Given the scope of `IntersectionControl`, as a summary of an entire intersection, I think the vagueness of `Signalled` and `Signed` is appropriate. The kind of sign at each arm of the intersection will need to live on the `Road` itself.

This was a pretty mechanical find-replace. The test diffs show the rename of 'StopSign' -> 'Signed', etc. and filling in a lot of 'Unknown' intersection types in `road_network.dot`.


osm2streets/src/intersection.rs
```diff
@@ -73,6 +73,6 @@ impl StreetNetwork {
 
 impl Intersection {
     pub fn is_border(&self) -> bool {
-        self.control == IntersectionControl::Border
+        self.kind == IntersectionKind::MapEdge
```
BudgieInWA, line 76: Thoughts on "border" vs "map edge". Should we rename one to be consistent with the other?

dabreegster: `MapEdge` is more clear to me. I chose "border" very long ago

dabreegster (approved):

PR #130, opened 2022-11-25
Start rendering sidewalk corners.

dabreegster:
#24. As a break from refactoring, something to start improving the visual output. When two sidewalks (or shoulders) meet up, we can extend the pavement into the intersection a bit and show how they join.

I started with the logic in A/B Street, but rewrote it from scratch to be simpler. A/B Street splits the problem into two steps -- generate a polyline representing the turning movement between sidewalk endpoints, then later render it. https://github.com/a-b-street/abstreet/blob/dfae1b54d56bd0e40d482e4120c985ec943c1f19/map_model/src/make/walking_turns.rs#L211 is part 1, and https://github.com/a-b-street/abstreet/blob/dfae1b54d56bd0e40d482e4120c985ec943c1f19/map_gui/src/render/intersection.rs#L307 is part 2. This is much more complex than this PR. One reason is that we insist on creating the turning movement even if the geometry is weird, and so the rendering has to handle weird geometry. In this PR since we're just rendering (for now), I've chosen to skip situations that don't match perfectly.

Some sample results, showing both good and bad:
Screenshot from 2022-11-25 14-21-10
Screenshot from 2022-11-25 14-21-01
Screenshot from 2022-11-25 14-21-44

This is of course preliminary, but hopefully a starting point to working on things like this, and figuring out how our estimation of where the pavement extends can affect vehicle movements around the corners.

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -89,6 +89,11 @@ impl JsStreetNetwork {
         self.inner.to_lane_markings_geojson().unwrap()
     }
 
+    #[wasm_bindgen(js_name = toIntersectionMarkingsGeojson)]
+    pub fn to_intersection_markings_geojson(&self) -> String {
+        self.inner.to_intersection_markings_geojson().unwrap()
+    }
+
     #[wasm_bindgen(js_name = toGraphviz)]
     pub fn to_graphviz(&self) -> String {
         // TODO Should we make the caller do the clone? Is that weird from JS?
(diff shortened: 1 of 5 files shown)


BudgieInWA:
I've not looked at any code yet. I like how the footpath edges matches up with the road edges (that is, when it looks like the original intersection geometry and the footpath geometry have used the same algorithm).

I also like how this reveals the actual road shape somewhere inside the intersection polygon! This prompted me to write up #138. I think we should be calculating intersection areas just for the road, then `shift`ing outwards for the footpaths.

dabreegster:
Should we merge this in as a first cut? After I detangle the current mess in #136, I'll look into trying the idea in #138 and doing this differently

BudgieInWA (approved):
Good to have this visualisation started.

PR #131, opened 2022-11-25
WIP: Rewrite dog-leg detection and rollout everywhere

dabreegster:
I was feeling pretty good about the results with this, but then found some false positive cases. I'm a bit stumped how to express the issue.

A good before - after:
Screenshot from 2022-11-25 14-46-08
Screenshot from 2022-11-25 14-46-12

A bad:
Screenshot from 2022-11-25 14-51-27
Screenshot from 2022-11-25 14-51-32

The intended dog-leg structure has the perpendicular road "continuing" on the other side of a main road. In the bad example, the "vertical side roads" are on the same side. Not sure how to formalize this yet. Any initial ideas, or pushback against the dog-leg approach generally?


dabreegster:
Reminder to self before finishing dog-leg: https://github.com/a-b-street/osm2streets/blob/670bd72acd6d82ef43f325b8f64b12f6700ba68e/osm2streets/src/transform/mod.rs#L47

BudgieInWA:
One thing that is clean from looking at the images, is that the intersection geometry algorithm doesn't do a great job with spread out intersections. I haven't looked understood the geometry algorithm yet, so I'm looking at this a pretty high level.

image

The top intersection looks pretty good, but because the internal `Road`s have been discarded, that middle southern arm doesn't have any obvious reference points from which to calculate its trim distance. In the bottom intersection, even the simplest case of two neighbouring and intersecting roads breaks.

I don't think it's necessary for the algo to be able to handle such spread out intersections in one go, but I think there will be situations that test it in these sorts of ways. 

The reason I bring this up, is that I think we can improve the situation by remembering the trim distance for each `Road` that we find in the first pass, and never making it any smaller in later passes, as the intersections are merged. This is reasonable, because the original intersection areas are a very good estimation of the overlap of the individual `Road`s, and thus which bits of the `Road` geometry is probably nonsense.

BudgieInWA:
I'm actually questioning the value of this transformation at all. What is the problem that this is trying to fix?

In the before of your good example, the east-west road is a "major" road that needs to continue through the intersections and each of the north-south roads are "minor" side roads. As we discussed before, I think we should continue with the model where the intersection takes over responsibility for the entire conflicting area (lets say the "intersection-max model") and learns how to draw lane marking for major roads that continue through it. (The alternative model ("intersection-min model") being one where the major roads continue until they meet, and the "side road intersection" only takes responsibility for drawing the outer edge of the road on the one side.)

One of the intersection-max model, is that `Road` rendering remains dead-simple. The intersections can even make use of it to draw major roads through intersection, which is what I started doing in the Lanes plugin.

The other benefit is that movements through the intersection is handled entirely within the intersection, and the movement along roads is handled exclusively by the road, with no special cases. Perhaps this is the problem you are trying to solve? The turn across traffic out of the narrow side road onto the wide main road will look pretty unnatural if it has to fit entirely within that skinny rectangle. If this is the problem you are trying to solve, then it is a problem even for T-intersections. I think the approach should be to detect this "thin rectangle" problem and extend the trim of the wider road.

In general, two intersections need to merged if they start overlapping, which the above approach will trigger when T-intersections are really close to each other. But I think we want to avoid merging intersections whenever we can. We want to merge all parts of a single logical intersection (the contiguous area where lane marking are gone) because interpreting the individual OSM nodes doesn't make sense without the context of the full intersection.

If you're trying to collapse dog-legs so that directions look like "continue across Main St" instead of "turn right at Main St then immediately left, back on North St" then I think we should solve that with a higher level "junction" concept that groups together `Intersection`s. This will be desired anyway, for stuff like "continue straight at the roundabout, using the second exit", and for intersections with slip lanes.

dabreegster:
Hmm, now I'm indeed doubting the point of this. Looking at another example... https://a-b-street.github.io/osm2streets/#20.14/47.63955/-122.31590 (and ignore the fragments of footpath)
Screenshot from 2022-11-26 11-58-39
Screenshot from 2022-11-26 11-59-05

All the streets involved are `residential`; there's not an obvious main or side road. My motivation for trying to merge this is partly rendering, but you're right that once we extend lane markings through intersections properly, the awkwardness of that tiny road segment goes away. An older reason is traffic simulation. Any road too short for a vehicle to fit fully is a source of gridlock, but if we have a logical "junction" grouping covering more of this, we can treat the sequence of movements as an atomic unit and not allow getting stuck somewhere in the middle.

So I guess one question becomes, when do we want to group the intersections together for some purpose? Just to the west, we have another candidate dog-leg, except the short road in between isn't really all that short:
Screenshot from 2022-11-26 12-03-57
Both examples I posted earlier feel kind of questionable now, not really the same thing.

The other confusion in my mind is still about how to treat a group of intersections together, once we decide on that group. The only concrete tool I know right now is to collapse the short road. But I'm starting to conceptualize how continuing lane markings through and logically treating movement sequences as atomic could work.

> We want to merge all parts of a single logical intersection (the contiguous area where lane marking are gone)

This feels like the key difference. https://www.openstreetmap.org/way/533573781 is a clear example of what we want to merge. It's physically in the middle of a clear intersection without lane markings. These dog-leg examples feel less clear. I'll hold off on this experiment and revisit whether it's a good idea at all.

> I haven't looked understood the geometry algorithm yet, so I'm looking at this a pretty high level.

Something I want to tackle soon is this. The code is absolutely a [haunted forest](https://john-millikin.com/sre-school/no-haunted-forests), though at least I wrote down how it works with lots of pictures. It's not that complicated from a high-level perspective, but weird edge cases and hacks accumulated over time. Many are likely irrelevant with new advances elsewhere, or should be revisited.

BudgieInWA:
> Any road too short for a vehicle to fit fully is a source of gridlock, but if we have a logical "junction" grouping covering more of this, we can treat the sequence of movements as an atomic unit and not allow getting stuck somewhere in the middle.

I think this is vital. Our `Intersection`s already have a lot of responsibility, and requiring any atomic routing manoeuvre to remain within a single intersection would be too onerous. An intersection would need to have a whole sub-network to model everything! I think we can avoid that by keeping intersections as small as possible (but we should merge logical intersections like your "clear example of what we want to merge" because we need the context of the whole intersection to understand it, just like road uses do) and introducing a `Junction` type which points to the parts of the network that make up the the logical junction.

Once we model major/minor movements (major movements having their lane markings continue) the routing should be able to allow cars to enter intersections and potentially block them if they are on a major movement.

I don't know quite how extensive a `Junction` should be, it depends how we need to use it. Anything that road users navigate in a single manoeuvre should be grouped. Having a single `Junction` where any two roads meet, grouping all of the slip lanes etc. seems useful (though for a different reason than your traffic sim routing). The `perth_stretched_lights` is an example that would benefit from grouping because all of those traffic lights are part of the same block (so searching for traffic light "islands" might be a heuristic).

In your first residential example, it's almost a moot point because there are no lane markings anywhere! So we should come to a very similar result if try to render it as two intersections vs one. Because there are no road markings, road users probably treat it as a single intersection. The equivalent of this is that there is no major/minor distinction, so we should see that tiny road segment as nonsense and collapse it. If east-west was a major road, and the others gave way, then the tiny road segment would connect two major movements and would thus have a sensible meaning, so it would not need not be merged.

dabreegster:
I got rid of dog-leg detection entirely, now that we properly handle roads trimmed into oblivion. We can revisit later, but I think the points you made here are correct

Issue #132, opened 2022-11-26
Simpler osm2lanes editor


dabreegster:
Eventually https://a-b-street.github.io/osm2lanes/editor/ will be a way to more easily fix lane tagging visually. Right now it's focused on a single way and doesn't show a map at all.

After a morning of manually fixing mistagged lane count and bus lanes using street explorer, I realized we could do something much simpler that'd be very quick to build and would immediately help out people who already know OSM tagging.

1) Something like street-explorer, but that just lets you pan around OSM and import a new area. Only show the lane marking layer; no layer toggling
2) After rendering somewhere, click a road segment. We'd check OSM way IDs and zoom out to highlight all segments of that same way
3) Pop up a key/value string tag table and let people manually edit
4) Regenerate lane specs and re-render after updates
5) Press a button to download a `.osc` file to open in JOSM, or later submit the change directly

This just smooths out the workflow I'm already doing right now manually, and I think would be a big step forward for the low effort required. And of course when we call osm2streets in this case, run minimal transformations; we want to edit OSM and so want to see as close a picture as possible. And in fact, if you click a road segment with multiple way IDs, we can bail out, since we somehow made some transformation there.

dabreegster:
https://github.com/a-b-street/osm2streets/tree/lane_editor
Something wicked (simple) this way comes...
https://user-images.githubusercontent.com/1664407/204092426-b4fe329a-759e-450f-b1c5-86053818756f.mp4

Robinlovelace:
Wow, that video above looks awesome! I think I was testing the wrong tool but FYI this is what I found when trying to add a car storage lane to the northern end of Scott Hall Road:


Uploading 2022-11-26 14-24-00.mp4…

Issues: diff seems to have removed the road name and no way to change type of parking?

Source: https://www.openstreetmap.org/way/23162214#map=17/53.84446/-1.53754 and https://www.google.com/maps/@53.8452842,-1.5364045,3a,75y,36.1h,60.02t/data=!3m6!1e1!3m4!1s4LqhDwLfY5N9jDLEOlCs0g!2e0!7i13312!8i6656

dabreegster:
This tool isn't deployed yet and the osm2lanes one doesn't work yet. Little
more patience :)

BudgieInWA:
[...]

PR #133, opened 2022-11-27
Start a simple visual OSM tags editor

dabreegster:
See #132 for the idea. If we assume the user knows OSM tagging, we can just let them edit tags and re-render with osm2streets. This didn't take long to put together, and another few hours of work would get it to the point where I would use it for real to fix a bunch of bus lane tagging.

https://user-images.githubusercontent.com/1664407/204142669-e8057989-90bc-4708-9a89-03e5bef77e83.mp4

Some quick next steps:
- Generate and download a `.osc` file with the changes, which can be opened in JOSM and uploaded
- Click off a road to reset the view
- Don't allow editing a road that's clipped, because the user won't know how far it stretches and if the lanes are correct
- Indicate the forwards direction of the way, to help with tagging

Longer term:
- Do oauth and submit OSM changes directly to the API
- Split a way when lanes change
  - This is hard to code up just at the OSM data model level, because relations have to be maintained. IIRC I've seen @tordans start a conversation somewhere around common code to do that. Happen to know anything we could use?
- Bring in the osm2lanes cross section cards view and don't require editing the tags manually at all

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 use abstutil::{Tags, Timer};
 use geom::PolyLine;
@@ -19,8 +19,7 @@ pub struct ImportOptions {
 #[wasm_bindgen]
 pub struct JsStreetNetwork {
     inner: StreetNetwork,
-    tags_per_way: HashMap<osm::WayID, Tags>,
-    original_geometry_per_way: HashMap<osm::WayID, PolyLine>,
+    ways: BTreeMap<osm::WayID, streets_reader::osm_reader::Way>,
 }
 
 #[wasm_bindgen]
@@ -75,17 +74,9 @@ impl JsStreetNetwork {
             street_network.apply_transformations(transformations, &mut timer);
         }
 
-        let mut tags_per_way = HashMap::new();
-        let mut original_geometry_per_way = HashMap::new();
-        for (id, way) in doc.ways {
-            tags_per_way.insert(id, way.tags);
-            original_geometry_per_way.insert(id, PolyLine::unchecked_new(way.pts));
-        }
-
         Ok(Self {
             inner: street_network,
-            tags_per_way,
-            original_geometry_per_way,
+            ways: doc.ways,
         })
     }
     #[wasm_bindgen(js_name = toGeojsonPlain)]
@@ -140,7 +131,7 @@ impl JsStreetNetwork {
     // HashMap
     #[wasm_bindgen(js_name = getOsmTagsForWay)]
     pub fn get_osm_tags_for_way(&self, id: i64) -> String {
-        abstutil::to_json(&self.tags_per_way[&osm::WayID(id)])
+        abstutil::to_json(&self.ways[&osm::WayID(id)].tags)
     }
 
     /// Returns a GeoJSON Polygon showing a wide buffer around the way's original geometry
@@ -158,7 +149,8 @@ impl JsStreetNetwork {
             .unwrap();
 
         // Show a wide buffer around the way
-        let polygon = self.original_geometry_per_way[&id].make_polygons(1.5 * width);
+        let polygon =
+            PolyLine::unchecked_new(self.ways[&id].pts.clone()).make_polygons(1.5 * width);
 
         abstutil::to_json(&polygon.to_geojson(Some(&self.inner.gps_bounds)))
     }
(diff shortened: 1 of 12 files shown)


osm2streets-js/src/lib.rs
```diff
@@ -116,6 +130,53 @@ impl JsStreetNetwork {
...
+
+    /// Modifies all affected roads and only reruns `Transformation::GenerateIntersectionGeometry`.
+    #[wasm_bindgen(js_name = overwriteOsmTagsForWay)]
+    pub fn overwrite_osm_tags_for_way(&mut self, id: i64, tags: String) {
+        let id = osm::WayID(id);
+        let tags: Tags = abstutil::from_json(tags.as_bytes()).unwrap();
+
+        for road in self.inner.roads.values_mut() {
+            if road.osm_ids.iter().any(|x| x.osm_way_id == id) {
+                // TODO This could panic, for example if the user removes the highway tag
```
dabreegster, line 169: Either this method should return `Result` or we can catch the panic and bubble up an error message

street-explorer/www/js/lane_editor.js
```diff
@@ -0,0 +1,205 @@
+import { downloadGeneratedFile } from "./files.js";
```
dabreegster, line 1: I adapted this from street explorer and am sharing some common code. Does that seem OK and maintainable?

street-explorer/www/js/lane_editor.js
```diff
@@ -0,0 +1,205 @@
...
+      if (this.map.getZoom() < 15) {
+        window.alert("Zoom in more to import");
+        return;
+      }
+
+      await this.importView(importButton);
+    };
+  }
+
+  async importView(importButton) {
```
dabreegster, line 32: Some of this code is duplicated, but the entire thing in street-explorer isn't appropriate. Maybe the overpass fetch could be a helper. Thinking in terms of components, an "import current view" button could be common, providing a callback / event for "here's some XML fetched"

street-explorer/www/js/lane_editor.js
```diff
@@ -0,0 +1,205 @@
...
+      const resp = await fetch(url);
+      const osmXML = await resp.text();
+
+      importButton.innerText = "Importing OSM data...";
+
+      this.network = new JsStreetNetwork(osmXML, {
+        debug_each_step: false,
+        dual_carriageway_experiment: false,
+        cycletrack_snapping_experiment: false,
+        inferred_sidewalks: false,
```
dabreegster, line 56: Match raw OSM as close as possible

street-explorer/www/js/lane_editor.js
```diff
@@ -0,0 +1,205 @@
...
+      })
+    );
+    this.layers.push(
+      L.geoJSON(JSON.parse(this.network.toLanePolygonsGeojson()), {
+        style: lanePolygonStyle,
+        // Make lanes clickable
+        onEachFeature: (feature, layer) => {
+          layer.on({
+            click: (ev) => {
+              if (feature.properties.osm_way_ids.length != 1) {
```
dabreegster, line 98: This shouldn't happen if we run no transformations, but who knows

street-explorer/www/js/lane_editor.js
```diff
@@ -0,0 +1,205 @@
...
+      {
+        style: (feature) => {
+          return { stroke: false, fill: true, color: "red", opacity: 0.5 };
+        },
+      }
+    ).addTo(this.map);
+
+    var html = `<table><tbody id="tags-table">`;
+
+    const tags = JSON.parse(this.network.getOsmTagsForWay(id));
```
dabreegster, line 135: There's a mix of setting `innerHTML` and using DOM methods here. It's a bit messy, but seems correct.

Maybe if iD ever has a standalone library for something like the tags table component, it'd be cool to automatically share support for stuff like autocomplete for tag values.

street-explorer/www/js/layers.js
```diff
@@ -49,7 +49,7 @@ export const makePlainGeoJsonLayer = (text) => {
   });
 };
 
-export const makeLanePolygonLayer = (text) => {
+export const lanePolygonStyle = (feature) => {
```
dabreegster, line 52: Just splitting this out so we can render lanes without getting the popup behavior

street-explorer/www/js/leaflet.js
```diff
@@ -0,0 +1,42 @@
+export function setupLeafletMap(mapContainer) {
```
dabreegster, line 1: Just moving common code between the new tool and street explorer

street-explorer/www/lane_editor.html
```diff
@@ -0,0 +1,52 @@
...
+    <script src="js/deps/geosearch.umd.js"></script>
+    <script src="js/deps/leaflet-hash.js"></script>
+
+    <script type="module">
+      import { LaneEditor } from "./js/lane_editor.js";
+      window.app = new LaneEditor(document.getElementById("map"));
+    </script>
+  </head>
+  <body>
+    <aside style="max-width: 35rem">
```
dabreegster, line 28: I copied street explorer, but increased this to fit the table

streets_reader/src/osm_reader/reader.rs
```diff
@@ -182,10 +182,6 @@ fn read_tags(reader: &mut Peekable<ElementReader>) -> Tags {
         let obj = reader.next().unwrap();
         let key = obj.attribute("k");
         let value = obj.attribute("v");
-        // Filter out really useless data
```
dabreegster, line 185: This was intended as a file size optimization, but it's inappropriate at this layer. That becomes very obvious now!

tordans:
> Split a way when lanes change
> This is hard to code up just at the OSM data model level, because relations have to be maintained. IIRC I've seen @tordans start a conversation somewhere around common code to do that. Happen to know anything we could use?

Yes, that is https://github.com/ideditor/id-sdk/issues/165. However, nothing came out of this, yet. Maybe another comment and a direct question to `@tyrasd` (current maintainer of iD) if he could see himself extracting this part of the code would help…

dabreegster:
OK, this now generates a `.osc` file that can be uploaded with JOSM. Plenty of rough edges, but I'm now using this for real, and it's much faster than my previous workflow for quick fixes!

dabreegster:
@BudgieInWA, any opposition to merging this in so far? I've been running locally and slowly using it for real.

BudgieInWA (approved):
Go for it, looking good!

Issue #68, opened 2022-08-16
Bindings for Java


dabreegster:
Let's start the equivalent of [osm2street-js](https://github.com/a-b-street/osm2streets/tree/main/osm2streets-js), but for Java. At an extremely cursory glance, https://github.com/jni-rs/jni-rs looks popular, but there are likely many options out there

dabreegster:
https://blog.frankel.ch/start-rust/7/ and https://docs.rs/jni/latest/jni/ look straightforward to me. I'll try exposing the same `osm2streets-js` API like this. Not sure how packaging / publishing works, but not important yet

dabreegster:
Actually, https://github.com/dushistov/flapigen-rs might be a higher-level approach

BudgieInWA:
I am able to build and run the build native library locally! My next challenge is to using it as a library in the JOSM2Streets plugin. I have the JOSM plugin building using gradle, but I only just have a grasp of how that build system is working, and don't know how distributing the built plugin will work yet. Including the native library will be tricky, but I think I there is a satisfying solution:

It looks like we can package the native libraries in a `.jar` while building `osm2streets-java`, and modify the JNI class to load the appropriate library out of the JAR, according to a couple of the answers at https://stackoverflow.com/questions/2937406/how-to-bundle-a-native-library-and-a-jni-library-inside-a-jar. Maybe this library will be good: https://github.com/scijava/native-lib-loader?

Then I can copy the `.jar` into the JOSM plugin project and add it as a dependency in the standard way (which I hope is easy). Publishing the bundled JAR to Maven should also be "straight forward".

dabreegster:
`native-lib-loader` indeed looks great. Do callers need to depend both on it and our jar with all the native builds, or can we have a single library that depends on this loader and calls it? I'm still a bit confused how this all wires together; I wonder if there's a real example of a project out there with everything set up.

Another question is how to build for all the platforms sanely. We can do this with Github actions, but I've never attempted to automatically bundle resulting files from each platform into one final asset.

BudgieInWA:
My understanding is that the result is a completely self contained `.jar`, with all of the JNI stuff as an implementation detail.

[...]

PR #134, opened 2022-11-28
Use jni-rs to generate Java bindings. #68

dabreegster:
Proof of concept! What's the minimum needed to use this in https://github.com/BudgieInWA/joSM2Streets/? What's the simplest modern way to publish to Maven or similar?


osm2streets-java/StreetNetwork.java
```diff
@@ -0,0 +1,24 @@
...
+import java.nio.file.Paths;
+
+class StreetNetwork {
+	static {
+		System.loadLibrary("osm2streets_java");
+	}
+
+	private static native long create(String osmXmlInput);
+
+	private static native String toGeojsonPlain(long pointer);
```
dabreegster, line 11: I gave up fighting to make these normal methods on a `StreetNetwork` object. The example code just passes are raw `long` pointers: https://github.com/jni-rs/jni-rs/blob/f8e78e3583ce850c2699c5e72cb8517d8fff09a6/example/HelloWorld.java#L7.
I think we can do this properly with `jobject`, but I'll leave that for a followup

osm2streets-java/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+javac StreetNetwork.java
+cargo build
+java -Djava.library.path=../target/x86_64-unknown-linux-gnu/debug/ StreetNetwork
```
dabreegster, line 6: You probably need to change this on you end; it's likely just `../target/debug`. I have the architecture bit in there so compiling native and wasm32 doesn't stomp on the build directories

osm2streets-java/StreetNetwork.java
```diff
@@ -0,0 +1,24 @@
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+class StreetNetwork {
+	static {
+		System.loadLibrary("osm2streets_java");
```
dabreegster, line 6: You may need to prepend `lib` here. I had to fiddle around to get this to work. The actual file is `libosm2streets_java.so` for me

osm2streets-java/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+                .unwrap();
+        let transformations = Transformation::standard_for_clipped_areas();
+        network.apply_transformations(transformations, &mut timer);
+
+        Self { inner: network }
+    }
+}
+
+#[no_mangle]
+pub extern "system" fn Java_StreetNetwork_create(
```
dabreegster, line 29: These signatures have to match the `.h`

osm2streets-java/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "system" fn Java_StreetNetwork_toGeojsonPlain(
+    env: JNIEnv,
+    _: JClass,
+    pointer: jlong,
+) -> jstring {
+    let streets = &mut *(pointer as *mut StreetNetwork);
+    let result = streets.inner.to_geojson().unwrap();
```
dabreegster, line 46: Once we figure out how to use a normal Java object, we can hopefully just call normal methods in the `impl` block above

BudgieInWA:
This runs for me, after tweaking the commands.

I don't know at all about packaging/publishing on maven, or packaging the built binary. I have to reference the class in the plugin build, which is currently targeting java 11 for some reason.

I'd be keen to jump on a call to try to figure it out some time.

dabreegster:
I'll start reading about the packaging process. And a call would be good; it's maybe also time to consider putting the WASM+JS module on NPM and talking about versioning. Next week is pretty open on my end so far

dabreegster:
Thanks to help from https://github.com/jni-rs/jni-rs/issues/394, the Java API is now sensible. If there's no opposition, I'll merge this tomorrow and look into Maven.

BudgieInWA (approved):
Looks good!

I was thinking of trying to manually include the dependency (via the .class file or something), but that means figuring out how to get the lib portion in there. Looking into publishing for Maven might actually be the best place to find the answers to those problems.

PR #135, opened 2022-11-28
Fix collapse short_short_road

dabreegster:
Individual commits are easier to view. A few changes:

- Move two methods out of `transform/` into a new `operations/` directory. I'm not opinionated about the naming, but it feels like these more complex operations should be grouped clearly.
- Remove the OSM link from all the `test.json` files, now that street explorer renders tests directly. That entire file is a bit redundant; it could just as easily be a `.txt` or something. Honestly I feel like filing issues in github referring to tests and talking about expected vs current behavior could be nicer. Should we just delete these?

And the big one: Follow up to #122 "Consider the standard 4-way dual carriageway" thread. I added a test case for this, and before this change, the intersection geometry algorithm actually breaks and we just draw a circle.


osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -26,36 +26,18 @@ impl StreetNetwork {
             (r.src_i, r.dst_i)
         };
 
-        // If either intersection attached to this road has been deleted, then we're probably
```
dabreegster: This check is no longer needed. We should never have a road that refers to non-existent intersections, ever. If we do, something below will crash, and we can track down the problem manually. (Maybe with that consistency check idea)

dabreegster (commented):
I'm going to merge this now to pull into A/B Street, but happy to address feedback of course

Issue #127, opened 2022-11-23
Simplify the early stages of creating a `StreetNetwork` from OSM data


BudgieInWA:
The start of main `osm_to_street_network` function is a complicated tangle of stateful modifications of the `StreetNetwork` with plenty of comments hinting at dependencies between seemingly unrelated steps. The delegation of responsibilities and relationship between `StreetNetwork::blank`, `extract_osm`, `split_up_roads` and `clip_map` is not clear, with mutations happening to the `StreetNetwork` all over the place.

https://github.com/a-b-street/osm2streets/blob/3d64f99d400b654367f40d86bee8678ade037e81/streets_reader/src/lib.rs#L102-L115

I think we should improve the public APIs of `StreetNetwork` and `OsmExtract` so that they help explain the ordering constraints and delegation of responsibilities. I'll share my ideas about that here, which requires a diversion into clipping and the `StreetNetwork` to OSM mapping.

## Clipping the Map

Clipping is an interesting problem because the interpretation of the map features is influenced by nearby features, that's the whole reason osm2streets needs to exist. I think that `StreetNetwork` ought to treat its features fundamentally as areas, not centerpoints and centerlines. (That's why I'm advocating for geometry calculations to move up earlier in the process.) If that is the case, then precisely clipping a `StreetNetwork` has to involve cutting `Road`s and `Intersection`s in half. This is not something that we should support at the representation level, because the representation still needs to use points and centerlines as reference for the distinct features (intersections and roads) that make up the areas.

In fact, it is an unavoidable problem that we can't accurately represent anything that is close to the edge of the raw data we have. We should acknowledge that fact and build it into our understanding and the API. There should be a buffer border of some distance around the area that we have data for that we consider to be nonsense output. Probably about half the width of a very wide road.

When doing the OSM to Streets conversion, I think we should trim the raw OSM to the boundary polygon in `OsmExtract`, before adding it into `StreetNetwork`. The trimming operation is better defined in the domain of the OSM data itself, because the OSM data is points and lines. There should be no clipping operation done on `StreetsNetwork` representation data, instead we could clip the end-result geometry precisely to the inside edge of the buffer boundary, or output that boundary and let the consumer do it.

## Mapping Between `StreetNetwork` and OSM

[...]

PR #137, opened 2022-11-28
Scrub A/B Street fields and half-baked logic

dabreegster:
#127 

Elevation, crossings, and barriers are all concepts not part of osm2streets (yet). They have real use in A/B Street today, but we can simplify this codebase with proper layering. Crossings and barriers are both things that will wind up in osm2streets eventually, but the half-baked version that exists now can just stay in A/B Street.

### DIFF
--- a/osm2streets/src/intersection.rs
+++ b/osm2streets/src/intersection.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 
-use geom::{Distance, Polygon, Pt2D};
+use geom::{Polygon, Pt2D};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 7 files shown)


BudgieInWA (approved):
Oh, wow. Having a good layering solution is such a powerful set up. A/B Street can demonstrate a use-case for some layers before they are tackled here. Great to these huge deletions, knowing the logic lives on elsewhere.

dabreegster:
Things could've probably started this way, but the initial split of code into this repo felt complicated at the time. Hindsight is showing it's not hard at all for the caller of `streets_reader` to also extract stuff from OSM and use it along with `StreetNetwork` later on

Issue #136, opened 2022-11-28
Intersection geometry refactor


dabreegster:
There are a few ideas rumbling around for improving intersection geometry code. The current thing is quite a scary mess. In some first attempts to detangle it, I think I've found some smaller cleanup steps that should happen in some order first.

- [ ] Work in-place, instead of taking `InputRoad` and returning `Results`. Every road should have start and end trim distances, which can both be passed in (the results from the previous round) and modified by the algorithm.
- [ ] Remove `intersection.point`; it's a meaningless field after the first transformation. We can set the initial intersection polygon guess to be a circle there, maybe.
- [ ] Get rid of `trim_roads_for_merging` and instead store start/end trim distance on every road
- [ ] Dedupe `RoadLine`, `Piece`, and maybe `Edge` from #130.

I think some/all of the above should happen before we attempt to remove the intersection geom transformation and instead always keep things updated as we make mutations elsewhere in the codebase.

More dramatic ideas:

- [ ] Radically simplify `generalized_trim_back`. We should only look for collisions between adjacent pairs of road edges.
- [ ] Revisit the whole `second_half` mess and the problems with two polylines hitting each other at multiple points
- [ ] Get rid of the `deduped.sort_by_key` stuff at the end if possible. See what tests break without it. We _should_ be able to produce the polygon points in the correct order, if our input roads are sorted correctly.
- [ ] Make the `pretrimmed_geometry` case less weird. Maybe take the main algorithm and first go update `trim_start` and `trim_end` everywhere. Then have a second method that just produces the polygon from that.
- [ ] Reconsider what we do for deadend intersections in general and simplify that method
- [ ] Think through `on_off_ramp` from scratch. Maybe this is triggered by `IntersectionKind` now and has nothing to do with highway type. Maybe there are better ideas to generate the geometry here.

BudgieInWA:
Without knowing the details of the current algorithm, I am heartened to read these individual points, because a lot of them make sense to me! That first list of steps all sound great to me (though I can't comment on `RoadLine`, `Piece`, or `Edge`) and iterating within the current transformation structure sounds sensible too.

I'm imagining one PR that adds `trim_start` and `trim_end` to `Road`, calculates it for the full network before any `Intersection`s are merged, (by only considering adjacent `Road`s) and having the intersection geometry just use `trim_start` and `trim_end`. Transformations should only ever increase trim distance for now - `Intersection`s will draw lane markings to compensate for some overly large trim distances.

[...]

Issue #91, opened 2022-09-15
Support direction = both ways


dabreegster:
There are many cases where bidirectional travel is allowed on one "lane." We should add that to the `Direction` enum, then make all caller code handle it properly. In osm2streets itself, the impact isn't huge. But many places downstream in A/B Street will be affected: rendering, routing, traffic simulation, and even blockfinding. Just starting this issue to start tracking notes on how to fix things there

BudgieInWA:
An "alternates" case too. I called it [TravelDirection in my types](https://github.com/a-b-street/osm2streets/blob/a19ab6ac8ef1dc4b812830fc1eae11687e28356c/osm2streets/src/units.rs#L22-L33).

Most reads of the value will boil down to "is travel supported in this direction", right? Routing could implement "alternating" and use that for "both ways" too.

dabreegster:
> An "alternates" case too. I called it [TravelDirection in my types](https://github.com/a-b-street/osm2streets/blob/a19ab6ac8ef1dc4b812830fc1eae11687e28356c/osm2streets/src/units.rs#L22-L33).

Ah right. I guess this fits signalized one lane "under construction" cases, or super narrow tunnels. And maybe even reversible / contraflow highways.

> Most reads of the value will boil down to "is travel supported in this direction", right?

I think so, but there are a few abuses of direction in A/B Street that I need to sort out. They started before I knew when to reason in terms of left/right or inside/outside. For example, sidewalks have a forwards/backwards direction that has to be set a certain way, or a bunch of things break (people walking on the wrong side, tracing the geometry of curbs, people transferring between a bus or car and the sidewalk).

A funny in-between example is parking lanes. Are they directional or not? In many places, legally you're supposed to park with the direction of traffic, even if it's not enforced much. But I've seen things like center parking lanes separating the two directions of traffic. Or center turn lanes (ref https://github.com/a-b-street/osm2lanes/issues/184).

PR #139, opened 2022-11-30
Parse `placement` tags and shift roads accordingly

BudgieInWA:
I have created the types `RoadPosition` and `Placement`, which I believe captures everything that we will need to contend with as we flesh out the support for placement. I see `RoadPosition` being useful in general, for talking about and working with different road positions. E.g. it might make sense to add `LeftEdge` and `RightEdge` etc.

I have renamed the two center line fields in order to see how it feels this way. The idea is that `reference_line` remains where the osm way is because it might be useful to have easy access to that. The main benefit I see is that the ends will all join up in the same way they do in OSM, which has some nice planar-graph properties, useful for sorting clockwise order, for instance.  After shifting the lines, they may cross over each other in confusing ways (especially the nonsense fist/last segments). Though maybe it's better to refer back to the original ways in `OsmExtract` for cases like that, I don't know yet.

I haven't followed through all the uses of `reference_line` and `center_line` yet - I wanted your input on that - to make sure they are using the right one in the right way. We have some decisions to make about this, in particular deciding what transformations should work with/modify, and if we should drop `reference_line` . Hopefully the concepts and terminology make it clear what should be used at each site.

We will need to calculate the appropriate `Placement::Varying` for `Road`s with `Placement::Transition` after we have determined the lane connectivity. Or maybe make up our own geometry for such roads that join a `Connection` or `Fork`, because the reference line often goes outside the road completely.

I haven't inspected or accepted the test diffs yet either, except for the new dedicated placement test, that looks great when it works, and broke one intersection.

placement before

placement after


osm2streets/src/lanes/mod.rs
```diff
@@ -87,6 +88,47 @@ impl LaneType {
...
+            LaneType::LightRail => false,
+            LaneType::Buffer(_) => false,
+            LaneType::Footway => false,
+            LaneType::SharedUse => false,
+        }
+    }
+
+    /// Determines if the lane is part of the roadway, the contiguous sealed surface that OSM
+    /// mappers consider the "road".
+    pub fn is_roadway(&self) -> bool {
```
BudgieInWA, line 114: This function (and `is_tagged_by_lanes_suffix` above) might need to be moved to `LaneSpec`, with some extra info stored to answer them properly. Or this gets clearer when we improve `LaneType` to tease apart the multiple different properties it is trying to represent.

The `is_roadway` concept that I use here might need to be a field on `LaneSpec` that is determined at parsing time.

osm2streets/src/lanes/mod.rs
```diff
@@ -87,6 +88,47 @@ impl LaneType {
         }
     }
 
+    /// Determines if the lane is a travel lane that is represented in OSM `*:lanes` tags.
+    /// Note that the `lanes` tag counts car driving lanes, excluding bike lanes, whereas the
+    /// `:lanes` suffix specifies that each lane, including bike lanes, should have a value between
+    /// `|`s. This function identifies the latter kind.
```
BudgieInWA, line 94: I don't know if you were aware of this, it's not very well presented on the wiki and has big implications in `osm2lanes` parsing.

osm2streets/src/lanes/mod.rs
```diff
@@ -355,3 +397,81 @@ impl fmt::Display for Direction {
...
+    /// The center of the separation between both directions of traffic, i.e. the dividing line,
+    /// median, or shared turning lane. For a oneway road, this is the "inside" edge of the road,
+    /// i.e. the right side of LHT and the left side of RHT.
+    Separation,
+    /// On the left edge of the named lane (from the direction of the named lane).
+    LeftOf(LtrLaneNum),
+    /// In the middle of the named lane.
+    MiddleOf(LtrLaneNum),
+    /// On the right edge of the named lane (from the direction of the named lane).
+    RightOf(LtrLaneNum),
```
BudgieInWA, line 451: The other way to do this could be `ForwardLane(usize, LanePosition)` and `BackwardLane` with `enum LanePosition { Left, Right, Middle }`.

osm2streets/src/lanes/mod.rs
```diff
@@ -355,3 +397,81 @@ impl fmt::Display for Direction {
...
+        use LtrLaneNum::*;
+        match self {
+            Forward(n) => Backward(*n),
+            Backward(n) => Forward(*n),
+        }
+    }
+}
+
+/// Identifies a position within the width of a roadway. Lanes are identified by their left-to-right
+/// position, as per the OSM convention.
```
BudgieInWA, line 432: Early on I was tempted to represent this using `Inside`/`OutsideOf(i32)` naming lanes relative to the separation (with negative indicating lanes on the backward side) and adding an invariant to `Road` that it stores its separation. After working with the minutia of calculating things from the left edge using `lanes_ltr` I'm more comfortable with it - in big part because my understanding is that in the presence of contraflow lanes, `*:lanes:forward` tags, and thus `placement:forward`, refer to the forward lanes in lrt order, regardless of their position.

osm2streets/src/lanes/mod.rs
```diff
@@ -355,3 +397,81 @@ impl fmt::Display for Direction {
...
+///
+/// Most commonly seen as a value of the placement tag, e.g.
+/// `placement=right_of:1` means that the OSM way is drawn along the right edge of lane 1.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum RoadPosition {
+    /// The center of the carriageway width, ignoring lanes. The default placement of OSM ways.
+    Center,
```
BudgieInWA, line 439: Seems awkward that this type uses "center" and "middle" to mean the same thing, but "center" matches with the convention in this repo, and "middle" matches the placement tag. Opinion?

osm2streets/src/lanes/placement.rs
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                } else {
+                    Err(())
+                }
+            }
+        }
+    }
+}
+
+impl TryFrom<&Tags> for Placement {
```
BudgieInWA, line 39: I am very happy with treating the parsing step of the whole scheme independently from seeing if the semantics of the tag values match the road they are tagged on.

Do you think I should back out from implementing the traits though? Or is this the sort of thing you are interested in working towards?

osm2streets/src/road.rs
```diff
@@ -219,7 +259,8 @@ impl Road {
         // If there's a sidewalk on only one side, adjust the true center of the road.
         // TODO I don't remember the rationale for doing this in the first place. What if there's a
         // shoulder and a sidewalk of different widths? We don't do anything then
-        let mut true_center = self.untrimmed_center_line.clone();
+        //TODO use self.left_edge_offset_of(RoadPosition::Center) or something?
+        let mut true_center = self.reference_line.clone();
```
BudgieInWA, line 263: I hope you can figure out the correct semantics here, depending on how `untrimmed_geometry` is used.

BudgieInWA (commented):
I've left out propagation of actual errors to keep things as clear as possible for now, and I don't know what error reporting approach we want to build towards.

osm2streets/src/lanes/mod.rs
```diff
@@ -355,3 +356,73 @@ impl fmt::Display for Direction {
         }
     }
 }
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum LtrLaneNum {
```
dabreegster, line 402: Before I read the below comment, I was confused what this meant. Always counting from the left, its the 0th, 1st, 2nd, etc lane that's forwards or backwards, regardless of contraflow. What about `lanes:both_ways`?

osm2streets/src/lanes/mod.rs
```diff
@@ -355,3 +356,73 @@ impl fmt::Display for Direction {
         }
     }
 }
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
```
dabreegster, line 401: nitpick: I'm guessing you copied the `derive`. Do we need to hash, serialize, order these?

osm2streets/src/road.rs
```diff
@@ -239,6 +278,137 @@ impl Road {
         self.lane_specs_ltr.iter().map(|l| l.width).sum()
     }
 
+    /// Calculates the number of (forward, both_ways, backward) lanes. The order of the lanes
+    /// doesn't matter.
+    pub fn driving_lane_counts(&self) -> (usize, usize, usize) {
```
dabreegster, line 283: `driving_lane` makes me think `LaneType::Driving`, but you're looking for the lanes suffix

dabreegster (commented):
I'll take another shot at reviewing in the morning, I can't process most of this right now. Responding to some of your comments though

dabreegster:
> Though maybe it's better to refer back to the original ways in OsmExtract for cases like that, I don't know yet.

I like the new naming. And for now, keeping `reference_line` here seems fine.

> We have some decisions to make about this, in particular deciding what transformations should work with/modify, 

It'd probably be better to work off the trimmed and transformed `center_line` in transformations generally. Things like determining if a road is short or not doesn't really care about untransformed center. Things like `estimate_trimmed_geometry` are hacks right now because we're not calculating trimmed geometry early yet. I'm hoping to change that in #136, but likely this PR lands first

osm2streets/src/lanes/mod.rs
dabreegster: Adding to `LaneSpec` and/or refining `LaneType` sounds fine to me. I still would like to avoid the full generality of `osm2lanes` though.. https://github.com/a-b-street/osm2lanes/blob/main/osm2lanes/src/road/lane.rs can express nonsensical things like a parking lane for `Foot`.

Rephrasing... if we start to refactor `LaneType` dramatically, let's please do it in a separate PR and a bit slowly, so I can keep downstream A/B Street code abreast of changes. #91 is an example that would probably be great to do in this repo, but I'd like time to support it everywhere

osm2streets/src/lanes/placement.rs
dabreegster: The separate parsing is awesome, and the unit tests show the niceness. I would vote to just change the API to `impl Placement { fn parse(tags: &Tags) -> anyhow::Result<Self> }` though. I'm not sure how the wider Rust community feels or what's standard in codebases, but the from/into traits don't show up as clearly in rustdocs and they're visually less clear to spot in code. (Something like `let x: Placement = some_complicated_thing().try_into()?` vs `let x = Placement::parse(some_complicated_thing())?`)

dabreegster:
> I've left out propagation of actual errors to keep things as clear as possible for now, and I don't know what error reporting approach we want to build towards.

We don't have a consistent story here yet. Feel free to sprinkle in `warn!` logging statements if it's useful.

Now that we've got an OSM lane editor actually wired up, I'm starting to think about this. I feel that osm2lanes went a bit overboard in creating types to precisely capture all possible things that can go wrong, but the end result is hard to use in the code, the final error messages can wind up nonsense (the `tags.subset` thing often doesn't actually plumb through the context needed to understand an error), and the immediate way we'd want to use these is probably just as a string message anyway. For something like `get_lane_specs_ltr`, I feel like just plumbing back a list of string warnings could be useful to display. In the broader osm2streets, maybe we could have a `streets.record_warning(...)` API build up a list of string errors and maybe display somewhere.

osm2streets/src/lanes/mod.rs
BudgieInWA: Pretty much that (but starting at 1). Contraflow lanes are not mentioned anywhere, but a strict reading of "the forward lanes in LtR order" can be followed by mappers and consumers equally well. Does `placement:backward=left_of:3` look like ` V|^^VV`?

Like tagging `turn:lanes[:forward]=left|left;through|through`, three lanes, 1, 2, 3. 

This is where i need to summarise the idea in comments! And link to the wiki quotes.

osm2streets/src/lanes/mod.rs
BudgieInWA: > separate PRs and a bit slowly.

Absolutely, I'm enjoying the standard set by supporting A/B Street at each iteration.

> I still would like to avoid the full generality of osm2lanes though.

This is an opportunity to see what a useful osm2lanes API would look like, so we can carve osm2lanes down into the support for this lib. `RoadPosition`/`Placement` is an example of the kind of complexity we end up needing (i think). 

I think it's ok that the types (aka vocabulary) let you say semantically nonsense things like "pedestrian parking". What if someone tagged it as park of crowd/queue management at some arena or something? We shouldn't crash. Parsing tags into fields seems simpler than trying to classify into an enum of distinct types right at the beginning. If parking can be tagged with access restrictions, then it's up to the renderer to scratch its head when it tries to find lane markings for foot parking.

BudgieInWA:
> I'm hoping to calculating trimmed geometry early in #136, but likely this PR lands first

Yep, bit by bit!

> It'd probably be better to work off the trimmed and transformed `center_line` in transformations generally. 

That's the idea. We just need to figure out the bare minimum sensible behaviour for existing uses, probably sprinkle some `update_center_line` calls around.

Also, new screenshots in the opening comment!

osm2streets/src/lanes/placement.rs
dabreegster: Yep, `anyhow` or any other error type lets us use `?`. `anyhow` is encouraged for applications and discouraged for libraries, but similar to the thread below about error/warning handling, I don't think increased code complexity is worth it here yet. We can be lazy and use string errors

osm2streets/src/road.rs
dabreegster: `untrimmed_road_geometry` is one of these weird bits of legacy code that I don't remember the original reason for. It just looks for cases where a road has a sidewalk on exactly one side, and shifts over the center line a bit. Most of the uses are in transformations just to check length, and those places could use `reference_line.length()` today instead. And once we maintain `center_line` at all times, most/all probably should use that.

The one "real" use of this is in the very beginning of `transform/intersection_geometry.rs`. We initialize a road's `center_line` to this. I _think_ we want to ditch this method entirely and just use the corrected center that's now calculated from placement.

osm2streets/src/road.rs
```diff
@@ -109,7 +119,37 @@ impl Road {
...
+    /// Calculates and sets the center_line from reference_line, reference_line_placement
+    /// (and TODO trim_start, trim_end).
+    pub fn update_center_line(&mut self, driving_side: DrivingSide) {
+        let ref_position = match self.reference_line_placement {
+            Placement::Consistent(p) => p,
+            Placement::Varying(p, _) => {
+                warn!("varying placement not yet supported, using placement:start");
+                p
+            }
+            Placement::Transition => {
```
dabreegster, line 137: Finally reading the Wiki...

> Instead of the value transition, one can use the [suffixes :start and :end](https://wiki.openstreetmap.org/wiki/Proposed_features/Suffix_start_and_end) to provide information about the placement at the start resp. end of the OSM.way. This is usually not necessary if such information can be retrieved from the previous resp. following OSM way.

To later support this correctly, we have to look at the previous and next OSM ways (and decide which one is "the same road" -- probably just by name?) and interpret explicit or implicit placement tags there?

dabreegster (commented):
I'm starting to look at some of the visual diffs. https://wiki.openstreetmap.org/wiki/Proposed_features/placement#Exemplary_tags:_one-way_road has some nice synthetic examples. I wonder if we could/should use Overpass and look for clear places demonstrating some of these? Or does `fremantle_placement` test things adequately do you think? It looks like it only uses `right_of` (with one `transition`)

osm2streets/src/transform/intersection_geometry.rs
```diff
@@ -4,12 +4,6 @@ use geom::{Circle, Distance};
 use crate::{IntersectionControl, IntersectionKind, StreetNetwork};
 
 pub fn generate(streets: &mut StreetNetwork, timer: &mut Timer) {
-    // Initialize trimmed_center_line to the corrected center
```
dabreegster, line 7: I'm still puzzling through how it happens, but take a look at the diff in `aurora_sausage_link`. It looks like we lose clipping to the map edge. In `Road::new` we're immediately doing `update_center_line`, so this should be set...
OHHH no it's simple. At the end of `streets_reader/src/clip.rs`, we need to call `update_center_line`. #127 would make this simpler possibly

osm2streets/src/road.rs
BudgieInWA: We look at the traffic flow lane by lane to determine the lane connections. With `Connection`s and `Forks` you can examine movements, do lane math and divvy up the lanes easily. Make some assumptions if its not trivial, as the mapper can tag the lane connectivity explicitly too.

I think we are supposed to tag `placement=transition` on the nonsense segments of motorway exits (the ones that go way outside the lanes that they represent, and so couldn't be tagged with any sensible `placement:start` anyway). We probably just assume them to be at the start of the applicable roads automatically, because we need to do it all over the place anyway.

When both Roads have their placement tagged (as non-transition), that actually tells us the lane connectivity at that node. Try mapping a segment where a right turn lane appears for a short distance and the ways stay straight with `right_of:1` for the 2 lane and 3 lane segments.

osm2streets/src/road.rs
BudgieInWA: I think I understand what is going on with all of these. It looks like the extra logic here is trying to pretend the line is placed at `RoadPosition::Center`, so I have implemented it to behave like that. I have touched all the uses of this fn (to add the `driving_side` param), but mostly left them as is. Ideally we replace the uses with more appropriate lookups as we go.

osm2streets/src/transform/intersection_geometry.rs
BudgieInWA: The approach in this PR is to treat `center_line` as derived state and update it as soon as its invalidated (like `kind`). 4b03cdc actually follows through on that and adds the required `update_center_line` calls (after the initial one in `Road::new`), which fixes the `clip` problem, and the one in my original screenshot.

BudgieInWA:
I'm ready for this to merge and to fix things in follow ups, if that works for you within your geom refactor workflow.

I have looked through most of the tests and things look good. Some intersections have different shapes, which I guess comes from a different `untrimmed_road_geometry` and `get_untrimmed_sides` implementation. I think the new ones are more accurate (unless they are buggy, I haven't tested them very thoroughly). 

Running StreetExplorer with new code but letting it load old test geom files from the dir gives a good way to diff the output: The "original test geometry" layer can be turned back on after generating lane markings, to see where the shapes have changed! So I haven't commited any tests yet. Feel free to do that if you are ready to merge.

One big question is the `reference_line`/`center_line` source of truth divide, which is an open question for now, which suggests merging is appropriate.

Another piece that we want soon, but probably after this PR merges, is to improve the implementation of `PolyLine::shift` so that a different start and end offset can be provided. The amount to shift each point should simply be a linear interpolation between `start_offset` and `end_offset` (based on distance along the line), with the remaining logic remaining as is, I think.

osm2streets/src/lanes/placement.rs
```diff
@@ -23,66 +22,60 @@ impl TryFrom<&str> for RoadPosition {
                             "left_of" => Ok(LeftOf(LtrLaneNum::Forward(lane))),
                             "middle_of" => Ok(MiddleOf(LtrLaneNum::Forward(lane))),
                             "right_of" => Ok(RightOf(LtrLaneNum::Forward(lane))),
-                            _ => Err(()),
+                            _ => Err(anyhow!("unknown lane position specifier: {kind}")),
```
dabreegster, line 25: You may also be able to use the `bail!` macro here, which does `return Err(anyhow!(...))`

dabreegster:
Awesome, this is a huge step forwards! I will:

1) Use the `bail!` macro in the parse routines if it helps readability
2) Fix merge conflicts
3) Run the tests and commit all the results if they look good
4) Merge

in the next hour or two, if that's the plan.

> Running StreetExplorer with new code but letting it load old test geom files from the dir gives a good way to diff the output

Oh yeah, I completely forgot we had that! Much easier than switching between browser tabs

> improve the implementation of PolyLine::shift so that a different start and end offset can be provided

Ooh, this sounds exciting! Like everything else, the implementation there is on the scary-legacy-code side. We can definitely have a go at cleaning it up. And https://github.com/georust/geo/pull/935 is in-progress to implement this logic in georust in a much more rigorous and well-tested way.

dabreegster:
Test changes are mixed. Perth stretched lights and peanut get much better; almost-parallel roads that were previously overlapping are now set apart a bit. There are some cases where the intersection geometry no longer matches up with the road, and I'm not entirely sure why yet... borough, bristol sausage link, and service road loop. Some examples:
Screenshot from 2022-12-04 11-55-55
Screenshot from 2022-12-04 11-55-50
Screenshot from 2022-12-04 11-54-48
Screenshot from 2022-12-04 11-54-40

And a few intersections get much worse, possibly due to the problem above. Northgate, quad_intersection, and the 2 Tempe tests:
Screenshot from 2022-12-04 12-03-19
Screenshot from 2022-12-04 12-03-15
Screenshot from 2022-12-04 11-58-44
Screenshot from 2022-12-04 11-58-40

I have a suspicion we're initializing from the wrong reference line when calculating geometry that I'll check into briefly, but I'm happy merging this and iterating

osm2streets/src/transform/separate_cycletracks.rs
```diff
@@ -83,7 +83,7 @@ fn find_cycleways(streets: &StreetNetwork) -> Vec<Cycleway> {
                 {
                     cycleways.push(Cycleway {
                         cycleway: cycleway_road.id,
-                        cycleway_center: cycleway_road.untrimmed_road_geometry().0,
+                        cycleway_center: cycleway_road.center_line.clone(),
```
dabreegster, line 86: This is fine here, and is the kind of thing we'll want to do almost everywhere. `untrimmed_road_geometry` should almost never be used once we maintain `center_line` as we go

PR #140, opened 2022-12-02
Start refactoring intersection geometry code

dabreegster:
This is a small start to refactoring. Instead of using this legacy `InputRoad` struct, just take a **copy** of the `Road` struct as input. All information needed for the algorithm is in there, and so when we add things like trim distances, it means we don't have to copy data back and forth between two structs. It also internally simplifies some code to populate the `Results` struct immediately.

There are a bunch of next steps, but I'll go discuss them in the issue separately. I think this is a solid first step with no diff in the tests. All of the next steps... might change some existing behavior!

### DIFF
--- a/osm2streets/src/road.rs
+++ b/osm2streets/src/road.rs
@@ -270,3 +270,48 @@ impl StreetNetwork {
         id
     }
 }
+
+/// The edge of a road, pointed into some intersection
+#[derive(Clone)]
+pub(crate) struct RoadEdge {
+    pub road: RoadID,
+    /// Pointed into the intersection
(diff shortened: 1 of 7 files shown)


osm2streets/src/road.rs
```diff
@@ -278,3 +270,48 @@ impl StreetNetwork {
         id
     }
 }
+
+/// The edge of a road, pointed into some intersection
```
dabreegster, line 274: I just moved this over from the sidewalk corner rendering code. In most of my next-step branches, I want to use this. We should keep adjusting how this works to get closer to the half-road view idea.

PR #141, opened 2022-12-06
Clicking off a road resets the edit view

matthieu-foucault:
As mentioned in https://github.com/a-b-street/osm2streets/issues/132#issuecomment-1339162927

After selecting a road, there were two options regarding handling clicks on the red overlay (I went with the first one):
- make it non-interactive, meaning that pointer events will go through it. The downside of this is that if you click on the padding around the road, it will be de-selected, which might be confusing.
- keep it interactive and prevent click events from being propagated. This would fix the issue above, but would make it more difficult to refine the selection when there are paths that are parallel to each other. e.g. in the example below, to select the sidewalk north of the road, you would have to de-select the road first:
image

### DIFF
--- a/street-explorer/www/js/lane_editor.js
+++ b/street-explorer/www/js/lane_editor.js
@@ -13,6 +13,11 @@ await init();
 export class LaneEditor {
   constructor(mapContainer) {
     this.map = setupLeafletMap(mapContainer);
+    this.map.on({
+      click: () => {
+        this.resetEditView();
+      },
+    });
     this.network = null;
     this.layers = [];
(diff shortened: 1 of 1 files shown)


dabreegster (approved):
Fantastic, thank you! I agree with the first option. I don't find it surprising that the red highlight isn't part of the hitbox. Any ideas how to better indicate the current selected way? The iD editor uses a similar effect, but with some animated glowing. That may be overkill.

matthieu-foucault:
> Any ideas how to better indicate the current selected way?

Nothing that comes to mind,

PR #144, opened 2022-12-07
Reset the center line to untrimmed before calling intersection_polygo…

### LINKED ISSUE #143: `GenerateIntersectionGeometry` makes existing intersections larger
Not sure if this is a known issue and/or belongs in #136 .

Using the lane editor, intersections get wider every time they are recalculated, quickly going from this:
image

to this:

image

If this is not something that's being worked on already, I'm happy to take a crack at it, with a bit of guidance.

dabreegster:
…n. Will revisit this at some point. #136, closes #143.


dabreegster:
The test diffs look fine.

Longer-term, we kind of have two approaches to decide between. Both should be compatible with the goal of always having a trimmed `center_line` and `intersection.polygon` filled out and not just calculating it last-minute in the transformation.

1) Keep assuming the input to the intersection geometry routines is untrimmed. I think we'll always have enough state around to manage that. The only thing to maybe figure out is that we never lengthen a center line that was shorter in a previous pass. That was partly the motivation for storing `trim_start` and `trim_end` explicitly.

2) Make the routines handle input that's already trimmed, and consider trimming further if something has changed. This feels harder to reason about to me, personally.

PR #145, opened 2022-12-07
Special-case degenerate intersections, and clean up old hacks

dabreegster:
Intersections with exactly 2 roads can be treated separately from the generalized corner-finding algorithm. I chose to make the trim-back length there smaller, but this is arbitrary and we can fiddle around later.

More importantly, I got rid of the mandatory trim back on all roads that happens in the general case. The result is awesome -- intersections are much more minimal suddenly. It even fixes some of the cases broken by placements, like bristol sausage (the before/after order might be off; the better-looking thing is the new one!):
Screenshot from 2022-12-07 13-15-36
Screenshot from 2022-12-07 13-15-32

Also service road loop:
Screenshot from 2022-12-07 13-20-32
Screenshot from 2022-12-07 13-20-28

For fun, the montlake roundabout:
Screenshot from 2022-12-07 13-17-50
Screenshot from 2022-12-07 13-17-45

When we have clusters of short roads in Perth, things look better -- the short roads get a tiny bit longer, at least making it easier to understand what's happening and figure out more steps later:
Screenshot from 2022-12-07 13-18-53
Screenshot from 2022-12-07 13-18-49

Sometimes the more minimal trim produces funny shapes, like at `Connections`, but that's fine. When we figure out how to zip the dual carriageways together or ignore the fake geometry at the start/end, this'll be less of a problem.
Screenshot from 2022-12-07 13-13-42
Screenshot from 2022-12-07 13-13-31

CC @matthieu-foucault as a jumping-off point into the intersection geometry logic. In case you're interested how this works, https://a-b-street.github.io/docs/tech/map/geometry/index.html is a somewhat dated overview. It talks about the "general" case, which is in `geometry/algorithm.rs`. Turns out it's helpful to peel off some of the special cases before trying to work on this general one more.


osm2streets/src/intersection.rs
```diff
@@ -18,6 +18,8 @@ pub struct Intersection {
     /// StreetNetwork; roads and intersections get merged and deleted.
     pub point: Pt2D,
     /// This will be a placeholder until `Transformation::GenerateIntersectionGeometry` runs.
+    ///
+    /// TODO Consistently make this clockwise.
```
dabreegster, line 22: Things like the sidewalk corner are... maybe?... sensitive to the polygon orientation. Anyway, once the dust settles, establishing this invariant would probably be helpful to downstream logic.

dabreegster (commented):
Going to merge this and keep going, but feedback welcome. There are arbitrary choices for degenerate intersections that I'm not opinionated about at all and happy to change. I just wanted to get rid of the always-trim behavior in the general case

PR #146, opened 2022-12-07
Rewrite the general intersection polygon algorithm. #136

dabreegster:
Finally, rewrite the main algorithm. The basic idea hasn't changed, just the implementation. Previously, we were intersecting every pair of road lines, because once upon a time roads weren't sorted clockwise. Now that they are, we just have to consider adjacent pairs.

Most of the diffs are actually invisible. A few places like `kingsway_junction` change, but both before and after look silly, so not noteworthy. There are **regressions** in cycleway_rejoin_road, leeds_cycleway, roosevelt_cycletrack due to roads sharing the same endpoints. I fiddled around a bit to try to fix this, but I'm going to follow up separately.

The new algorithm is just 100 lines of code. That's the step forward here! :)


osm2streets/src/geometry/algorithm.rs
```diff
@@ -80,224 +76,135 @@ pub fn intersection_polygon(
...
+    for pair in edges.windows(2) {
+        let one = &pair[0];
+        let two = &pair[1];
 
-        for (r2, pl2) in &road_lines {
-            if r1 == r2 {
-                continue;
-            }
-
-            // If two roads go between the same intersections, they'll likely hit at the wrong
```
dabreegster, line 144: The regression is because I'm tossing away this code. I don't understand the problem clearly yet. Will fix up in a followup, along with regressions from the placements work in tempe and northgate tests

osm2streets/src/geometry/algorithm.rs
```diff
@@ -80,224 +76,135 @@ pub fn intersection_polygon(
...
-            }
+        // Look for where the two road edges collide, closest to the intersection.
+        if let Some((pt, _)) = one.pl.reversed().intersection(&two.pl.reversed()) {
+            // For both edges, project perpendicularly back to the original center, and trim back
+            // to that point.
+            for side in [one, two] {
+                if let Some((_, angle)) = side.pl.dist_along_of_point(pt) {
+                    let perp = InfiniteLine::from_pt_angle(pt, angle.rotate_degs(90.0));
+
+                    // Make the road center point away from the intersection, so we find the hit
```
dabreegster, line 130: Lots of the complexity comes from this. Most of the time polylines just hit at one point and it doesn't matter. But we have to spend lots of effort here to find the hitpoint closest to the intersection we're operating on. There's probably an easier way to do this.

osm2streets/src/geometry/algorithm.rs
```diff
@@ -80,224 +76,135 @@ pub fn intersection_polygon(
...
-            shortest_center.reversed()
-        };
-        if let Some(existing) = new_road_centers.get(r1) {
-            if new_center.length() < existing.length() {
-                new_road_centers.insert(*r1, new_center);
             }
-        } else {
-            new_road_centers.insert(*r1, new_center);
         }
+        // TODO If there's no hit, consider extending both lines and seeing if they hit
```
dabreegster, line 160: seattle_slip_lane has a case where I think this would help, will try later

osm2streets/src/geometry/algorithm.rs
```diff
@@ -80,224 +76,135 @@ pub fn intersection_polygon(
...
-                 stuff's too short"
-            );
         }
     }
+    endpts.push(endpts[0]);
+    if let Ok(ring) = Ring::deduping_new(endpts) {
+        results.intersection_polygon = ring.into_polygon();
+    }
 
-    // There are bad polygons caused by weird short roads. As a temporary workaround, detect cases
```
dabreegster, line 270: So long, absolutely ridiculous crazy code. I won't miss you.

PR #147, opened 2022-12-07
Selection overlay indicates road direction

matthieu-foucault:
To indicate the road direction and facilitate tagging, I've been toying with the idea of indicating the direction with the selection overlay itself, instead of adding another arrow, which would add clutter and may be tricky to place with curved roads, e.g.

image

This is made with `make_arrow`, but it's a bit janky when there's a curve at the end of the road or with some loops.

image

image

image

image

(looks like the intersection detection is also having issues with this last one)

I'm thinking of making a different arrow style:

image
This style would allow to see the direction of the road by looking at either end, but it might make some loops look even more confusing.

image
This might be better if we keep the chevrons from being created at the ends of lines. This might not work well with small cul-de-sac though (e.g. my second screenshot)

Any other ideas? Let me know what you think and if I'm looking at this the wrong way.

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -150,8 +150,8 @@ impl JsStreetNetwork {
             .unwrap();
 
         // Show a wide buffer around the way
-        let polygon =
-            PolyLine::unchecked_new(self.ways[&id].pts.clone()).make_polygons(1.5 * width);
+        let polygon = PolyLine::unchecked_new(self.ways[&id].pts.clone())
+            .make_arrow(1.5 * width, geom::ArrowCap::Triangle);
 
         abstutil::to_json(&polygon.to_geojson(Some(&self.inner.gps_bounds)))
     }
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 
 use abstutil::{Tags, Timer};
-use geom::PolyLine;
+use geom::{Distance, Line, PolyLine, Polygon};
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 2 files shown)


dabreegster:
The arrow looks OK. For loop cases, if the arrow was placed a little bit back from the end, it might help? You could use something like `polyline.exact_slice(Distance::ZERO, 0.8 * polyline.length()).make_arrow` for that, and then maybe fill in the rest of the polyline.

> This style would allow to see the direction of the road by looking at either end

It'd still be just one arrow cap at the end of the polyline? Not sure this would look that much different than the triangular cap.

> This might be better if we keep the chevrons from being created at the ends of lines.

Some kind of regularly repeated chevrons along most of the polyline (maybe 10-90% of it, to avoid the ends) would look best, IMHO. It works well enough in iD:
Screenshot from 2022-12-08 10-59-21

If you go this route, it could be more convenient to write the code in this repo. Later we can move it to `geom` in the abstreet repo. Or you can clone the other repo and use https://github.com/a-b-street/osm2streets/blob/4ddab02c2218a6fc56846c5aa13a04a015fd3125/Cargo.toml#L11

matthieu-foucault:
Still tweaking the chevrons placement and frequency, but this is what I have now. 
I added an opaque 1px stroke to the overlay, otherwise the chevrons are hard to see.

image

dabreegster:
This is looking pretty good to me! I like the approach of subtracting from the thickened polyline, instead of drawing something that might block lane markings. I'm also wondering if we should turn down the opacity of the red a bit, maybe to 30%, to not cover up the lanes as much?

matthieu-foucault:
Alright, I think I have something that's ready to merge: 
- chevrons are uniformly spread every 50m along the polyline, with a minimum of one chevron.
- overlay opacity is reduced to 0.3

A couple of examples: 

image
image
image
image

dabreegster (approved):
Awesome, the result is legible, and it's easier to see lane markings now. Thanks for the PR!

PR #148, opened 2022-12-08
Fix geometry for one case where two roads go between the same pair of…

dabreegster:
… intersections. #136

Screenshot from 2022-12-08 12-02-29
Screenshot from 2022-12-08 11-49-50

In this one case, the problem was the hit point between the two roads being detected at the wrong end:
Screenshot from 2022-12-08 11-39-09
As the comment says, this is a bug in polyline. And apparently it's not the first weird one: https://github.com/a-b-street/abstreet/blob/aa40be4e82acd390d3532eb13e61fdce0d0a0fd8/geom/src/polyline.rs#L790
I'm going to see if I can rip out this scary old line intersection code I wrote myself years ago and switch to georust now. :)


dabreegster:
Also, loop roads in leeds_cycleway and roosevelt_cycletrack are still broken for some other reason, I'll look at those too

PR #149, opened 2022-12-08
Improve intersections in some cases by extending untrimmed road edges…

dabreegster:
… a bit. #136

I'm not sure we should do this. But just to discuss the idea...

seattle_slip_lanes improves:
Screenshot from 2022-12-08 14-58-43
Screenshot from 2022-12-08 14-58-49
Though this is a weird case, because got overlapping intersections, or very nearly.

aurora_sausage_link reveals an interesting problem.
Screenshot from 2022-12-08 14-59-51
Screenshot from 2022-12-08 15-00-01
It's OK before, but we create less minimal intersections, and then the road doesn't meet the intersection cleanly. Because the trimmed back center is right on a slight curve, when we project left and right, the miter/bevel/whatever join acts a bit differently.

Is this an idea worth pursuing and refining? I'm more tempted to focus on something like seattle_slip_lanes where it gets us a bit closer to something good, but keep focusing on this one test case and figure out **everything** needed to make it look better.

### DIFF
--- a/osm2streets/src/geometry/general_case.rs
+++ b/osm2streets/src/geometry/general_case.rs
@@ -36,13 +36,27 @@ pub fn trim_to_corners(
         }
 
         // Look for where the two road edges collide, closest to the intersection.
-        if let Some((mut pt, _)) = one.pl.reversed().intersection(&two.pl.reversed()) {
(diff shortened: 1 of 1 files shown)


dabreegster:
Don't think this is as relevant an idea now, closing

PR #151, opened 2022-12-12
Explicitly use clipping boundaries in StreetExplorer and tests

dabreegster:
Before we can refactor how `streets_reader` works or make progress on intersection geometry refactor, we need to understand the current clipping behavior clearly. Step one of that is to actually see the boundary! And for all test cases, we need to specify the clip.

You can now drag custom boundaries in StreetExplorer and import them. I manually drew a boundary for every test case and regenerated everything. I made some best judgment about what we're trying to capture in each test case, but was pretty faithful to the originals. We can now clearly see how the generated output fits the boundary:
Screenshot from 2022-12-12 21-27-37
Screenshot from 2022-12-12 21-27-09

Next up, I'll overhaul clipping as described in #127.


dabreegster:
When the roads blatantly go past the boundary, that's usually https://github.com/a-b-street/osm2streets/blob/e1e01d2e285f3fe72737702292245f446d79854f/osm2streets/src/geometry/terminus.rs#L22. This is what I brought up in https://github.com/a-b-street/osm2streets/issues/136#issuecomment-1344860378. Either here or there -- what should we do for extremely short roads by the boundary? As shown there, when we don't generate geometry for those very short roads, it can mess up. Extending past the boundary helps, but is weird and complicates the geometry refactor

dabreegster:
Oh yeah, the other caveat: All of the test case osm.xml files we've saved comes from the old world, where we asked Overpass for a huge bounding box. When we make new test cases from here on, we'll request a smaller amount of OSM data in the first place. Just something to keep in mind.

dabreegster:
Going to merge this to reduce rebasing with some other experiments, but happy to address feedback later

PR #153, opened 2022-12-15
Always pass untrimmed input into the intersection geometry routines. …

dabreegster:
…Start recording trim_start and trim_end and use that to calculate the final trimmed center line. #136

Alright, I finally made progress. This now
1) always consistently passes in untrimmed centers to the geometry routine
2) figures out the trim start/end afterwards
3) detects cases where roads get trimmed to oblivion

The test diffs all look expected to me. There are a bunch of spurious changes to map edges, terminii, and degenerate intersections, because now they see the full untrimmed length of the road and make a decision about failing or extending something based on that. (Instead of getting lucky with the ordering, like they were previously.) Lots of the behavior here is still wrong, but we can fix it up separately.

There are visible diffs in every case where we trim a road into oblivion, because now I'm leaving that broken road totally untrimmed in the final output. One good example before:
Screenshot from 2022-12-15 12-15-09
This is now:
Screenshot from 2022-12-15 12-15-16
Which looks visibly worse. I've confirmed the console logs show it's detecting this road as broken. If you hide lane polygons, you see some decent intersection shapes peeking through. If we merge some of these roads, maybe the result will start to look sane.
Screenshot from 2022-12-15 12-15-25

I'm not attempting to auto-merge the bad cases yet, or to get rid of the transformation step. Will try those next, and not sure about the ordering between the two changes.


osm2streets/src/geometry/mod.rs
```diff
@@ -82,6 +83,8 @@ impl InputRoad {
 pub struct Results {
     pub intersection_id: IntersectionID,
     pub intersection_polygon: Polygon,
+    /// The only transformation to `center_line` passed in must be to trim it (reducing the length)
```
dabreegster, line 86: I think you were arguing for this interface to return the trim distances directly. We could clean that up as another step too

osm2streets/src/road.rs
```diff
@@ -443,6 +455,45 @@ impl Road {
     pub fn common_endpoint(&self, other: &Road) -> CommonEndpoint {
         CommonEndpoint::new((self.src_i, self.dst_i), (other.src_i, other.dst_i))
     }
+
+    /// This trims a polyline on both ends. Positive trim distances mean to shorten the polyline,
+    /// and negative mean to extend the polyline in a straight line by some amount. If this returns
+    /// `None`, then the two trim distances are incompatible -- the entire polyline disappears.
+    ///
+    /// TODO Maybe move this to PolyLine directly. This is a utility method currently for A/B
```
dabreegster, line 463: I'm going to try using this in the A/B Street editor and operate off the same concept of two trim distances

osm2streets/src/road.rs
```diff
@@ -443,6 +455,45 @@ impl Road {
...
+        mut pl: PolyLine,
+        trim_start: Distance,
+        trim_end: Distance,
+    ) -> Option<PolyLine> {
+        // The two ends trimmed past each other
+        if trim_start + trim_end > pl.length() {
+            return None;
+        }
+
+        // Note we use maybe_exact_slice and bail out upon failure of the resulting line being too
```
dabreegster, line 475: Aka, this is the epsilon check part of the above. `pl.length() - trim_start - trim_end` is nonzero but too small for geom layer

osm2streets/src/transform/intersection_geometry.rs
```diff
@@ -1,36 +1,41 @@
 use abstutil::Timer;
 use geom::{Circle, Distance};
 
-use crate::{IntersectionControl, StreetNetwork};
+use crate::{Road, StreetNetwork};
 
 pub fn generate(streets: &mut StreetNetwork, timer: &mut Timer) {
-    // TODO intersection_polygon assumes untrimmed lines as input, so reset here. Once we always
```
dabreegster, line 7: This is now redundant, because `to_input_road` doesn't use it, and we set `road.center_line` at the end of this file

osm2streets/src/transform/intersection_geometry.rs
```diff
@@ -49,21 +54,29 @@ pub fn generate(streets: &mut StreetNetwork, timer: &mut Timer) {
                         road.center_line.last_pt()
                     };
                     set_polygons.push((i.id, Circle::new(pt, Distance::meters(3.0)).to_polygon()));
-
-                    // Also don't attempt to make Movements later!
```
dabreegster, line 53: This is some weird old A/B Street assumption we can get rid of now. When intersection geometry broke at a traffic signal, my old code for figuring out signal timing broke. I'll workaround that elsewhere.

osm2streets/src/transform/intersection_geometry.rs
```diff
@@ -49,21 +54,29 @@ pub fn generate(streets: &mut StreetNetwork, timer: &mut Timer) {
...
+
+    for road in streets.roads.values_mut() {
+        let untrimmed = road.get_untrimmed_center_line(streets.config.driving_side);
+        if let Some(pl) =
+            Road::trim_polyline_both_ends(untrimmed.clone(), road.trim_start, road.trim_end)
+        {
+            road.center_line = pl;
+        } else {
+            // TODO Mark for collapsing?
+            error!("{} got trimmed into oblivion", road.id);
```
dabreegster, line 72: This shows up consistently where I expect it to / where the road looks visibly untrimmed. Maybe it's as simple as setting `internal_junction_road` and repeating a few transformation passes. That's a little unsatisfying in the long run... I wouldn't want to wind up regenerating geometry in the whole network after fixing a few things.

osm2streets/src/geometry/mod.rs
dabreegster: ... and I've made this change now, to save duplicating logic in A/B Street. :)

dabreegster:
Going to merge to avoid rebases, but eager for feedback

PR #154, opened 2022-12-15
Overhaul clipping

dabreegster:
As discussed in #127, clipping should be implemented at the OSM document level, not for `StreetNetwork`. This PR makes this change. The result definitely feels simpler.

3 changes as a result. The first are spurious, slight jitters of existing geometry due to using RDP simplification at a different point, or doing some geometry operations with floating point sensitivity in a slightly different order.

Second, I got rid of the tiny roundabout collapsing logic entirely. We could consider restoring it later, but it currently just papers over a real problem we should deal with in another way. The original motivation was to help the A/B Street traffic sim logic, which doesn't like vehicles on tiny roads, but oh well. Let's actually figure out how to deal with these:
Screenshot from 2022-12-15 16-26-49
Screenshot from 2022-12-15 16-26-40

Finally, we're now much more precise with clipping roads that dip in and out of a boundary. An example:
Screenshot from 2022-12-15 16-22-37
Screenshot from 2022-12-15 16-22-30


streets_reader/src/lib.rs
```diff
@@ -84,6 +86,14 @@ fn extract_osm(
         timer.next();
         out.handle_way(*id, way, &streets.config);
     }
+    timer.start_iter(
+        "processing OSM ways split into pieces",
+        doc.clipped_copied_ways.len(),
+    );
+    for (id, way) in &doc.clipped_copied_ways {
```
dabreegster, line 93: What's up with these? The problem is this:
Screenshot from 2022-12-15 16-10-56
We've got a way that crosses the boundary twice. In an earlier version of this PR, we got this incorrect result. Now we split this one way into 3 pieces, discard the middle one (out of bounds). And thankfully multiple ways per OSM ID don't cause any headache elsewhere, since we're already using things like opaque IDs.

streets_reader/src/osm_reader/clip.rs
```diff
@@ -0,0 +1,89 @@
+use geom::{Distance, PolyLine, Polygon};
+
+use super::Document;
+
+impl Document {
+    // TODO This destroys the guarantee that the Document represents raw OSM. Do we need to be
```
dabreegster, line 6: Depending what else we do with `Document`, we have to be careful here. `way.pts` and `way.nodes` get out of sync. Nothing downstream uses `way.nodes` right now, but if that ever changes...

streets_reader/src/osm_reader/clip.rs
```diff
@@ -0,0 +1,89 @@
...
+                self.clipped_copied_ways.push((*id, copy));
+            }
+        }
+
+        // Remove the "original" from ways that were split into pieces
+        for (id, _) in &self.clipped_copied_ways {
+            self.ways.remove(id);
+        }
+
+        // TODO Handle ways that're areas
```
dabreegster, line 51: Longer-term, `osm_reader` could be split into its own crate. It's independent of the rest of osm2streets logically. But I don't want to go through all the headache of figuring out how to clip areas and relations yet, since we don't have test cases here set up to look at those

streets_reader/src/osm_reader/multipolygon.rs
```diff
@@ -7,22 +7,73 @@ use osm2streets::osm::{OsmID, RelationID, WayID};
 
 use super::{Document, Relation};
 
-pub fn get_multipolygon_members(
```
dabreegster, line 10: No real changes here, just organizing the methods onto Document

streets_reader/src/split_ways.rs
```diff
@@ -121,37 +88,29 @@ pub fn split_up_roads(
                     }
                 }
 
+                let i1_node_id = input.osm_node_ids.get(&pts[0].to_hashable()).cloned();
+                let i2_node_id = input.osm_node_ids.get(&pt.to_hashable()).cloned();
+
                 let untrimmed_center_line = simplify_linestring(std::mem::take(&mut pts));
                 match PolyLine::new(untrimmed_center_line) {
                     Ok(pl) => {
+                        // TODO If either endpoint is on the boundary, we won't record any OSM ID.
```
dabreegster, line 97: I want to look into this as a followup. OK with getting rid of `OriginalRoad` entirely? A caller can look at a Road's 0 or more WayIDs and for each endpoint, 0 or more NodeIDs. There's lots of possibilities there to link to original OSM thing, and callers need to deal with it

dabreegster:
Cool! I rebased and regenerated tests. Also got rid of one of the roundabout tests; `montlake_roundabout` covers it and has weirder geometry.

streets_reader/src/osm_reader/clip.rs
BudgieInWA: Exactly the idea I was having. Think about it in those terms for good *separation of concerns*, but only implement it as needed here.

PR #155, opened 2022-12-16
Collapse roads trimmed into oblivion

dabreegster:
Alright, now all the refactoring is starting to really pay off. When we trim roads back from intersections and some road disappears entirely, that road is too short to have its own geometry. This PR sends it through the consolidation pass. The results are pretty fantastic.

Screenshot from 2022-12-16 13-34-19
Screenshot from 2022-12-16 13-34-14

Screenshot from 2022-12-16 13-36-06
Screenshot from 2022-12-16 13-36-02

I also had to first fix an edge case with `polygon_from_corners`. The comments / pictures below should explain:
Screenshot from 2022-12-16 13-29-53
Screenshot from 2022-12-16 13-29-48
Screenshot from 2022-12-16 13-28-37

I'm now extremely tempted to remove the dog-leg intersection detection entirely. As we discussed, the motivation for that was basically to solve problems like this, which used to be more unclear. Before we had intersection geometry generated on totally untrimmed input, I didn't even understand the root problem.


dabreegster:
Yeah, I removed the old dog-leg experiment. No effect on tests, because it's disabled in this repo anyway. The ideas in #131 are good to remember -- we should keep going down the minimal-intersection path and fix root problems a different way.

BudgieInWA:
Looking beautiful!
Now we can try to reason about these intersections in general terms; with lane movements enumerated, intersected, and drawn. Some with lane markings, some without. Probably the stop lines too.

I'm wondering if we ought to keep the collapsed roads alive, referenced in a kinda "merged intersection" `struct LogicalIntersection { ids, roads }` thing. It could take over the final geometry algorithm, uses existing trim distances and supporting multiple nodes. The existing trim distance calculation stuff gets to continue ignoring existing trims, and remain in use only for the individual node "intersections". Often the road segments are a real continuation of a road (the exact piece we want to render as the "major" road through the intersection.

dabreegster:
Keeping around the original structures has some uses for sure. Also interpreting turn restrictions -- being able to look at what's possible in the original graph, then apply that to the final one. But I don't know how to concretely start introducing this kind of `LogicalIntersection` structure. Maybe we need to work on something that really motivates it, like rendering lane markings through intersections. Could it be time to try that?

Thinking about the `seattle_slip_lane` example a bit more, really the east-west main road should be the thing determining the geometry here. The separate cyclepath shouldn't influence things so much. Maybe if we zip it, things would help. Or maybe now's the time to try calculating geometry and ignoring non-road pieces in a first pass.

dabreegster:
Maybe a concrete next step is rendering lane markings in connections and forks. We determine the main road that continues uninterrupted, then at the `render.rs` layer only, draw markings using untrimmed geometry instead of trimmed. Then cases like this where we totally delete one of the original pieces of the main road will become obvious and give us a testable thing to check as we introduce `LogicalIntersection`

BudgieInWA:
> Maybe a concrete next step is rendering lane markings in connections and forks. We determine the main road that continues uninterrupted, [and draw it untrimmed].

We also need to invent the geometry to replace the non-sense angled Y shaped reference lines in sausage links. Either extending the single carriageway, or drawing arcs based on turn level movement paths. 

Lane level movements is the key here. What we currently have is more like the "turns" (they match up to the "turn restrictions"), describing each direction. Then you divvy lanes up according to the connectivity tags or convention, and discover what kind of intersection you really have. Perfect lane allocations for a nice fork or converge, or lanes decreasing for a merge-in freeway entry? Each set of lanes has a kind, which determines which one covers the others.

> ignoring non-road pieces in a first pass.

Ignore all non vehicle lanes - like footpaths - makes sense I think. Keep bike routes in, they are part of the vehicle flow. That way we minimum trimming and wont get too confused by footpath logic while we're trying to understand the traffic interactions.

> Then cases like this where we totally delete one of the original pieces of the main road will become obvious and give us a testable thing to check as we introduce `LogicalIntersection`

Sounds good!

PR #156, opened 2022-12-16
Prototype: Draw lane markings through the intersection for connections and forks.

dabreegster:
Just exploring what this might look like. Hardcoded for `aurora_sausage_link`.
Screenshot from 2022-12-16 16-58-24
Actually looking for connections or forks is maybe not so relevant. The first obvious place in my mind to try to do this is at intersections between main and minor roads, where one side has a stop or give-way sign. So, problem 1, we don't have a representation for that yet. Should there be `IntersectionKind` to differentiate some of the roads have priority over the others?

Regardless of the `Kind` question, we need to start representing priority between conflicting roads. We can start looking at [highway=stop](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop) in OSM, and when it's missing, just use `RoadRank`. How do we want to express this? Should it just be calculated on-the-fly? Should `Intersection` store the priority along with its list of ordered roads?

We can have region-specific rendering to show the lower priority roads. The US has stop signs on the side; here's one not-birds-eye-view idea from A/B Street that doesn't necessarily look so great:
Screenshot from 2022-12-16 17-04-08
Give way symbols in the UK will be easier to incorporate in 2D:
Screenshot from 2022-12-16 17-04-36

### DIFF
--- a/osm2streets/src/render.rs
+++ b/osm2streets/src/render.rs
@@ -91,7 +91,7 @@ impl StreetNetwork {
             for (lane, pl) in road
                 .lane_specs_ltr
                 .iter()
-                .zip(road.get_lane_center_lines().into_iter())
+                .zip(road.get_lane_center_lines(self).into_iter())
             {
                 pairs.push((
                     pl.make_polygons(lane.width)
@@ -138,7 +138,7 @@ impl StreetNetwork {
 
         for road in self.roads.values() {
             // Always oriented in the direction of the road
-            let mut lane_centers = road.get_lane_center_lines();
+            let mut lane_centers = road.get_lane_center_lines(self);
 
             for (idx, pair) in road.lane_specs_ltr.windows(2).enumerate() {
                 // Generate a "center line" between lanes of different directions
--- a/osm2streets/src/road.rs
+++ b/osm2streets/src/road.rs
@@ -7,8 +7,8 @@ use geom::{Angle, Distance, PolyLine};
 use crate::lanes::{Placement, RoadPosition};
 use crate::{
     get_lane_specs_ltr, osm, CommonEndpoint, Direction, DrivingSide, InputRoad, IntersectionID,
-    LaneSpec, LaneType, MapConfig, OriginalRoad, RestrictionType, RoadID, RoadWithEndpoints,
-    StreetNetwork,
+    IntersectionKind, LaneSpec, LaneType, MapConfig, OriginalRoad, RestrictionType, RoadID,
+    RoadWithEndpoints, StreetNetwork,
 };
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


dabreegster:
@BudgieInWA, just a gentle ping here. Is priority between roads a concept we need to start adding? Should it fit into `IntersectionKind` or be orthogonal to it?

BudgieInWA:
> Regardless of the `Kind` question, we need to start representing priority between conflicting roads. We can start looking at [highway=stop](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop) in OSM, and when it's missing, just use `RoadRank`. How do we want to express this? Should it just be calculated on-the-fly? Should `Intersection` store the priority along with its list of ordered roads?

Yes, we need to start representing those priorities. I think we need to represent the priority/`TrafficInterruption` for each end of each `Road`, where it meets the intersection. The `Intersection` can then use them to figure out which movements are continuations. I don't know if should be stored in `Intersection` or `Road`. `Intersection`s need just the summarised `TrafficInterruption`, while `Road`s will need more detail to represent where exactly the stop line is and advanced stop lines for bikes. 

I think we should aim for a solution that works when all yields are tagged, and layer the inferring of yields on top somehow. The lack of tagging should not stop us from discovering the simplest way to fit this concept in. Maybe `RoadRank` is a reliable indicator that can be calculated on the fly, or maybe the inference has to be a transformation that looks at the full `Intersection` and marks the yields that it calculates.

> Should there be `IntersectionKind` to differentiate some of the roads have priority over the others?

The road priority is largely orthogonal to `IntersectionKind` because there can be all sorts of combinations of continuations, yields and signals in an intersection so a single classification is not enough to implement these lane marking and stuff like that. I do suspect that it might be useful for `IntersectionKind` to distinguish between (for lack of proper terms) `TrueIntersection` (no continuations) and `SideIntersection` (at least one continuation).

---

I've got more ideas on the other questions here that I'll get to later.

BudgieInWA:
> We've talked about how `to_lane_markings_geojson` can't take a per-road view. Things like dash and arrow spacing need the uninterrupted picture. But let's say version one could use the approach in this PR -- use the untrimmed road centers when appropriate. That gets us a little closer.

One of the base capabilities that we need here is to determine which movements are uninterrupted, and which ones aren't. I imagine this being calculated just after `IntersectionKind` and being stored in the `Movement`s themselves. `Connection`s and `Fork`s are a good place to start, because the continuations should be straightforward. For complex intersections, I guess we would need to enumerate potential continuations (based on the `TrafficInterruption` of the involved `Road` ends) and see if they don't conflict with each other.

`Movement`s also need to represent lane connectivity and have geometry. 

Then I think it is best to generate lane markings for each `Movement` over the extent of the intersection, instead of extending `Road` markings into the intersections. Maybe the extended `Road` approach will be a nifty special case that results in fewer seams, but as discussed, we need to solve the artefacts at seams anyway. One problem with the current approach is that there may be a difference in the angle of the untrimmed segments that meet in the middle of a `Connection`. Because our offset algorithm doesn't yet support non-square ends, the geometry of the road markings won't join up, as demonstrated on the left of the following image.

non-square ends

If we generate geometry for lane markings independently for areas inside `Intersection`s, then we won't have this problem because our `Road`s join at right angles. Having correct geometry would be awesome, even if the dash spacing/offset looks crappy, because the geometry is one of the things that will be needed as part of the global-view lane marking algorithm, and we can start verifying it straight away.

My other motivation for going straight to `Intersection` generated markings, is that I want to support other intersection markings that follow movements too, like these turn-guides:

image

and certain aspects of the continuation lane markings are different through an intersection. For instance, solid markings that are crossed by side-road traffic might be dashed through the intersection, like the bike lanes and center dividing markings here:

different markings through intersections

BudgieInWA:
I think the rendered stop line or give way line itself should be the primary visualisation of these features. It seems to me that many locales have distinct markings for these cases. Though I don't mind the A/B street approach that you show either. I think positioning the respective road sign in the middle of the lane - where the white give way diamond would be in the UK, as per your Give Way example - is a good clear and explicit visualisation for certain contexts.

PR #158, opened 2022-12-26
Introduce consistency checks

dabreegster:
We had talked about internal sanity checks a while ago. While figuring out where orphaned intersections were being introduced, I decided to start these

### DIFF
--- a/osm2streets/src/validate.rs
+++ b/osm2streets/src/validate.rs
@@ -1,0 +1,52 @@
+use crate::StreetNetwork;
+
+impl StreetNetwork {
+    /// Validates various things are true about the StreetNetwork, panicking if not.
+    pub fn check_invariants(&self) {
+        for r in self.roads.values() {
+            for i in r.endpoints() {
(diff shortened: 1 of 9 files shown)


osm2streets/src/validate.rs
```diff
@@ -0,0 +1,52 @@
+use crate::StreetNetwork;
+
+impl StreetNetwork {
```
dabreegster, line 3: Any other ideas for what we should check? Another idea is that `update_center_line`, `update_movements`, and (soon) `update_geometry` are all no-ops, to enforce that we're not forgetting to call something somewhere. (And I may open an issue about guaranteeing that in a more principled way)

BudgieInWA: Not off the top of my head. But i love that there is a clear place to add any checks that come up while debugging.

Checking for no-op `update_` calls sounds very useful for enforcing those implicit dependencies, especially while trying to optimise where the update calls are made.

PR #161, opened 2022-12-26
Show movements per lane #67

dabreegster:
Just a small step towards richer movements. Visualizing all of them at once is too overwhelming, so just prepare for displaying something per lane. The actual filtering is still per-road right now, of course.

https://user-images.githubusercontent.com/1664407/209577533-9000615c-42ae-435e-a5bb-c00cea63c060.mp4

Is it still useful to show all movements at once, or should we remove that layer?


BudgieInWA (approved):
I like this UX!

You can remove the full movements layer i think. We can easily add something similar back in to debug if needed when making changes to movements.

PR #164, opened 2022-12-30
Switch to Vite

dabreegster:
This stops manually vendoring dependencies and switches to npm+vite. Adding new dependencies later gets much easier. Please make sure the instructions for running locally in the readme work for you and make sense.

I also got rid of the d3 graphviz rendering. As a side effect, loading big tests like `leeds_cycleway` gets noticeably faster -- I guess that was actually a bottleneck!

It's possible I've messed up the github actions, but it's impossible to tell without just deploying it. :(

I left the smooth zoom plugin as a manual dep, because I couldn't get it working. The original repo has no NPM package. I found https://www.npmjs.com/package/projektpro-leaflet-smoothwheelzoom but it doesn't seem to work?


.github/workflows/web.yml
```diff
@@ -13,17 +13,25 @@ jobs:
...
       - name: Build web app
         run: |
           cd street-explorer
           wasm-pack build --release --target web ../osm2streets-js
-          # Make a full copy of what we'll publish, dereferencing symlinks
-          cp -RLv www/ publish_www
-          # wasm-pack generates this file, but it'll prevent the action below from working
-          rm -fv publish_www/js/osm2streets-js/.gitignore
+          npm run build --if-present
+          # Manually copy the tests
```
dabreegster, line 30: We could use npm scripts to do this differently, maybe

street-explorer/README.md
```diff
@@ -2,17 +2,23 @@
...
+`npm install` once.
 
 ```
-street-explorer/> ./serve_locally.sh
+# To rebuild the Rust library and WASM bindings
+npm run wasm
+# To run a local server
+npm run dev
+# To rebuild the Rust library and launch the local server
+npm run wasm && npm run dev
```
dabreegster, line 15: This is probably the common case

street-explorer/css/main.css
```diff
@@ -1,5 +1,3 @@
-
```
dabreegster, line 1: I ran prettier on everything

dabreegster:
And while I'm here, I cleaned up the unused code for handling drag-and-drop files. Now that we need a `.osm` and a `.geojson` boundary as input, not sure we'll ever be able to revive this

BudgieInWA (approved):
Looking good! I'm glad you found an elegant build tool that plays nicely with `wasm-pack`.

The build instructions in the README work for me!

Issue #159, opened 2022-12-26
Propagating changes


dabreegster:
There are a few cases today where it's the caller's responsibility to restore invariants after one operation:

1) After `remove_road`, it's possible an intersection will be left orphaned, with no roads pointing to it. Those must be removed. https://github.com/a-b-street/osm2streets/blob/a9021348abded03cbedbdb77b382aafd52e6a37f/osm2streets/src/transform/collapse_intersections.rs#L123 is one place I just added to manually fix this.
2) `update_movements` has to be called when anything connected to an intersection changes
3) `update_center_line` needs to happen when `reference_line` changes
4) `collapse_short_road` may produce a loop-road that should get removed later. https://github.com/a-b-street/osm2streets/blob/a9021348abded03cbedbdb77b382aafd52e6a37f/osm2streets/src/operations/collapse_short_road.rs#L119

I'm working on a 5th, regenerating intersection geometry + trims.

Relatedly, I forgot where, but there's some branch where I can improve final geometry, but only by rerunning some transformations after generating intersection geometry. That process calculates trims, discovers roads too short to exist, and then we [collapse short roads again](https://github.com/a-b-street/osm2streets/blob/a9021348abded03cbedbdb77b382aafd52e6a37f/osm2streets/src/transform/mod.rs#L43). There's a case where if we `CollapseDegenerateIntersections` after that, we make even more progress somewhere.

An idea that's arisen somewhere before is to do more granular change tracking. Low-level operations like `remove_road` and `collapse_intersection` can precisely say which roads and intersections they touch. We could make a helper track those changes, and re-run transformations and operations on modified things. How exactly should that work? Should every transformation right now be re-expressed to operate on just a piece of the network, and we have some rules about retrying all transforms when something changes? Just starting this issue to brainstorm

dabreegster:
https://github.com/a-b-street/osm2streets/issues/136#issuecomment-1330336829 was the description

dabreegster:
Just dumping raw notes related to this:

        - first look at whole network, then do dirty bit things
        - geometry says this looks like u-turn, fix it up
                - what changes now that we determine that?
        - transformations are like rawmap editor actions, higher level
        - cycleway snapping is a case where the collapsing might happen too early
                - how should collapsing actually work? keep the graph the same
                - low-level street network OSM node
                - Junction with 1+ nodes, uses pretrimmed distances
                - DC join needs to radically alter reference line of the Y shape
                
                - 3 levels of stuff?
[...]

PR #167, opened 2023-01-03
Remove the transformation that shrinks overlapping roads.

dabreegster:
#159 
This is indeed too blunt an instrument. I considered keeping it just for A/B Street, but I don't really want to rely on things we're not testing here, and it's only slightly helping in some situations that're broken anyway.

Some particularly notable regressions (where we were just kind of getting lucky before):
Screenshot from 2023-01-03 16-04-50
Screenshot from 2023-01-03 16-04-45
Screenshot from 2023-01-03 16-02-41
Screenshot from 2023-01-03 16-02-35

These were problems that came up in https://github.com/a-b-street/osm2streets/issues/136#issuecomment-1365287556. I think merging this first will make the transition to geometry-as-an-operation easier.


dabreegster:
Noting that in the first example (`seattle_slip_lane`), there's overlap because we incorrectly infer sidewalks

PR #168, opened 2023-01-03
Always maintain road trims and intersection geometry

dabreegster:
Finally finally we're wrapping up #136. The code is hopefully easier to understand now, and the results are slowly improving. The order updates and transformations happen is still sensitive and we'll have to do something more rigorous in #159.

Before this PR, I got rid of shrinking roads in #167 and updated one test's OSM data in 2b89cc120c45940d5a2b3e861c598d01cc78b598. This reduced the diff from this PR and made it easier to understand the subtle ordering between things.

A sample of changes. Nothing gets much worse; lots of places change in kind of neutral ways. Highlights: `kingsway_junction`:
Screenshot from 2023-01-03 18-24-58
Screenshot from 2023-01-03 18-25-05

`quad_intersection`
Screenshot from 2023-01-03 18-26-28
Screenshot from 2023-01-03 18-26-33

`seattle_slip_lane`:
Screenshot from 2023-01-03 18-27-13
Screenshot from 2023-01-03 18-27-09

`tempe_light_rail`:
Screenshot from 2023-01-03 18-28-19
Screenshot from 2023-01-03 18-28-16

Actually going to not merge this one until a review, since it's pretty important


osm2streets/src/intersection.rs
```diff
@@ -196,10 +196,15 @@ impl StreetNetwork {
         intersection.roads = road_centers.into_iter().map(|(r, _, _)| r).collect();
     }
 
-    /// Updates the derived properties of an intersection.
-    ///
+    /// Updates an intersection's derived properties -- geometry (and attached roads) and
```
dabreegster, line 199: I think we always want to do these two when something attached to the intersection changes. We only need to do `sort_roads` in a few cases like inserting a new road though. Maybe in #159, we just start granularly tracking a dirty set of intersections and call this?

osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -51,34 +45,15 @@ impl StreetNetwork {
             let mut trim_roads_for_merging = BTreeMap::new();
             for i in [keep_i, destroy_i] {
                 for road in self.roads_per_intersection(i) {
-                    // If we keep this in there, it might accidentally overwrite the
```
dabreegster, line 54: This comment doesn't make any sense. I think it was pre-opaque IDs, when deleted road borrowed IDs in a funny way

osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -51,34 +45,15 @@ impl StreetNetwork {
...
-                    }
                     // If we're going to delete this later, don't bother!
                     if road.internal_junction_road {
                         continue;
                     }
 
-                    if let Some(pl) = self.estimate_trimmed_geometry(road.id) {
-                        if road.src_i == i {
-                            if trim_roads_for_merging.contains_key(&(road.id, true)) {
-                                panic!(
```
dabreegster, line 67: I got rid of these asserts because I hit a panic in `leeds_cycleway` and I couldn't justify why this is a problem. If we wind up setting this twice, it's fine? It'll be based on `road.center_line`, which goes through `update_geometry` like it should.

osm2streets/src/operations/update_geometry.rs
```diff
@@ -0,0 +1,66 @@
...
+            let untrimmed = road.get_untrimmed_center_line(self.config.driving_side);
+            if let Some(pl) =
+                Road::trim_polyline_both_ends(untrimmed.clone(), road.trim_start, road.trim_end)
+            {
+                road.center_line = pl;
+            } else {
+                error!("{} got trimmed into oblivion, collapse it later", road.id);
+                road.center_line = untrimmed;
+                // Collapse it later
+                road.internal_junction_road = true;
```
dabreegster, line 62: Big question! Right now this can only ever become `true`; it never goes back to `false` if we do something that causes a road to no longer get trimmed away. Not sure how that'd be possible, but not sure how to rephrase how we track this.

osm2streets/src/road.rs
```diff
@@ -115,8 +114,8 @@ impl Road {
         result
     }
 
-    /// Calculates and sets the center_line from reference_line, reference_line_placement
-    /// (and TODO trim_start, trim_end).
```
dabreegster, line 119: I took away the `TODO` cause I'm not sure it makes sense. `update_geometry.rs` is the place that applies the trim. What do you think?

osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -51,34 +45,15 @@ impl StreetNetwork {
...
-                            if trim_roads_for_merging.contains_key(&(road.id, false)) {
-                                panic!(
-                                    "trim_roads_for_merging has a dst_i duplicate for {}",
-                                    road.id
-                                );
-                            }
-                            trim_roads_for_merging.insert((road.id, false), pl.last_pt());
-                        }
+                    if road.src_i == i {
+                        trim_roads_for_merging.insert((road.id, true), road.center_line.first_pt());
```
dabreegster, line 54: So instead of recalculating both intersections from their current state and figuring out the trim, we just use whatever the current trim is. This appears to help `quad_intersection` and `tempe_light_rail` and a few others. I don't entirely understand why.

dabreegster:
Ping @BudgieInWA

dabreegster:
Going to merge and work on a few smaller fixes, but still some questions here I'd love feedback on later

Issue #165, opened 2022-12-31
Stop lines


dabreegster:
Splitting out from #72. The representation:

```rust
pub struct Road {
    ...
    pub stop_line_start: StopLine,
    pub stop_line_end_: StopLine,
}

pub struct StopLine {
    /// Relative to the road's reference_line. Stop lines at the start of the road will have low
    /// values, and at the end will have values closer to the reference_line's length.
    pub vehicle_distance: Distance,
    /// If there is an advanced stop line for cyclists different than the vehicle position, this
    /// specifies it
    pub bike_distance: Option<Distance>,
    pub interruption: TrafficInterruption,
}

/// How a lane of travel is interrupted, as it meets another or ends.
#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
pub enum TrafficInterruption {
    Uninterrupted,
    Yield,
    Stop,
    Signal,
    DeadEnd,
}
```

As a first step, we can assign this based on heuristics and explicitly mapped things in OSM. An example heuristic could be figuring out that minor roads yield or stop and major roads are interrupted at uncontrolled intersections. We can then render it (necessary for debugging/development and also to improve visuals).

Is the representation what we want? I'm not sure what the distances should be relevant to... probably `get_untrimmed_center_line` like `trim_start,end`, not the reference line?

I'll add notes below as I figure out all the things in OSM that help us figure this out.

dabreegster:
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dgive_way
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals#Tag_all_incoming_ways (but are these at the stop positions?)
- https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dasl
- https://wiki.openstreetmap.org/wiki/Tag:road_marking%3Dsolid_stop_line
  - an example in `tempe_split`: https://www.openstreetmap.org/way/656408745. The way doesn't share nodes with the road; we'd have to look for geometric intersection. Is this correct tagging?
  - `highway=crossing` can also imply a stop line, like https://www.openstreetmap.org/node/2239876831. Likely there's a regional design standard about minimum distance between a crosswalk and stop line

@matkoniecz, we're starting to represent different types of stop lines. Do you know of any other tags we should pay attention to? Thanks!

[...]

PR #169, opened 2023-01-04
Start representing and extracting explicit stop lines, with many

dabreegster:
problems. #165

Trying to start with the simple cases quickly exposed lots of issues, so I wanted to get early feedback. I think this is worth merging and iterating on, but there are also major design problems that we'll need to solve.

So far it only records and renders **explicitly** tagged stop lines. I started by setting these up for every road with defaults, but it quickly reveals a problem -- we have lots of short Roads that aren't places where vehicles stop, and so we definitely don't want to draw inferred stop lines in those places!

# Case study: borough_sausage_links

This is a good test case, because stop lines around here are very dramatically far back from the intersection itself. From cycling around in person, this is immediately noticeable and strange with a US perspective. :)

Screenshot from 2023-01-04 11-52-13
Screenshot from 2023-01-04 11-51-47

The two traffic signal nodes tell us about the bike box, and the result looks decent! We happened to get lucky here and preserve the stop lines when we collapse a sausage link.

Screenshot from 2023-01-04 11-53-40
Screenshot from 2023-01-04 11-53-28
On the western arm, we're not so lucky. `transform/sausage_links.rs` will need to merge stop lines for the two sides.

This area reveals another puzzle. Stop lines are sometimes marked on the **outgoing** roads of an intersection too: https://www.openstreetmap.org/node/9733205389. I guess this is to handle cases when turning vehicles don't clear the junction before the pedestrian phase goes on. I'm not sure this is something we want to represent, or how to work it in.

# Case study: seattle_slip_lane

Screenshot from 2023-01-04 11-59-39
The main road's stop lines are coming from https://www.openstreetmap.org/node/4829741818, a signalized crossing node. This case highlights what happens when we're not yet successfully trimming geometry. And also maybe some nuance about basing stop line distances on reference lines.

### DIFF
--- a/osm2streets/src/operations/collapse_short_road.rs
+++ b/osm2streets/src/operations/collapse_short_road.rs
@@ -71,6 +71,7 @@ impl StreetNetwork {
         // If the intersection types differ, upgrade the surviving interesting.
         if destroy_i.control == IntersectionControl::Signalled {
             self.intersections.get_mut(&keep_i).unwrap().control = IntersectionControl::Signalled;
+            // TODO Propagate to stop lines
         }
 
         // Remember the merge
--- a/osm2streets/src/render.rs
+++ b/osm2streets/src/render.rs
@@ -8,8 +8,8 @@ use geom::{ArrowCap, Distance, Line, PolyLine, Polygon, Ring};
 
 use crate::road::RoadEdge;
 use crate::{
-    DebugStreets, Direction, DrivingSide, Intersection, IntersectionID, LaneID, LaneType, Movement,
-    StreetNetwork,
+    DebugStreets, Direction, DrivingSide, Intersection, IntersectionID, LaneID, LaneSpec, LaneType,
+    Movement, Road, StreetNetwork,
 };
 
 impl StreetNetwork {
@@ -180,6 +180,16 @@ impl StreetNetwork {
                             ]),
                         ));
                     }
+                }
+            }
+
+            // Stop line distances are relative to the direction of the road, not the lane!
+            for (lane, center) in road.lane_specs_ltr.iter().zip(lane_centers.iter()) {
+                for (polygon, kind) in draw_stop_lines(lane, center, road) {
+                    pairs.push((
+                        polygon.to_geojson(gps_bounds),
+                        make_props(&[("type", kind.into())]),
+                    ));
                 }
             }
 
@@ -500,5 +510,58 @@ fn make_sidewalk_corners(streets: &StreetNetwork, intersection: &Intersection) -
             results.push(ring.into_polygon());
         }
     }
+    results
+}
+
+fn draw_stop_lines(
+    lane: &LaneSpec,
+    center: &PolyLine,
+    road: &Road,
+) -> Vec<(Polygon, &'static str)> {
+    let mut results = Vec::new();
+
+    if !matches!(
+        lane.lt,
+        LaneType::Driving | LaneType::Bus | LaneType::Biking
+    ) {
+        return results;
+    }
+    let thickness = Distance::meters(0.5);
+
+    let stop_line = if lane.dir == Direction::Fwd {
+        &road.stop_line_end
+    } else {
+        &road.stop_line_start
+    };
+
+    // The vehicle line
+    if let Some(dist) = stop_line.vehicle_distance {
+        if let Ok((pt, angle)) = center.dist_along(dist) {
+            results.push((
+                Line::must_new(
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(90.0)),
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(-90.0)),
+                )
+                .make_polygons(thickness),
+                "vehicle stop line",
+            ));
+        }
+    }
+
+    if let Some(dist) = stop_line.bike_distance {
+        if let Ok((pt, angle)) = center.dist_along(dist) {
+            results.push((
+                Line::must_new(
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(90.0)),
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(-90.0)),
+                )
+                .make_polygons(thickness),
+                "bike stop line",
+            ));
+        }
+    }
+
+    // TODO Change the rendering based on interruption too
+
     results
 }
--- a/osm2streets/src/road.rs
+++ b/osm2streets/src/road.rs
@@ -54,6 +54,42 @@ pub struct Road {
     pub complicated_turn_restrictions: Vec<(RoadID, RoadID)>,
 
     pub lane_specs_ltr: Vec<LaneSpec>,
+
+    pub stop_line_start: StopLine,
+    pub stop_line_end: StopLine,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct StopLine {
+    /// Relative to the road's reference_line. Stop lines at the start of the road will have low
+    /// values, and at the end will have values closer to the reference_line's length. This is only
+    /// set when the stop line is explicitly specified; it's never inferred.
+    pub vehicle_distance: Option<Distance>,
+    /// If there is an advanced stop line for cyclists different than the vehicle position, this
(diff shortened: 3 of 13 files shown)


osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -96,6 +96,7 @@ impl StreetNetwork {
         // If the intersection types differ, upgrade the surviving interesting.
         if destroy_i.control == IntersectionControl::Signalled {
             self.intersections.get_mut(&keep_i).unwrap().control = IntersectionControl::Signalled;
+            // TODO Propagate to stop lines
```
dabreegster, line 99: Ordering dataflow gets even more fun. `TrafficInterruption` and `IntersectionControl` are related.

And a fun case: https://www.openstreetmap.org/node/10249797837 is a `highway=stop` node right before a traffic signal. The mapper meant it as a stop line. But from reading the wiki, I'm not sure this is correct -- I think traffic signal nodes on the incoming ways would be more appropriate?

osm2streets/src/render.rs
```diff
@@ -491,3 +501,56 @@ fn make_sidewalk_corners(streets: &StreetNetwork, intersection: &Intersection) -
...
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(90.0)),
+                    pt.project_away(lane.width / 2.0, angle.rotate_degs(-90.0)),
+                )
+                .make_polygons(thickness),
+                "bike stop line",
+            ));
+        }
+    }
+
+    // TODO Change the rendering based on interruption too
```
dabreegster, line 553: We can iterate here to show give way / stop / signal, and to put a bike symbol in the bike box, etc

osm2streets/src/road.rs
```diff
@@ -54,6 +54,42 @@ pub struct Road {
...
+    pub stop_line_start: StopLine,
+    pub stop_line_end: StopLine,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct StopLine {
+    /// Relative to the road's reference_line. Stop lines at the start of the road will have low
+    /// values, and at the end will have values closer to the reference_line's length. This is only
+    /// set when the stop line is explicitly specified; it's never inferred.
+    pub vehicle_distance: Option<Distance>,
```
dabreegster, line 67: Note the big change here. By default roads have no vehicle or bike stop line, only setting it if explicitly found something in OSM

streets_reader/src/extract.rs
```diff
@@ -190,3 +204,11 @@ impl OsmExtract {
         true
     }
 }
+
+fn parse_dir(x: Option<&String>) -> Option<Direction> {
```
dabreegster, line 208: There's always `both` as a possibility

streets_reader/src/split_ways.rs
```diff
@@ -202,24 +204,86 @@ pub fn split_up_roads(
...
+                    if let Some((dist, _)) = road.reference_line.dist_along_of_point(pt.to_pt2d()) {
+                        let stop_line = if dir == Direction::Fwd {
+                            &mut road.stop_line_end
+                        } else {
+                            &mut road.stop_line_start
+                        };
+                        stop_line.vehicle_distance = Some(dist);
+                        stop_line.interruption = TrafficInterruption::Signal;
+                    }
+                    // TODO If dist_along_of_point fails, it's because we smoothed the line. This
```
dabreegster, line 233: So one of my big questions is this: ultimately I think we want the stop line distance to be based on the untrimmed center, just like `trim_start`. We could do `center_line.project_pt(node_pt)` here, but any part of the code that modifies reference_line or placement will need to do something to preserve the stop line distances.

We could instead store the raw `Pt2D` position of a stop line, and do this `project_pt` thing whenever we need it (just for rendering now, eventually for a traffic simulation rule or for inferring movement geometry or something else). Feels like a messier data model, but maybe if we wrap with a nice API it's OK?

Or maybe when we modify reference line, we need to do so in a kind of "transactional" way -- not directly, but something like `road.update_reference_line(polyline, placement)`. This will first calculate the current `Pt2D` of any stop lines, then change the reference line, then do `project_pt` and calculate the equivalent distance with the new center.

streets_reader/src/split_ways.rs
```diff
@@ -202,24 +204,86 @@ pub fn split_up_roads(
...
+                    }
+                }
+            }
+        }
+    }
+
+    for pt in input.signalized_crossings {
+        if let Some(road) = pt_to_road.get(&pt).and_then(|r| streets.roads.get_mut(r)) {
+            if let Some((dist, _)) = road.reference_line.dist_along_of_point(pt.to_pt2d()) {
+                // We don't know the direction. Arbitrarily snap to the start or end if it's within
```
dabreegster, line 273: This is going to be wrong sometimes, see https://github.com/a-b-street/abstreet/issues/795#issuecomment-963688984

tests/src/bristol_contraflow_cycleway/geometry.json
```diff
@@ -2263,7 +2263,7 @@
         "type": "Polygon"
       },
       "properties": {
-        "control": "Signalled",
+        "control": "Signed",
```
dabreegster, line 2266: This previously was marking the footway crossing just north of https://www.openstreetmap.org/node/21310514 as signalized. This is incorrect; there's actually a pelican crossing at the traffic signal node, but the crossing isn't mapped. There's an unrelated footpath just to the north unrelated to the signal

dabreegster:
Ping @BudgieInWA

streets_reader/src/split_ways.rs
BudgieInWA: I think it would be ideal to have the distance relative to the center line, because that's what we use to practically determine the position of the road later. Reference line probably shouldn't be modified (it's basically just the OSM geom), but the centerline might change. (It would be great to have a transformation that infers missing pacement tags based on lane connectivity, because it is undertagged). I think I like the idea of using a method to update the center line which updates these distances (and the `trim` distances) appropriately.

streets_reader/src/split_ways.rs
BudgieInWA: If we are turning all footways into `Road`s will this be relavent any more? Each crossing will be an intersection in it's own right (which will often get merged into a signalised intersection).

osm2streets/src/render.rs
```diff
@@ -491,3 +501,56 @@ fn make_sidewalk_corners(streets: &StreetNetwork, intersection: &Intersection) -
...
+
+    let stop_line = if lane.dir == Direction::Fwd {
+        &road.stop_line_end
+    } else {
+        &road.stop_line_start
+    };
+
+    // The vehicle line
+    if let Some(dist) = stop_line.vehicle_distance {
+        if let Ok((pt, angle)) = center.dist_along(dist) {
```
BudgieInWA, line 528: Hmm, so `dist` is along the road reference line, not even the road center line, and certainly not the lane centerline. Projecting a point to the lane center might be appropriate here. Or render one stop line for the whole road.

Eventually the projection angle could come from the intersection (because the stop lines are often parallel to the intersecting road).

BudgieInWA (approved):
This is a good exploration of the problem. I am in favour of merging this in so that `stop_line_start` and `_end` are available while I look into major/minor roads in intersections. I don't have much clarity yet on where that info will live, so having traffic interruption for each road end will be interesting to work with. 

I've not grokked how the tagged nodes make their way through `extract.rs` and `spit_ways.rs` into the `Road`s, but I saw nothing surprising in there.

streets_reader/src/split_ways.rs
dabreegster: Sounds good, so I'll try that "transactional" approach when I pick this back up

osm2streets/src/render.rs
dabreegster: Right... for the sake of this PR, I'll keep it as it is, and think about how to make all the distances be properly along the center line, and keep that updated over time.

dabreegster (commented):
I totally lost context on the complexities here, but as you say, we can start making use of the stop lines, plug into the new rendering, and be properly motivated to iterate on any weirdness here

PR #171, opened 2023-01-18
Package the native lib and load it from the .jar

BudgieInWA:
This lets me build a `StreetNetwork.jar` that can be depended on and called from the JOSM2Streets plugin! See #68.

The Java class has been put in the `org.osm2streets` package.
The build script packages the `.class` and `.so` files into a `.jar`.
The native lib is unpacked and loaded directly, instead of searching `java.library.path`.

This iteration only supports linux because it packages `target/debug/osm2streets_java.so` and loads it without checking the current platform. The next iteration should package different target builds and load the one appropriate for the running system.


dabreegster (approved):
Awesome, `run.sh` still works fine for me! Is it pretty straightforward to publish this to Maven? Or maybe that's unnecessary at this point

osm2streets-java/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -ex
+
+javac -d build src/*.java -target 11 -source 11
+cargo build
```
dabreegster, line 5: Of course you'll want `--release` here and below for anything on Maven, or if you're not making changes on this end frequently

BudgieInWA:
I'm going to merge this as-is as a debug configuration, and leave the release configuration bundling multiple build targets etc. as a future problem.

PR #172, opened 2023-01-19
Add layer to GeoJSON output for roads

dabreegster:
Requested by @jakecoppinger. This is https://wiki.openstreetmap.org/wiki/Key:layer, usually 0.

I was unsuccessful in getting Leaflet to use this for z-index. I tried sorting GeoJSON features by the layer, but there are problems between layers. The lane polygon and lane markings layers are separate, but both should respect a global z-index. Leaflet does this with https://leafletjs.com/examples/map-panes/, but you can't change a layer's pane after creation, and you also can't specify a different pane for each feature created from GeoJSON, AFAICT. Hopefully it's easier in MapLibre.

Test diffs are just the new property showing up

### DIFF
--- a/osm2streets/src/render.rs
+++ b/osm2streets/src/render.rs
@@ -42,6 +42,7 @@ impl StreetNetwork {
                     ),
                     ("src_i", road.src_i.0.into()),
                     ("dst_i", road.dst_i.0.into()),
+                    ("layer", road.layer.into()),
                 ]),
             ));
         }
@@ -100,6 +101,7 @@ impl StreetNetwork {
                     make_props(&[
                         ("type", format!("{:?}", lane.lt).into()),
                         ("road", road.id.0.into()),
+                        ("layer", road.layer.into()),
                         ("index", idx.into()),
                         ("width", lane.width.inner_meters().into()),
                         ("direction", format!("{:?}", lane.dir).into()),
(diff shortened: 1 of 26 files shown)


dabreegster:
Oh yeah, two more things to note. First, there's no `layer` defined on intersections. I'm not sure how to define it -- should it be the min or the max of layers of connecting roads?

Second, in A/B Street I've used a transparent effect on roads underneath things. If you have a road at ground level, it'll be implicitly layer 0, and a bridge over it is layer 1 or something. Nothing in OSM tells you the ground-level road is underneath anything. https://github.com/a-b-street/abstreet/blob/main/map_model/src/make/bridges.rs finds these cases and marks them. It's weird to abuse `layer` like this, since it makes it seem like these roads are tunnels. So I haven't carried that trick over here.

jakecoppinger:
> First, there's no layer defined on intersections. I'm not sure how to define it -- should it be the min or the max of layers of connecting roads?

Are multi level intersections expected to occur? Wouldn't a bridge come down to level 0 if joining an intersection (or vice-versa).

jakecoppinger:
Happy with just the layer pass through for my use case. I suppose it doesn't prevent more advanced analysis being added to another property in future.

Thanks so much for working on this!

dabreegster:
> Are multi level intersections expected to occur? Wouldn't a bridge come down to level 0 if joining an intersection (or vice-versa).

Intersections themselves should probably be pretty flat, or that's a terrifying place to stop! So maybe z-ordering them with the minimum layer of roads is the reasonable thing. I'll hold off on adding that until we have some visualization in StreetExplorer for it, so we can sanity check it looks reasonable

PR #176, opened 2023-02-02
Only count relevant lanes when interpreting placements

BudgieInWA:


dabreegster:
This didn't change any tests, is that because the inferred sidewalks were still in the way?

BudgieInWA:
The inferred sidewalks is actually what makes it possible to see the bug! But the only test with placements is in Fremantle... outside of the AU polygon, and in RHT, with the sidewalk on the wrong side!

image

dabreegster:
Oof :(

I commented a picture in the other PR, I think arizona_highways has a few off-ramps where placement would really help. We can fix upstream in OSM (maybe using https://a-b-street.github.io/osm2streets/lane_editor.html#18.28/33.63037/-111.89031 !!) to get more coverage

PR #177, opened 2023-02-02
Use a more conservative understanding of the osm by default

BudgieInWA:
Only use mapped sidewalks and adjust the assumed widths

The 3m default lane width comes from the JOSM Lane Attributes *Map Paint Style*, which is a bit of a historic standard. These widths match up with almost all roads I'v mapped in WA, save for the clearly narrower or wider sections, usually around intersections and stuff.


osm2streets/src/lanes/mod.rs
```diff
@@ -271,20 +271,20 @@ impl LaneSpec {
             // https://www.gov.uk/government/publications/cycle-infrastructure-design-ltn-120 table
```
dabreegster, line 271: These comments get out of date, may as well delete them. We keep talking about it, but I think a relatively easy next step is to

1) Generalize `driving_side` to return the country code anywhere on earth
2) Turn this function into data-driven config, and allow per-country overrides

dabreegster (changes requested):
Thank you for finally ripping off this bandaid. Only one change, then LGTM.

The default option in street explorer should change to match, otherwise it's very strange to see the geometry layer change when you generate details.`index.html`, flip the `checked`

Any concerning changes to the tests? Going through quickly...

`arizona_highways`, we lose some off-ramps, like
Screenshot from 2023-02-02 09-17-30
Screenshot from 2023-02-02 09-17-22
Can placement help here? If so, I'll actually go update the tagging so we can start getting better results in test areas.

`borough_sausage` loses the sausage collapsing, because of the two-part crossing. Not a bad thing, forces us to think about how we want to model this.
Screenshot from 2023-02-02 09-21-36

The experimental dual carriageway classification in northgate still does something, but we'll have to make it understand crossings.

`roosevelt_cycletrack` has this nice symmetric structure now. The experiment stops working there, but this is good motivation to revisit it and see if we can detect the sidewalks also.
Screenshot from 2023-02-02 09-26-40

`seattle_slip_lane` just totally gives up >_<
Screenshot from 2023-02-02 09-27-42
Screenshot from 2023-02-02 09-27-37

`tempe_light_rail` looks good, kind of surreal to see a consolidated junction surrounded by crossings
Screenshot from 2023-02-02 09-29-12

osm2streets/src/lanes/mod.rs
dabreegster: https://github.com/a-b-street/osm2lanes/pull/266 starts this process. I can get us to the point where we have a locale code easily in this repo, and we can figure out how to make use of it

PR #178, opened 2023-02-02
Allow placement to be modified in the web UI

dabreegster:
The results look wrong to me, am I using the wrong tag, or do we need to recalculate more things after changing this?
Screenshot from 2023-02-02 11-08-12

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -5,7 +5,9 @@ use geom::{Distance, LonLat, PolyLine, Polygon};
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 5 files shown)


BudgieInWA:
> The results look wrong to me, am I using the wrong tag, or do we need to recalculate more things after changing this?

The osm way runs along the `left_of:1`st lane. Looking forward along the way, the left of the first lane (the left edge) is where the osm way runs. `right_of:1` would be assumed (down the middle).

dabreegster:
Thanks! Playing around with it more, I think I finally understand what the tagging scheme means. I'll merge this PR (because the re-processing appears correct) and maybe look into an opacity or layer toggle tool (to see the satellite more easily to match things up). I'll hold off on edits to OSM for this case for the moment.

Issue #138, opened 2022-11-29
Determine road geometry before path geometry


BudgieInWA:
I think `Road` and `Intersection` and the first pass of building and transforming a `StreetNetwork` should be vehicle centric. Not because cars are more important or because osm2streets should aim to support them better, but because roads and adjacent infrastructure are designed first to influence the behaviour of the bigger, faster road users.

The shape of the roadway is designed to communicate what cars should be doing, with curbs following smooth lines to match the driving lines and stuff like that. When other road users are prioritised in the design, this is achieved by routing the other users around the car infrastructure, or changing the path that the cars take (and thus changing the shape of the road).

If we know where the cars go, we can predict where the curbs are (or should be) and thus where the pedestrians go. This might hold more true in some places (like where I live in Perth) than others, but I think the model is the best way to construct shapes anyway.

Bikes complicate the matter a bit, but I think about it as three tiers: Car, Bikes then Pedestrians. Pedestrian infrastructure often fits around dedicated bike infrastructure in the same way it does around car infrastructure. And bike infrastructure sometimes fit around car infrastructure in the same way. But sometimes bikes are treated as part of vehicle traffic (such as bike lanes), which complicates things a bit. 

In fact think about railways: They are a tier higher than cars. That is, the shape of a railway tracks isn't determined by any roads, bike paths or footpaths; trains are gonna *train*, and everyone else is gonna respect that. Instead, features of the railway are going to be positioned to leave enough room for the other tiers to fit around it, but those railway features will be represented in the railway data. This is the idea I'm seeing with roads vs pedestrians.

## Actionable takeaways

The "road width" should be the width of the whole carriageway (or both carriageways plus the median) "curb to curb". The "center" of a road should be the center of that width. This matches up with the OSM definition (as best as I can tell; we'd have to do a survey to confirm that mappers place the way at the center of the road width instead of the dividing line, when those don't coincide).

Road trimming and intersection shapes should be calculated by looking just at vehicle lanes. Paths should then be stuck on the side of road shapes. In the first instance, I think A/B street will just need to adjust where geometry is positioned (with then help of a `width()` and `total_width()` probably). Eventually, footpaths around intersections could be calculated with a `shift` operation.

The *intersection kind* calculation already looks at only driving traffic. (As a consequence, only driving movements are calculated. Bike movements will need to be calculated at some point, and maybe factored into `Kind` in some way.)

## Thoughts?

[...]

PR #181, opened 2023-02-07
Introduce a semantic representation for road markings

BudgieInWA:
This PR adds a suite of types for representing road markings, breaking them into a few broad categories based on geometry, then further distinguishing them by their meaning. Each marking type can be enhanced with properties describing the possible variations, such as the ability to overtake across road dividing center lines.

It is going to be a bit tricky to include all the information needed to define all the markings in a locale-independent way, as different locales will use different treatments. For example, the buffer rendering lacks the diagonal lines, because a reference line is needed to determine the angles for the hatches over the length of the median.

The tradeoff though, is a powerful separation of concerns, with `output.rs` determining which markings should be present and `paint.rs` handling the minutiae of rendering them into painted areas. I found it very easy to keep things in the right place during the implementation of this PR.

This changes a lot of the markings arbitrarily, but I'm keen to cut over the goejson rendering to `output.rs` as soon as possible, so we're all looking at and testing the same rendering code. That might mean implementing `get_designations` and an "inspection" geojson layer for all the misc. info that should be displayed in StreetExplorer.


osm2streets/src/lanes/classic.rs
```diff
@@ -594,7 +594,8 @@ fn osm_separation_type(x: &String) -> Option<BufferType> {
     match x.as_ref() {
```
dabreegster, line 594: Just noting, we should try and sync with https://wiki.openstreetmap.org/wiki/Proposed_features/separation#Typical_separation_values and taginfo at some point. I wrote `BufferType` quite a while ago at this point!

osm2streets/src/lanes/mod.rs
```diff
@@ -136,6 +139,16 @@ impl LaneType {
         )
     }
 
+    pub fn to_traffic_mode(&self) -> Option<TrafficMode> {
+        use LaneType::*;
+        match self {
+            Bus | SharedLeftTurn | Driving => Some(TrafficMode::Motor),
+            SharedUse | Biking => Some(TrafficMode::Bike),
```
dabreegster, line 146: Just nitpicking on the meaning here - this is the "fastest" mode that primarily uses this lane type. Of course bikes also use the above lanes

osm2streets/src/marking.rs
```diff
@@ -1,46 +1,117 @@
+// We use geom and stay in map space. Output is done in latlon.
```
dabreegster, line 1: Just noting, the git diff is weird here. This file doesn't exist in `main`. And the PR is showing a merge conflict? The origin of this file is mysterious

osm2streets/src/marking.rs
```diff
@@ -1,46 +1,117 @@
...
     through: bool,
     left: bool,
     right: bool,
     slight_left: bool,
     slight_right: bool,
-    u_turn: bool,
+    reverse: bool,
+}
+impl TurnDirections {
+    pub fn through() -> Self {
```
dabreegster, line 59: We could consider https://crates.io/crates/enumset or similar later to make the API less boilerplatey, but starting simple is good

osm2streets/src/paint.rs
```diff
@@ -0,0 +1,235 @@
...
+use geom::{Angle, Distance, Line, PolyLine, Polygon, Pt2D, Ring};
+
+use crate::lanes::TrafficMode;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct PaintArea {
+    // Because I'm lazy and don't want to make different "map space" and "lonlat space" PaintArea
+    // types, I'm using geo::Polygon, so we can just swap out the coords in place. Not ideal.
+    /// A simple ring.
+    pub area: geo::Polygon,
```
BudgieInWA, line 13: I don't know what the best solution is here. `InternalPaintArea` and `PaintArea` as lonlat and map space alternatives, or output types parametrised by the point type, something like:
```rs
pub struct PaintArea<P: Into<float_pair>> {
    area: Vec<P>,
    color: PaintColor,
}
```

These output types are specifically for exporting to other languages like Java, but the same level of abstraction should be useful if you are using the StreetNetwork coordinate space, like A/B Street does.

osm2streets/src/output.rs
```diff
@@ -89,52 +76,90 @@ impl StreetNetwork {
...
+                    if let Ok(edge_line) = lane_centers[0].shift_left(first_lane.width / 2.0) {
+                        markings.push(Marking::longitudinal(
+                            edge_line,
+                            LaneEdgeKind::edge(),
+                            [LaneType::Buffer(BufferType::Verge), first_lane.lt],
+                        ));
+                    }
+                }
+            }
+            // Add longitudinal markings between lanes.
```
dabreegster, line 106: It is so much nicer to describe the semantics in a layer first, and get to rendering details later!

osm2streets/src/output.rs
```diff
@@ -145,75 +170,56 @@ impl StreetNetwork {
...
-                    center
-                        .must_shift_left((lane.width - thickness) / 2.0)
-                        .make_polygons(thickness)
-                        .into(),
-                ));
-
-                // Diagonal stripes along the lane
-                let step_size = Distance::meters(3.0);
+                // Add arrows along the lane, starting at the end.
+                let step_size = Distance::meters(20.0);
```
dabreegster, line 179: Hmm, we still have to make locale / rendering type decisions here? It'd maybe be neat to say something like "this lane has this symbol spaced along it somehow" and have the rendering layer decide specifics, especially if that layer is the one interpreting locale more. But this is OK too.

osm2streets/src/paint.rs
```diff
@@ -0,0 +1,235 @@
+use crate::{marking, marking::Marking};
+
+// We use geom and stay in map space. Output is done in latlon.
+use geom::{Angle, Distance, Line, PolyLine, Polygon, Pt2D, Ring};
+
+use crate::lanes::TrafficMode;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct PaintArea {
+    // Because I'm lazy and don't want to make different "map space" and "lonlat space" PaintArea
```
dabreegster, line 10: I've thought about making some of the `geom` types take a generic parameter to distinguish this more carefully, but it's felt like overkill so far. (And also all of the math would have to change!)

osm2streets/src/paint.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+impl PaintArea {
+    pub fn new(area: Ring, color: PaintColor) -> Self {
+        Self {
+            area: area.into_polygon().into(),
+            color,
+        }
+    }
+
+    pub fn from(area: Ring) -> Self {
```
dabreegster, line 24: What if we name this `white_area` or something? White's maybe a generic default for paint in most places, but this feels like a detail that's hard to notice at the call site if you just see `PaintArea::from(ring)`

osm2streets/src/paint.rs
```diff
@@ -0,0 +1,235 @@
...
+const LINE_WIDTH_THIN: Distance = Distance::const_meters(0.08);
+const LINE_WIDTH_THICK: Distance = Distance::const_meters(0.30);
+const DASH_LENGTH_SHORT: Distance = Distance::const_meters(1.0);
+const DASH_GAP_SHORT: Distance = Distance::const_meters(1.0);
+const DASH_LENGTH_LONG: Distance = Distance::const_meters(2.0);
+const DASH_GAP_LONG: Distance = Distance::const_meters(4.5);
+
+impl Paint<PolyLine> for marking::Longitudinal {
+    fn paint(&self, separator: &PolyLine) -> Vec<PaintArea> {
+        // TODO incorporate colors throughout instead of only collecting rings:
```
dabreegster, line 73: Maybe something like the [GeomBatch API](https://github.com/a-b-street/abstreet/blob/f42aaafb24b3533b63bf1cedb747dfaa95bbf4b9/widgetry/src/geom/mod.rs#L38) would be helpful eventually

dabreegster (approved):
Everything here looks pretty good. Hard to reason about the correctness or style choices without seeing in St Explorer or JOSM yet, but happy to have a bit of unused code before that lands.

osm2streets/src/lanes/mod.rs
BudgieInWA: Yes, the usecases in this PR care about the "biggest" kind of traffic, which indicates the kind of markings that will be present. A shared foot/cycle path will have cycle markings but a shared bus/cycle lane will have road markings.

There's probably a better method name for it.

osm2streets/src/marking.rs
BudgieInWA: Yes, i know what the wierdness is but haven't merged to fix it yet.

osm2streets/src/paint.rs
dabreegster: I think using the same type for both and just documenting the difference is OK for now. If you have any ideas, we should chat with the georust folks, because they've been thinking about this too: https://docs.rs/geo/latest/geo/geometry/struct.Coord.html

> Unlike Point (which in the future may contain additional information such as an envelope, a precision model, and spatial reference system information), a Coord only contains ordinate values and accessor methods.

osm2streets/src/output.rs
BudgieInWA: Hmm, good point.

osm2streets/src/paint.rs
BudgieInWA: A fully generic geom library with all the algorithms would be hectic!

osm2streets/src/paint.rs
BudgieInWA: Good call.

BudgieInWA:
image

image

osm2streets/src/lanes/mod.rs
BudgieInWA: I have introduced the term "Traffic Class" for this, and added `Rail` as a variant. I realised my goal here is to distinguish between the distinct kinds of infrastructure: paths, tracks, roads and rail, as I've talked about in #138, looking at roads before paths.

osm2streets/src/marking.rs
BudgieInWA: I plan on parsing turn directions (and implementing a basic turn arrow rendering :P) in a follow up PR.

BudgieInWA:
@dabreegster  I've added some docs and did some renames over the last few commits that you should find interesting. I'm going to merge this shortly, in preparation for tackling turn arrows and intersection markings.

Stop lines have also been added, and look great! It's obvious playing in JOSM that many stop lines aren't picked up, but I'm glad we've got some coming through!

image

osm2streets/src/output.rs
```diff
@@ -87,54 +72,179 @@ impl StreetNetwork {
...
+            // While stop lines are measured against the reference line, we need the reference line
+            // offset. Not bothering to support complicated positioning yet.
+            let ref_offset = road.left_edge_offset_of(
+                if let Placement::Consistent(p) = road.reference_line_placement {
+                    p
+                } else {
+                    RoadPosition::Center
+                },
+                self.config.driving_side,
+            );
```
BudgieInWA, line 174: So, we can still make use of the reference line at the end of the process. The code below produces stop lines that will always cross the origination OSM node (I think; as long as placement is `Consistent`). But this shows the benefit of simplifying to a consistently placed `center_line` early on.

osm2streets/src/lib.rs
```diff
@@ -29,9 +29,11 @@ mod geometry;
 mod ids;
 mod intersection;
 mod lanes;
+mod marking;
```
dabreegster, line 32: Stray thought for later: marking, paint, output, and render maybe belong lumped together in a submodule at some point, if these keep growing

osm2streets/src/marking.rs
```diff
@@ -0,0 +1,134 @@
+//! Semantic definitions of the road markings used to control traffic.
+//!
+//! These types describe the *meaning* of the markings, in a locale-independent way, not necessarily
+//! how they look. See [`crate::paint`] for generating locale-dependant renderings of these markings.
+//!
+//! This initial version of these types is based on, and takes its nomenclature from, the Australian
+//! Manual of Uniform Traffic Control Devices (MUTCD). As such, the distinctions between different
+//! kinds of markings that are represented as variants are tailored to support the marking scheme
+//! used in Australia, not necessarily in other parts of the world. The intention is to expand these
```
dabreegster, line 9: I like the approach of starting with specific things you're familiar with. Hasty overgeneralization can be harder to back out of

osm2streets/src/marking.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// A set of turn directions that are allowed.
+// TODO: move to lanes/mod.rs
+pub struct TurnDirections {
+    through: bool,
+    left: bool,
+    right: bool,
+    slight_left: bool,
+    slight_right: bool,
+    reverse: bool,
```
dabreegster, line 76: As in U-turn?

dabreegster (approved):
This is looking awesome!

osm2streets/src/marking.rs
BudgieInWA: Yeah, `reverse` is the OSM tag value for U-turns.

PR #182, opened 2023-02-12
Parse and paint turn arrows

BudgieInWA:
The implicit arrows (drawn when there are no turn arrows specified) shown at the bottom need tweaking, but otherwise this looks serviceable!

`EnumSet` is super easy to use and very ergonomic.

image

### DIFF
--- a/osm2streets/src/lanes/mod.rs
+++ b/osm2streets/src/lanes/mod.rs
@@ -5,7 +5,7 @@ mod placement;
 mod tests;
 mod turns;
 
-use enumset::EnumSet;
+use enumset::{EnumSet, EnumSetType};
 use std::fmt;
 
 use serde::{Deserialize, Serialize};
@@ -264,7 +264,8 @@ pub struct LaneSpec {
     pub lt: LaneType,
     pub dir: Direction,
     pub width: Distance,
-    /// Turn restrictions for this lane. An empty set indicates no indicated restrictions (though
+    /// Turn restrictions for this lane. An empty set represents that no restrictions are indicated
+    /// (though local rules might still dictate restrictions).
     /// Turns for specific vehicle types (`turn:bus:lanes` and such) are not yet captured.
     pub allowed_turns: EnumSet<TurnDirection>,
 }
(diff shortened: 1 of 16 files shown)


osm2streets/src/lanes/mod.rs
```diff
@@ -434,6 +435,23 @@ impl fmt::Display for Direction {
     }
 }
 
+pub type TurnDirections = EnumSet<TurnDirection>;
+
```
BudgieInWA, line 439:
```suggestion
```

osm2streets/src/lanes/mod.rs
```diff
@@ -434,6 +435,23 @@ impl fmt::Display for Direction {
...
+/// A turn direction as defined by <https://wiki.openstreetmap.org/wiki/Key:turn>.
+#[derive(Debug, enumset::EnumSetType)]
```
BudgieInWA, line 441:
```suggestion
#[derive(Debug, EnumSetType)]
```

osm2streets/src/lanes/mod.rs
```diff
@@ -262,10 +264,9 @@ pub struct LaneSpec {
     pub lt: LaneType,
     pub dir: Direction,
     pub width: Distance,
-    /// Turn indications specific to this lane, from
-    /// <https://wiki.openstreetmap.org/wiki/Key:turn>. Turns for specific vehicle types
-    /// (`turn:bus:lanes` and such) are not yet captured.
-    pub turn_restrictions: Vec<String>,
+    /// Turn restrictions for this lane. An empty set indicates no indicated restrictions (though
```
dabreegster, line 267: Comment cuts off?

osm2streets/src/lanes/mod.rs
```diff
@@ -434,6 +435,23 @@ impl fmt::Display for Direction {
...
+    Through,
+    Left,
+    Right,
+    SlightLeft,
+    SlightRight,
+    SharpLeft,
+    SharpRight,
+    MergeLeft,
+    MergeRight,
+    Reverse,
```
dabreegster, line 452: A comment could help -- does this mean a U-turn?

osm2streets/src/transform/separate_cycletracks.rs
```diff
@@ -123,7 +123,7 @@ fn snap(streets: &mut StreetNetwork, input: Cycleway) {
         lt: LaneType::Buffer(BufferType::Planters),
         dir: Direction::Fwd,
         width: LaneSpec::typical_lane_width(LaneType::Buffer(BufferType::Planters)),
-        turn_restrictions: Vec::new(),
+        allowed_turns: Default::default(),
```
dabreegster, line 126: If this gets tedious in the future, we could consider a `LaneSpec::new(LaneType, Direction)` that plugs in `typical_lane_width` does `default` for turns. Maybe when we get to per-lane stop lines or something, it'll be time

dabreegster (approved):
Awesome to see this progress! We're very close to lane-level turning movements now, right? We have the overall movements per directed roads, and now we can glue source and destination lanes together. The big chunk of missing logic is how to handle 3 lanes merging into 2 and things like that.

BudgieInWA:
> We're very close to lane-level turning movements now, right? The big chunk of missing logic is how to handle 3 lanes merging into 2 and things like that.

You're right that we've got (almost) all the data plumbed through that we need. I think they'll get implemented on a case-by-case basis because of complications like mismatched lane counts. Also, classifying which arms of an intersection constitute "slight left" vs "left" and things like that.

The last piece of that puzzle is the [`connectivity` relation](https://wiki.openstreetmap.org/wiki/Relation:connectivity), which answers the question for us in complicated situations (and is our escape hatch to avoid inventing inferences that are too clever). But plumbing it through is complicated, just like complex turn restriction relations.

PR #205, opened 2023-02-18
When we have two roads forming a loop, make sure the hit point we

### LINKED ISSUE #204: Collapsing degenerate roads breaks
https://a-b-street.github.io/osm2streets/#18.41/51.47558/-2.53673
When we collapse degenerate roads, the geometry somewhere blows up, and then this makes us incorrectly collapse a road

https://user-images.githubusercontent.com/1664407/219696918-4731b63b-afd3-4e43-8140-6819e6dc61db.mp4

dabreegster:
find is on the correct side. Fixes #204

The fixed result:
Screenshot from 2023-02-18 12-12-09

No other tests change. An explanation of the problem:
Screenshot from 2023-02-18 12-12-19
While we're calculating geometry for the red intersection, we look for a collision on the right edge of both roads shown (ignore the horizontal one on the right side). There's no collision near the red intersection at all (because it's a T-shape, and this is normal there). But since we have loop roads, there's actually a collision on the cyan side. Since its the only collision, we use that to trim, and the vertical road gets completely trimmed away.

This took me quite an effort to track down. https://github.com/a-b-street/osm2streets/tree/fix_204 has some intermediate logging / more debugging viz. I want to make the interactive debug walkthrough even more detailed, with sub-steps in the middle of the geometry algorithm. Being able to step through piece-by-piece and look at each edge of every road being considered, see the hit point, see the potential trim candidates, etc... that would've been massively helpful here.


osm2streets/src/geometry/general_case.rs
```diff
@@ -43,6 +51,29 @@ pub fn trim_to_corners(
...
+                let other_i = if road1.src_i == results.intersection_id {
+                    road1.dst_i
+                } else {
+                    road1.src_i
+                };
+
+                let dist_to_this_intersection =
+                    intersection_endpoints[&results.intersection_id].dist_to(pt);
+                let dist_to_other_intersection = intersection_endpoints[&other_i].dist_to(pt);
+                if dist_to_other_intersection < dist_to_this_intersection {
```
dabreegster, line 71: This is the best idea I've got for figuring out if we found a hit on the "correct" side. It might be brittle near super tiny roads or something. (But none of our other tests are affected at all by this change, so fine for now.) Any better ideas?

osm2streets/src/road.rs
```diff
@@ -564,6 +564,9 @@ pub(crate) struct RoadEdge {
     /// Pointed into the intersection
     pub pl: PolyLine,
     pub lane: LaneSpec,
+    /// Which edge of a road? Note this is an abuse of DrivingSide; this just means the left or
```
dabreegster, line 567: This was a bit of intermediate debugging that was particularly helpful, just carrying it along for next time

osm2streets/src/geometry/general_case.rs
BudgieInWA: After thinking "just check for intersections at the correct end, lol" and writing my comment, i was thinking about it while walking to the train, and realised "correct end" is difficult to define for arbitrary points along the road! This is the beat i can think of too.

If this becomes a problem, Maybe we can try to detect the "no intersection because the roads are pretty much parallel" case explicitly first. First check if the projection (infinite extension at this end) of both roads cross, then check if each edge endpoint sits within that projection. (Something along those lines, its still fuzzy in my head).

Issue #202, opened 2023-02-17
Collapsing a road breaks


dabreegster:
https://a-b-street.github.io/osm2streets/#20.28/51.47795/-2.52882

https://user-images.githubusercontent.com/1664407/219694301-ce73d775-ac9c-456f-8e68-31ec6b170145.mp4

Not really sure what's going wrong here yet, but the final result having no intersection geometry is causing downstream problems in the LTN tool. Fixing the root cause would be great.

BudgieInWA:
Something that might be relevant about this crash: after the collapse, there is a road that joins the intersection at both ends! I wonder if that is the cause for all these crashes.

The small intersection (with "S" underneath it) is merged with the large "tion Way" intersection, and that tiny sliver of road at the top also connects both. After the merge, that tiny sliver is no longer trimmed, and reveals it's full length.

PR #206, opened 2023-02-23
Immediately remove loops created by collapsing a short road. #202

dabreegster:
Ben, your observation in the issue was correct -- one of the operations created a tiny loop road, and seemingly nothing tried to later remove it. We would only mark it as `internal_junction_road` if it gets trimmed into oblivion, but calculating geometry for this case is... confusing to say the least, possibly just undefined.

This PR improves a few test cases substantially.
Screenshot from 2023-02-23 14-00-09
Taipei after:
Screenshot from 2023-02-23 14-00-13

Screenshot from 2023-02-23 13-59-32
Seattle slip lane after:
Screenshot from 2023-02-23 13-59-36

I'm not going to add a new test case for the area in the issue, since existing tests seem to capture the same problem. That area indeed comes out great with this PR:
Screenshot from 2023-02-23 14-00-54


osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -2,18 +2,12 @@ use std::collections::BTreeMap;
...
 
 impl StreetNetwork {
-    /// Collapses a road, merging the two intersections together. Returns (the surviving
-    /// intersection, the deleted intersection). (Note these will be the same when the road's a
-    /// loop.)
-    pub fn collapse_short_road(
-        &mut self,
-        short_r: RoadID,
-    ) -> Result<(IntersectionID, IntersectionID)> {
+    /// Collapses a road, merging the two intersections together. This may also delete other roads
```
dabreegster, line 8: Nothing made use of the return value

PR #207, opened 2023-03-05
Add turn movements

BudgieInWA:
We don't need all of this before merging, but here's the plan so far:
- [x] Rename `Movement` -> `Turn` (road level), add `TurnMovement` (lane level)
- [x] Calculate `TurnMovement`s for degenerate (2 road) intersections
- [ ] Calculate movements for `IntersectionKind::Connection`s with more than two roads.
- [ ] Do some default allocations for arbitrarily complex intersections.
- [ ] Add `TurnKind` to determine what kind of lane markings the `TurnMovement`s get.
- [x] Render "turn guide" lane markings for "turn" turns.
- [ ] Render ordinary lane markings for "continuation" turns (ideally, reusing the normal logic).

This is making some good progress towards lane level movements. In fact, some movements are calculated and drawn! I added the movement centrelines in post, the left edge of the movements are being drawn:
 
image

We need angled road ends, #95, to draw these markings properly.

### DIFF
--- a/osm2streets/src/intersection.rs
+++ b/osm2streets/src/intersection.rs
@@ -3,8 +3,7 @@ use std::collections::BTreeMap;
 use geom::{Circle, Distance, PolyLine, Polygon, Pt2D};
 use serde::{Deserialize, Serialize};
 
-use crate::lanes::LtrLaneNum;
-use crate::{osm, DrivingSide, IntersectionID, RoadID, StreetNetwork};
+use crate::{movements, osm, DrivingSide, IntersectionID, LaneID, RoadID, StreetNetwork};
 use TrafficConflict::*;
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
@@ -99,11 +98,11 @@ pub struct Turn {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct TurnMovement {
     /// The lane of the "from" road that the movement starts in.
-    pub from: LtrLaneNum,
+    pub from: LaneID,
     /// The lane of the "to" road that the movement ends in.
-    pub to: LtrLaneNum,
+    pub to: LaneID,
     /// The path of travel.
-    pub center_line: PolyLine,
+    pub path: PolyLine,
 }
 
 impl Intersection {
@@ -218,11 +217,12 @@ impl StreetNetwork {
         intersection.roads = road_centers.into_iter().map(|(r, _, _)| r).collect();
     }
 
-    /// Updates an intersection's derived properties -- geometry (and attached roads) and
+    /// Updates an intersection's derived properties -- geometry (and attached roads), turns and
     /// movements.
     pub fn update_i(&mut self, i: IntersectionID) {
         self.update_geometry(i);
         self.update_turns(i);
+        self.update_movements(i);
     }
 
     /// The kind and turns of a `MapEdge` are handled independently, so this method skips them.
@@ -316,6 +316,39 @@ impl StreetNetwork {
                 Cross => IntersectionKind::Intersection,
             },
         )
+    }
+
+    /// Calculates and fills in the turn movements for an intersection.
+    fn update_movements(&mut self, i: IntersectionID) {
+        let intersection = self.intersections.get_mut(&i).unwrap();
+
+        match intersection.kind {
+            IntersectionKind::Connection => {
+                for turn in intersection.turns.iter_mut() {
+                    let src_road = self.roads.get(&turn.from).unwrap();
+                    let src_end = src_road.end_of(i).unwrap();
+                    let dst_road = self.roads.get(&turn.to).unwrap();
+                    let dst_end = dst_road.end_of(i).unwrap();
+
+                    // 1. TODO If a connectivity relation exists, use that.
+
+                    if let Some(movements) =
+                        movements::from_placement((src_road, src_end), (dst_road, dst_end))
+                    {
+                        turn.movements = Some(movements);
+                    } else if let Some(movements) = movements::default(
+                        (src_road, src_end),
+                        (dst_road, dst_end),
+                        self.config.driving_side,
+                    ) {
+                        turn.movements = Some(movements);
+                    }
+                }
+            }
+            _ => {
+                // Not implemented yet
+            }
+        }
     }
 }
 
--- a/osm2streets/src/intersection.rs
+++ b/osm2streets/src/intersection.rs
@@ -1,8 +1,9 @@
 use std::collections::BTreeMap;
 
-use geom::{Circle, Distance, Polygon, Pt2D};
+use geom::{Circle, Distance, PolyLine, Polygon, Pt2D};
 use serde::{Deserialize, Serialize};
 
+use crate::lanes::LtrLaneNum;
 use crate::{osm, DrivingSide, IntersectionID, RoadID, StreetNetwork};
 use TrafficConflict::*;
 
@@ -24,7 +25,7 @@ pub struct Intersection {
     /// All roads connected to this intersection. They may be incoming or outgoing relative to this
     /// intersection. They're ordered clockwise around the intersection.
     pub roads: Vec<RoadID>,
-    pub movements: Vec<Movement>,
+    pub turns: Vec<Turn>,
 
     // true if src_i matches this intersection (or the deleted/consolidated one, whatever)
     // TODO Store start/end trim distance on _every_ road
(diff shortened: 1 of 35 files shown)


osm2streets/src/intersection.rs
```diff
@@ -295,6 +317,39 @@ impl StreetNetwork {
...
         )
     }
+
+    /// Calculates and fills in the turn movements for an intersection.
+    fn update_movements(&mut self, i: IntersectionID) {
+        let intersection = self.intersections.get_mut(&i).unwrap();
+
+        match intersection.kind {
+            IntersectionKind::Connection => {
+                for turn in intersection.turns.iter_mut() {
```
dabreegster, line 327:
```suggestion
                for turn in &mut intersection.turns {
```

Optional, not really sure what style is preferable. If we later need to do anything else with the iterator, like `.enumerate()`, then how you have it is nicer (because we'd switch back to it)

osm2streets/src/intersection.rs
```diff
@@ -81,8 +81,29 @@ pub enum IntersectionControl {
...
+    pub to: RoadID,
+    pub via: IntersectionID,
+    // direction: TurnDirection,
+    /// The lane level movements, if they have been calculated.
+    pub movements: Option<Vec<TurnMovement>>,
+}
+
+/// A specific path that a vehicle can take through a turn.
+///
+/// Lane numbers are counted from the perspective of the direction of the turn.
```
dabreegster, line 97: This stumped me for a few minutes; can we add an example or clarify the comment? LaneIDs normally index into `lanes_ltr`. Normally whatever direction we're thinking about a road, the lane ID is "absolute" -- 0 means the left side of the road based on its OSM-defined orientation. It feels like here we have a sudden explosion of possibilities, if we point from the right side of one road to the left side of another in the backwards direction, for instance...

And actually now I'm wondering if this refers to `LtrLaneNum` and not `LaneID` and am more confused :\

osm2streets/src/movements.rs
```diff
@@ -0,0 +1,187 @@
...
+            if !advanced {
+                break;
+            }
+        }
+    }
+    Some(result)
+}
+
+/// Limitations: ignores both ways lanes
+pub fn default(
```
dabreegster, line 101: Getting tests of some form here is important IMHO. We could maybe make unit tests if we can succinctly express input and output, or maybe it's better to do our usual snapshot+diff testing approach, and get more rendering into Street Explorer

osm2streets/src/output.rs
```diff
@@ -161,6 +165,33 @@ impl StreetNetwork {
                 }
             }
 
+            // Lanes through intersections
+            for inter in self.intersections.values() {
+                for Turn { movements, .. } in &inter.turns {
+                    let Some(movements) = movements else { continue };
```
dabreegster, line 171: OK I actually hadn't understood the point of let-else till this; very cool!

osm2streets/src/output.rs
```diff
@@ -304,6 +335,10 @@ impl StreetNetwork {
 
         areas
     }
+
+    fn lane(&self, lane: LaneID) -> Option<&LaneSpec> {
```
dabreegster, line 339: Could we move to `road.rs` in the `impl StreetNetwork` block?

osm2streets/src/render.rs
```diff
@@ -369,10 +389,17 @@ impl StreetNetwork {
 
         let mut pairs = Vec::new();
         for ((from, _), polygon) in self.movements_for_intersection(i) {
-            if from == road.id {
+            if from == id {
                 pairs.push((polygon.to_geojson(Some(&self.gps_bounds)), make_props(&[])));
             }
         }
+        if pairs.is_empty() {
```
dabreegster, line 396: In general, when should we expect turns to exist, but not movements? Right now `update_i` does both. Is this temporary before we have movements defined in all cases?

osm2streets/src/road.rs
```diff
@@ -467,6 +515,23 @@ impl Road {
         Ok((left, right))
     }
 
+    pub fn intersection_at(&self, end: RoadEnd) -> IntersectionID {
```
dabreegster, line 518: These new helpers are nice; there's probably many places throughout the code where they can come in handy!

dabreegster (commented):
Really cool to see this!

Not that I'm likely to play with the JOSM side soon, but is https://github.com/BudgieInWA/JOSM2Streets/ still up-to-date, or are there new rendering goodies corresponding to this?

osm2streets/src/output.rs
BudgieInWA: Yep, a bunch of nesting can be replaced with early returns! The `else` block can run any code, as long as it diverges (doesn't run to the end).

osm2streets/src/render.rs
```diff
@@ -66,12 +67,12 @@ impl StreetNetwork {
...
                     (
-                        "movements",
+                        "turns",
                         serde_json::Value::Array(
                             intersection
-                                .movements
+                                .turns
                                 .iter()
-                                .map(|(a, b)| format!("{a} -> {b}").into())
+                                .map(|Turn { from, to, .. }| format!("{from} -> {to}").into())
```
BudgieInWA, line 75: I need to add `TurnMovements` to the test output to observe these changes.

osm2streets/src/render.rs
BudgieInWA: Yes, that was a temporary measure. I don't know where it will land though. I think we will be able to calculate movements any time we calculate turns, so we will be able to remove the `Option`.

I was playing with using an `Option` in places where data is only present after some step of the algorithm, instead of using `*::dummy()` kind of values.

osm2streets/src/road.rs
BudgieInWA: Using `RoadEnd` has been indispensable while dealing with the possibility space explosion in `crate::movements`, and these helpers remove a lot of noise from those implementation sites that make it possible to see the (still very complicated) logic there.

osm2streets/src/movements.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    // Pair up the lanes, matching the inside lanes to each other.
+    loop {
+        let (src_lane, src_pt) = src_lanes[src_i];
+        let (dst_lane, dst_pt) = dst_lanes[dst_i];
+        result.push(TurnMovement {
+            from: src_lane,
+            to: dst_lane,
+            path: PolyLine::new(vec![src_pt, dst_pt])
+                .unwrap_or_else(|_| PolyLine::must_new(vec![src_pt, Pt2D::new(0.0, 0.0), dst_pt])),
```
BudgieInWA, line 141: This is where I would want to create a nice curved path, instead of just jamming two points in.

After working with it, it might be better to remove the geometry aspect from here and calculate it later, or on the fly. That matches up better with where geometry of other kinds is calculated.

Also, the `or_else` polyline sometimes crashes when the first one does, but not always!

osm2streets/src/movements.rs
dabreegster: Hmm... nothing would stop us from drawing something synthetic in JOSM and adding the .osm to `tests/` like we normally do

osm2streets/src/movements.rs
dabreegster: On-the-fly geom could be nice, yeah, and would have reduce file size requirements maybe. (But it's not a clear-cut tradeoff; lazily calculating in a UI later can be sluggish.)

> Also, the or_else polyline sometimes crashes when the first one does, but not always!

I actually didn't look at this carefully; inserting (0, 0) in the middle is an attempt to make the total line length larger? In what kind of cases is the above `PolyLine::new(vec![src_pt, dst_pt])` failing -- when the two points are very close together? What should we do in those cases -- maybe it's valid to just have no movement geometry?

BudgieInWA: Yes, when the points are close together. Given our current approach, we shouldn't have any zero length movements, but I think it happens coincidently sometimes.

I'm going to experiment with removing `path` from `TurnMovement` so that turns and movement are only concerned with connectivity for now (and don't have to be recalculated if trim distances change, or something like that). I want to think about storing lane centerlines and movement paths in the same way.

BudgieInWA:
> Really cool to see this!
> 
> Not that I'm likely to play with the JOSM side soon, but is [BudgieInWA/JOSM2Streets](https://github.com/BudgieInWA/JOSM2Streets/) still up-to-date, or are there new rendering goodies corresponding to this?

The JOSM2Streets build embeds `StreetNetwork.jar`, which in turn embeds the rust lib. If the osm2streets java classes haven't changed their API, then you can build your own `StreetNetwork.jar` and replace the copy in the JOSM2Streets repo before building. You'd need to do that for these changes, I haven't pushed up the new JAR yet.

oneil512:
Hi, is this still being considered?

dabreegster:
I don't think Ben or I have had time to work on this, so not anytime soon. Are you interested in picking it up?

paaspaas00:
Hi @BudgieInWA is this being worked on?

PR #208, opened 2023-03-05
Speed up two slow operations, which were unnecessarily searching all …

dabreegster:
…roads linearly.

Correctness: Tests stay the same, but we're not currently checking anything that turn restrictions would affect.

I had an enormous map with ~300,000 roads that took 93s to collapse short roads! I don't have access to the computer processing that right now, but I found a smaller map that took 15s each for the two operations (collapsing short roads and degenerate intersections). With this PR, drops down to 0.3s and 2s, so I'll consider this a good-enough win.

### DIFF
--- a/osm2streets/src/operations/collapse_short_road.rs
+++ b/osm2streets/src/operations/collapse_short_road.rs
@@ -8,6 +8,10 @@ impl StreetNetwork {
     /// Collapses a road, merging the two intersections together. This may also delete other roads
     /// connected two the merged intersection, if they become a loop on that intersection.
     pub fn collapse_short_road(&mut self, short_r: RoadID) -> Result<()> {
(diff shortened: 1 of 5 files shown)


osm2streets/src/operations/collapse_short_road.rs
```diff
@@ -35,28 +39,26 @@ impl StreetNetwork {
...
         // Retain some geometry...
-        {
-            let mut trim_roads_for_merging = BTreeMap::new();
-            for i in [keep_i, destroy_i] {
-                for road in self.roads_per_intersection(i) {
-                    // If we're going to delete this later, don't bother!
-                    if road.internal_junction_road {
-                        continue;
-                    }
+        let mut trim_roads_for_merging = BTreeMap::new();
```
dabreegster, line 42: Just removing the unnecessary { } grouping here

PR #210, opened 2023-03-10
Work on cycleway snapping for St George

dabreegster:
This is just an experiment, but I wanted a nice place to post screenshots and explain problems as I'm working through them


dabreegster:
# Can we clean up short roads as we go along?

Screenshot from 2023-03-10 11-11-08
Screenshot from 2023-03-10 11-11-13

After snapping one cycleway, the two short perpendicular pieces are very messy. Could we immediately merge the short bits? If the adjacent piece of cycleway hasn't been snapped, then it won't count as a degenerate intersection yet. Could we try merging anyway?

# Debugging needed: show adjacent roads/intersections

The above also shows something that'd be helpful -- I want to pick a road or an intersection and see the stuff immediately connected to it. Or visualize the connectivity graph on top of this. Sometimes in the middle of transformations, it's hard to understand what's actually connected.

# Pre-merged  case

Screenshot from 2023-03-10 11-14-54

The left intersection doesn't have that short road, because the first run of intersection geometry meant the short road collapsed into oblivion and was merged. I'm trying in this PR to detect and handle that case. It works in this one case, at least.

... but after a few merges nearby, things look very odd:
Screenshot from 2023-03-10 11-16-14
This is one of the reasons I'd like to see roads/intersection  connectivity more explicitly. Is the graph correct here, and just the geometry too small? I think using the "original" reference lines of the tiny side-road would be good here. I think the original trims from when the cycletrack was separate are too short after making the main road+cycleway thicker.

# Short roads need to be perpendicular!

Screenshot from 2023-03-10 11-18-44

That top-left part is wrong. I don't think it's safe for us to snap the cycleway the same way here. The short green segment isn't perpendicular to the two parallel roads; it kind of merges in. Do we also need to check for angle?

After snapping anyway, it doesn't look great:
Screenshot from 2023-03-10 11-19-44

# Very problematic matching

Screenshot from 2023-03-10 11-20-18

It picked one of the perpendicular short segments as the cycleway! Completely wrong... An angle check here wouldn't help; the "main road" it found is parallel, because it found another perpendicular short segment. How can we distinguish these cases?

# Not intended, but nice

Screenshot from 2023-03-10 11-22-34
The inner roundabout bit isn't a typical case where the cycletrack is strictly parallel to the main road; the distance of the buffer varies in a detailed way. But the snapping result looks OK:
Screenshot from 2023-03-10 11-22-42

dabreegster:
> Can we clean up short roads as we go along?

Maybe!
Screenshot from 2023-03-10 11-38-45
Looks much better when we do:
Screenshot from 2023-03-10 11-38-59
But why is that bottom-right intersection so strange? It's not finding the corner between the perpendicular big roads. Do we need to recalculate intersection geometry when we thicken the big road?
... yes. Much better:
Screenshot from 2023-03-10 11-42-49

# Extending side roads

The side roads originally reached into the middle of the main road. But because we chop up things into pieces between the original cycletrack, the final result is that we "lose" the rectangularness of it:
Screenshot from 2023-03-10 11-45-36

# Adding lanes to one side

The problem is visible here, but over in `roosevelt_cycletrack` even more so:
Screenshot from 2023-03-10 11-46-53
When we try and snap the sidewalks, we don't want the driving lanes to shift position. And the sidewalks should wind up in roughly the same place; the different amount of gap implicitly tells us something about the width of the sidewalk or the width of the buffer between the road and the sidewalk. How can we preserve that? Placement? Adjusting the width of lanes we tack on?

PR #211, opened 2023-03-10
Add more debugging capabilities to Street Explorer

dabreegster:
There are a bunch of debug steps, so a dropdown is very convenient to jump to the relevant ones:

https://user-images.githubusercontent.com/1664407/224373696-ac85a4c5-adb7-4d6f-a0f6-40a3abd50d88.mp4

And interactively modifying the network can be helpful too:

https://user-images.githubusercontent.com/1664407/224373986-9e7817df-c79c-488d-ba8e-5423677b7b9d.mp4

Next up: a way to experimentally snap one piece of separate cycleway/footway.

### DIFF
--- a/street-explorer/js/main.js
+++ b/street-explorer/js/main.js
@@ -133,13 +133,14 @@ class TestCase {
     const geometry = await loadFile(prefix + "geometry.json");
     const boundary = JSON.parse(await loadFile(prefix + "boundary.json"));
 
-    const geometryLayer = makePlainGeoJsonLayer(geometry);
+    // TODO The geometry layer is vastly simplified from makePlainGeoJsonLayer, because we don't have the network to plumb in.
+    const geometryLayer = L.geoJSON(JSON.parse(geometry));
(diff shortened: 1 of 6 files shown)


street-explorer/js/main.js
```diff
@@ -133,13 +133,14 @@ class TestCase {
     const geometry = await loadFile(prefix + "geometry.json");
     const boundary = JSON.parse(await loadFile(prefix + "boundary.json"));
 
-    const geometryLayer = makePlainGeoJsonLayer(geometry);
+    // TODO The geometry layer is vastly simplified from makePlainGeoJsonLayer, because we don't have the network to plumb in.
```
dabreegster, line 136: When you first load a test from a file, the "geometry" layer is totally minimal and not clickable now:
Screenshot from 2023-03-10 16-51-38
I couldn't find an easy way to plumb around all the stuff we need for the more detailed clickable intersection polygon layer.

PR #212, opened 2023-03-10
Start reviving the cycleway snapping / sidepath zipping experiment

dabreegster:
First, rename "snap cycletrack" to "zip sidepath". Ben, I think you've used "zip" before as the phrasing for this kind of operation. "Sidepath" matches https://wiki.openstreetmap.org/wiki/Proposed_features/Key:is_sidepath and stops being specific to cycleways. We will also want to try this for sidewalks:
Screenshot from 2023-03-10 19-28-01

And second, pull the detection + repair logic into an operation, so that we can interactively call it from the UI:

https://user-images.githubusercontent.com/1664407/224409389-667e6670-9afd-4c7a-835d-cebbdd040726.mp4

These are smaller+simpler changes, paving the way for more complicated changes to the logic in #210.


osm2streets/src/transform/collapse_intersections.rs
```diff
@@ -80,10 +80,6 @@ fn should_collapse(road1: &Road, road2: &Road) -> Result<()> {
         _ => bail!("one of the placements isn't consistent"),
     }
 
-    if road1.is_cycleway() && road2.is_cycleway() {
```
dabreegster, line 83: This had no effect. Even if it was false, we would choose to collapse.

PR #217, opened 2023-05-16
Add speed limit to Roads

dabreegster:
I need this for https://github.com/acteng/atip/issues/69 (so CC @Sparrow0hawk and @Pete-Y-CS).

I'm choosing an implementation that has less magic/guessing than https://github.com/a-b-street/abstreet/blob/6c44da83d36d0b3197525b8ca834e57a66493d24/map_model/src/objects/road.rs#L246, and also one much simpler (in terms of error reporting especially!) than https://github.com/a-b-street/osm2lanes/blob/0c166e5917748cd2dcb6eb3b798cf38136e776e6/osm2lanes/src/metric.rs.

In the future, it'd be awesome to integrate with https://github.com/westnordost/osm-legal-default-speeds.

### DIFF
--- a/osm2streets/src/geometry/mod.rs
+++ b/osm2streets/src/geometry/mod.rs
@@ -69,6 +69,7 @@ impl InputRoad {
             name: None,
             internal_junction_road: false,
             layer: 0,
+            speed_limit: None,
             reference_line: PolyLine::dummy(),
             reference_line_placement: crate::lanes::Placement::Transition,
             trim_start: Distance::ZERO,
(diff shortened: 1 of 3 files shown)


tordans:
JFYI, for our Project https://radverkehrsatlas.de/ we also process maxspeed (https://github.com/FixMyBerlin/atlas-geo/tree/develop/app/process/maxspeed, focus is Germany only, for now) and have a very basic fallback when no exact value is given but one of the many source/zone tags are present https://github.com/FixMyBerlin/atlas-geo/blob/develop/app/process/maxspeed/MaxspeedFromZone.lua.

dabreegster:
Thanks for the pointer! In the short term based on what's needed in ATIP (my current work project, UK focused), we might have a similar UK-specific fallback, or maybe invest in using osm-legal-default-speeds directly or indirectly.

PR #218, opened 2023-07-07
Support parking:<position>=lane syntax

ginnyTheCat:
Before osm2street would only look at the `parking:lane:<position>=parallel|diagonal|perpendicular` tags. However, as the [wiki notes](https://wiki.openstreetmap.org/wiki/Key:parking:lane), this way of tagging is being replaced by the `parking:<position>=lane` syntax. Because many ways still use the old syntax this PR only adds the new syntax without removing support for the old one.

I'm unsure why `tags.is` requires `Vec` instead of `&Vec`  though.

|Before | After |
| --- | --- |
|image | image |

### DIFF
--- a/osm2streets/src/lanes/classic.rs
+++ b/osm2streets/src/lanes/classic.rs
@@ -462,10 +462,15 @@ fn add_bike_lanes(
 
 fn add_parking_lanes(fwd_side: &mut Vec<LaneSpec>, back_side: &mut Vec<LaneSpec>, tags: &Tags) {
     let has_parking = vec!["parallel", "diagonal", "perpendicular"];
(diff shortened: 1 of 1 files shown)


dabreegster (approved):
Thank you for the PR! I see your screenshot is from A/B Street; I'll try and get around to bumping the osm2streets version used there and doing another release soonish.

> I'm unsure why tags.is requires Vec instead of &Vec though.

No good reason; I think I wrote [is_any](https://github.com/a-b-street/abstreet/blob/ffe995556ceb35d8085e8c459d811dd37ed04005/abstutil/src/collections.rs#L309) before I knew Rust well. We could refactor it. (There's also an awkward circular dependency between the abstreet repo and osm2streets for things like this abstutil crate, which makes it a few steps to do refactors like this.)

PR #219, opened 2023-07-09
Rewrite the web app in Svelte

dabreegster:
Street explorer and the lane editor have thus far been written in framework-less JS. I've found it rather painful as the web apps have grown, and have felt slowed down when trying to rapidly experiment with new ideas (thickening road center lines until they hit buildings, live-tuning parameters for edge costs in pathfinding, extracting POIs from OSM). I've become quite comfortable in Svelte, TS, and MapLibre over the past months from work projects. In the background, I've slowly been exploring rewriting the osm2streets apps using this stack in https://github.com/dabreegster/osm2svelte. This PR is finally bringing the results of that experiment over here, replacing the old JS app!

Existing JS app: https://a-b-street.github.io/osm2streets
New Svelte app: https://dabreegster.github.io/osm2svelte  (I'll change the github workflows to deploy here once this PR is merged)

Quick demo of the street explorer:

https://github.com/a-b-street/osm2streets/assets/1664407/73773556-a6a8-41e5-b20e-a9a5ffe4c8e9

And the lane editor:

https://github.com/a-b-street/osm2streets/assets/1664407/bad8bd88-7ac1-4ec3-9131-9474a0f53d09

## Big changes

- Change from Leaflet to MapLibre GL. I'm much more familiar with the latter from work. The API for working with MapLibre is harder than Leaflet -- attaching popups or changing opacity for a hover effect as some examples -- but getting Leaflet to work nicely with Svelte was a bit of a challenge anyway. The `osm2streets-svelte` library has some opinionated helpers like `Layer.svelte` that make the end-user API quite nice.
- The UI is more reactive. The user initially sees generated layers; there's no "Generate details" button. After changing import settings, the change happens immediately.
- The road ordering debug layer is now shown with popups per intersection, rather than for everywhere

There are a few unimplemented features from the previous version, notably debug layers. I'll file an issue with those tasks, but I don't want to block cutting over to this on that, because small iterations are easier once I'm using the new UI day-to-day and am motivated to improve it. :)

## A common library

I'm going to keep the osm2svelte repo around and use it for my wild experiments. After this PR is merged, I'm going to delete dupe code from there and depend on things here by npm github dependencies. My intention is to later split out the `osm2streets-svelte` subdirectory as a standalone NPM module, document it better, and encourage people to build their own web apps making use of it.

## Reviewing

Happy for a real code review, but I recognize this is absolutely massive. @BudgieInWA, if you're happy with the overall tech choices, that's the main thing I'm looking for.

CCing @andrewphilipsmith for osm2streets / A/B Street context. When we start "lifting" things like the turn restriction code to happen at the osm2streets layer, this is a place for us to interactively debug and make simple UIs for diff testing. `npm run wasm && npm run dev` is orders of magnitude faster than rebuilding A/B Street for tiny changes in one base crate, because we don't need to link to dozens of dependencies to effectively reimplement a web browser. :)

And CCing @yongrenjie if there's any interest in discussing Svelte style. I've made some choices here I'm happy with, but other things (ordering and reactive statements) are very awkward.

I'll leave a bunch of comments on the code to try and intro some Svelte and explain weird choices.


.github/workflows/web.yml
dabreegster, line 1: I'll add this back after this PR is merged. Otherwise the current app gets replaced immediately and comparison is harder

web/README.md
```diff
@@ -0,0 +1,34 @@
+# osm2streets Svelte web app
+
+This directory contains web apps for interactively exploring and using osm2streets. The current deployed version is at <https://a-b-street.github.io/osm2streets/>. The apps include:
+
+- Street Explorer: import OSM data from Overpass or test cases, visualize and debug the output
+- Lane editor: modify OSM way tags and visualize the results, to make editing lane tags easier
+
+## Installation
```
dabreegster, line 8: Set up instructions are here

web/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "wasm": "wasm-pack build --dev --target web ../osm2streets-js",
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview",
+    "check": "svelte-check --tsconfig ./tsconfig.json",
+    "fmt": "npx prettier --write *.html src/**",
+    "list-tests": "./public/list_tests.sh"
```
dabreegster, line 13: Have to manually run this when adding new tests. There could still be nicer solutions here...

web/src/lane-editor/App.svelte
```diff
@@ -0,0 +1,43 @@
...
+  import Layout from "../osm2streets-svelte/Layout.svelte";
+  import Map from "../osm2streets-svelte/Map.svelte";
+  import EditWayControls from "./EditWayControls.svelte";
+
+  onMount(async () => {
+    await init();
+  });
+</script>
+
+<Layout>
```
dabreegster, line 16: Svelte is component based, so we can very naturally express the overall organization of the app and share components so much more easily than with vanilla JS

web/src/lane-editor/EditWayControls.svelte
```diff
@@ -0,0 +1,52 @@
...
+
+  let layerStyle = {
+    type: "fill",
+    paint: {
+      "fill-color": "red",
+      "fill-opacity": 0.3,
+    },
+  };
+
+  $: {
```
dabreegster, line 22: https://svelte.dev/tutorial/reactive-statements

This runs whenever anything referenced inside of here changes. In this case, `$clickedLane`. (The `$` is the syntax for accessing a store: https://svelte.dev/tutorial/auto-subscriptions)

web/src/osm2streets-svelte/Layer.svelte
```diff
@@ -0,0 +1,158 @@
...
+  import type {
+    GeoJSONSource,
+    MapLayerMouseEvent,
+    MapMouseEvent,
+  } from "maplibre-gl";
+  import { onDestroy, onMount } from "svelte";
+  import { map } from "./store";
+  import { emptyGeojson, getLayerZorder } from "./utils";
+
+  // This component manages a GeoJSON source with exactly one MapLibre layer attached to it.
```
dabreegster, line 12: A little bit of complexity here. Inspired by https://github.com/dimfeld/svelte-maplibre/, and one of the possible next steps is to refactor and just use that instead

web/src/osm2streets-svelte/Layer.svelte
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  function onMouseMove(e: MapLayerMouseEvent) {
+    if (e.features.length > 0 && hoverId != e.features[0].id) {
+      unhover();
+      // generateId means this'll be a number
+      hoverId = e.features[0].id as number;
+      $map!.setFeatureState({ source, id: hoverId }, { hover: true });
+
+      // Per
```
dabreegster, line 118: This kind of thing is a very annoying aspect of MapLibre's API, but we can work around it cleanly here

web/src/osm2streets-svelte/Layout.svelte
```diff
@@ -0,0 +1,23 @@
...
+    padding: 8px;
+  }
+
+  .left {
+    width: 25%;
+    height: 100vh;
+    overflow: scroll;
+  }
+  .main {
+    /* TODO 75% wraps?! */
```
dabreegster, line 20: Even 73% is broken on my other laptop with a 2x display factor. There's some basic CSS stuff I'm still not understanding. :\

web/src/osm2streets-svelte/import/ImportControls.svelte
```diff
@@ -0,0 +1,164 @@
...
+
+  function loading(e: CustomEvent<string>) {
+    imported = { kind: "loading", msg: e.detail };
+  }
+</script>
+
+<OverpassSelector
+  bind:this={overpassSelector}
+  map={$map}
+  on:loading={loading}
```
dabreegster, line 133: Using lots of event dispatchers here: https://svelte.dev/examples/component-events

web/src/osm2streets-svelte/import/Osm2streetsSettings.svelte
```diff
@@ -0,0 +1,82 @@
+<script lang="ts">
+  export let settings = {
```
dabreegster, line 2: Note we don't have to do anything to detect changes to the form and propagate. `bind:value` makes it a two-way to this prop here, and then since the caller does `bind:` as well, they can then use reactive statements and listen for changes. https://svelte.dev/tutorial/component-bindings

web/src/osm2streets-svelte/layers/DynamicConnectedRoads.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
+  import type { GeoJSON } from "geojson";
+  import Layer from "../Layer.svelte";
+  import LayerControls from "../LayerControls.svelte";
+  import { hoveredIntersection, network } from "../store";
+
+  let gj: GeoJSON | undefined = undefined;
+  let show = true;
+  $: if ($network && $hoveredIntersection) {
```
dabreegster, line 9: All of the layers are reactive, changing usually when the loaded StreetNetwork does. And in this case, based on hover state

web/src/osm2streets-svelte/osm_input/BuiltInSelector.svelte
```diff
@@ -0,0 +1,75 @@
+<script lang="ts">
```
dabreegster, line 1: The two `osm_input` components don't have any concept of `StreetNetwork` or osm2streets. I have some experiments where I want to process OSM data completely differently, but still conveniently extract from overpass or use something built-in

web/src/osm2streets-svelte/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+  // When adding a new layer somewhere, force the programmer to decide where it
+  // should be z-ordered.
+  if (!found) {
+    throw new Error(`Layer ID ${layer} not defined in layerZorder`);
+  }
+  // If beforeId isn't set, we'll add the layer on top of everything else.
+  return beforeId;
+}
+
+// Later entries are drawn on top
```
dabreegster, line 90: This will need rethinking later, if this stays part of the common library meant to be used by multiple apps

web/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+     * Typecheck JS in `.svelte` and `.js` files by default.
+     * Disable checkJs if you'd like to use dynamic types in JS.
+     * Note that setting allowJs false does not prevent the use
+     * of JS in `.svelte` files.
+     */
+    "allowJs": true,
+    "checkJs": true,
+    "isolatedModules": true,
+    "types": ["vite/client"],
+    "strict": false,
```
dabreegster, line 20: I'd like to turn this on eventually, but the tail is very long for fixing every problem...

dabreegster (commented):
#220 for all the later improvements

web/src/osm2streets-svelte/Layout.svelte
yongrenjie: It's to do with the [`box-sizing` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing): by default, when you write `width: 73%` this refers to how wide the _contents_ of the div are. The 8px padding is then added on top of that, so that the entire div is actually 73% + 8px wide (very counterintuitive...)

The solution is to add `box-sizing: border-box` to both `div.left` and `div.main`. Or you can go one step further and [apply this globally]((https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/)), so that you don't run into this problem elsewhere. That's what [I'm doing now](https://github.com/alan-turing-institute/demoland-web/blob/main/src/main.css), at least. 😄 

```css
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```

BudgieInWA:
After only reading through the PR description, I am happy to see this, and happy for it to merge! I'm going to keep this PR around to look back at later.

web/src/osm2streets-svelte/Layout.svelte
dabreegster: Ahh that's indeed an odd default (legacy choice, if I had to guess). Thanks for the tip + explanation! I've fixed just in Layout for now, because I'm not exactly sure about osm2streets-svelte (intended as an NPM library) overriding a global style, or if every app using it should generally do that.

dabreegster:
Merging, then will fix up GH actions. (Hopefully in one commit, but if it turns into whack-a-mole with bugs, I'll switch to a tmp branch)

PR #226, opened 2023-11-02
Add Support for .osm.pbf Input 

### LINKED ISSUE #225: Handle pbf input too
Today we support XML input: https://github.com/a-b-street/osm2streets/blob/main/streets_reader/src/osm_reader/reader.rs

I've been using the `osmpbf` crate in other projects, like https://github.com/dabreegster/od2net/blob/main/od2net/src/network/create_from_osm.rs. We should support this in osm2streets too -- pbf files are so much smaller than xml, and can be faster to parse.

sqrtM:
## Summary

This pull request introduces support for the PBF input format.

## Changes Made

- Added support for reading PBF data.
- Added `osmpbf` dependency

## Issue

Closes #225

## Testing

Added `virgin_islands` folder to the test suite. It is large enough to show the capability and speed of the PBF format over XML, while not being so large that it becomes inhibitive for the developer (or Github size limits).

## Screenshots

image
Main island visualized on geojson.io using the generated geometry.

## Additional Notes

There are a couple of considerations that are important to take note of here. If any of them are not to your liking, just let me know, and I'll fix them right up! :

- PBF is **NOT** supported in WASM as of this pull request. This should be handled in a seperate issue if we wish to implement it.
- When deciding the bounding box for the pbf input, the priority is as follows:
    1. Check for a bounding box in the blob header
    2. If a bounding box does not exist in the header, use the data from boundaries.json
    3. If boundaries.json does not exist, scrape bounds from nodes directly.
    4. If scrape fails, panic.
- Similarly, this functionality is making use of a `BlobReader` rather than the `par_map_reduce` function. This was done to get access to the more accurate bounding box of the header without needing the user to supply one or scrape.
- And finally, just to be clear, the "additions" number is so large because of the large files in the `virgin_islands` directory.

-----

Thank you very much for your time and have a nice day!


dabreegster:
Thank you so much for the PR! Just to let you know, I'm rushing to finish some urgent work before State of the Map next weekend. I don't think I'll be able to look at this til the 13th. Sorry for the delay :\

sqrtM:
> Thank you so much for the PR! Just to let you know, I'm rushing to finish some urgent work before State of the Map next weekend. I don't think I'll be able to look at this til the 13th. Sorry for the delay :\

oh, man, no worries. get to it when you get to it and I'll be around to clean up after the review. not a problem in the least. have a great week!

streets_reader/src/osm_reader/reader.rs
```diff
@@ -19,8 +21,154 @@ use super::{Document, Node, Relation, Way};
...
+                                }
+                            }
+                            PbfElement::Relation(relation) => {
+                                let mut tags = Tags::new(BTreeMap::new());
+                                for (k, v) in relation.tags() {
+                                    tags.insert(k, v);
+                                }
+                                let id = RelationID(relation.id());
+                                if doc.relations.contains_key(&id) {
+                                    //bail!("Duplicate {}, your .osm is corrupt", id);
```
dabreegster, line 123: Why commented out; were you seeing this happen?

tests/src/lib.rs
```diff
@@ -21,40 +24,57 @@ mod tests {
...
-        )?;
+        let clip_pts = if Path::new(format!("{path}/boundary.json").as_str()).exists() {
+            Some(LonLat::read_geojson_polygon(&format!(
+                "{path}/boundary.json"
+            ))?)
+        } else {
+            None
+        };
+
+        let (mut street_network, _) = match Path::new(format!("{path}/input.osm").as_str()).exists() {
```
dabreegster, line 35: An `if` seems simpler than a match here

tests/src/lib.rs
```diff
@@ -21,40 +24,57 @@ mod tests {
...
-        let clip_pts = Some(LonLat::read_geojson_polygon(&format!(
-            "{path}/boundary.json"
-        ))?);
-        let (mut street_network, _) = streets_reader::osm_to_street_network(
-            &std::fs::read_to_string(format!("{path}/input.osm"))?,
-            clip_pts,
-            MapConfig::default(),
-            &mut timer,
-        )?;
+        let clip_pts = if Path::new(format!("{path}/boundary.json").as_str()).exists() {
```
dabreegster, line 27: I like that we're testing this new logic, but the new test has a boundary file. Should we remove it?

dabreegster (changes requested):
So sorry it took ages to get around to this!

- [ ] Could you run `cargo fmt` please? There are lots of spurious diffs; I'm guessing your text editor does something odd
- [ ] The new virgin_islands test is 44MB total in the git repo. That makes cloning the repo a bit painful on a slower connection. Could you use `osmium` or similar and clip down to a smaller area, so that we stick around the 1-3ish MB of other tests? It is definitely nice to see it run properly on a large area, but that's not a test we need to commit

sqrtM:
great! thanks a bunch for the review ! :) 

gonna be a BIT busy with the day job this week, but this should be done sooner than later!

be back soon with the changes + rebase. have a nice rest of the week!

tests/src/lib.rs
sqrtM: my vote would be to keep it. it means that all new pbf tests can optionally include a bounding box IF they want to, but it won't ever be required, since we can read the pbf header. that feels more flexible to me, but it's of course your call !

tests/src/lib.rs
sqrtM: good catch! switched it to `let (mut street_network, _) = if Path::new(format!("{path}/input.osm").as_str()).exists() ` in the new commits.

sqrtM:
alright -- as for your current suggestions, i think everything should be implemented! feel free to have another glance over everything at your leisure. 
- ran cargo fmt
- shrunk the pbf test down to just a few blocks
- removed helper function 
- fixed my silly commenting-out of the bail! (sorry about missing that -- have a look at how I worked around that in the new version. it doesn't bail, but it does print an error and return unit value. that seemed like the most sensible solution. let me know if you'd prefer something else!)
- switched the match statement to an if

let me know if there's anything else you'd like or any other suggestions you may have. thanks!!
(PS sorry for accidentally closing the pr. i'm honestly not entirely sure what caused that. but everything should be good!)

sqrtM:
Happy to help! let me know if there's anything else i can do for you. thanks!

Issue #232, opened 2024-01-19
Try switching to Muv


dabreegster:
https://gitlab.com/LeLuxNet/Muv/ is a new library that's doing the "osm2lanes" job of transforming OSM tags to lane data much better than the code currently used here or in the older https://github.com/a-b-street/osm2lanes experiment. We should try cutting over to it.

It could likely help with a number of bugs: #231, #230, #229, #223, #222, #91, #89.

It's a bigger project to try this cutover, but still somewhat self-contained. If anyone's interested in trying, please do!

tordans:
@dabreegster do know of a place to learn more about Muv that that link? It does not give a lot of context or details :).

dabreegster:
@GinnyTheCat is the author

ginnyTheCat:
Hi,
I built a small example integrating muv-osm into osm2streets [here](https://github.com/ginnyTheCat/osm2streets/tree/muv) with a live example being deployed on https://muv.lelux.net/osm2streets which helps a lot in checking how streets parsed compare to the osm2lanes variant. Another way to check the muv-osm output on a more technical level is running `cargo r --example lanes -F lanes` in the `muv-osm` directory, inputting the tags and getting a lane-by-lane output.

Looking at the issues above #230, #222 and #89 work in muv already. Both way directions as in #91 are supported on the muv side, but have no mapping to osm2streets yet.

I'm working on documenting muv more but [at the example](https://gitlab.com/LeLuxNet/Muv/-/blob/main/muv-osm/examples/lanes.rs?ref_type=heads) or the [muv-osm to osm2lanes](https://github.com/ginnyTheCat/osm2streets/blob/muv/osm2lanes/src/algorithm.rs) source code should help a bit. If there are any more questions I'm more than happy to answer them here or wherever you prefer.

robhubi:
An excellent site for test cases: https://wiki.openstreetmap.org/wiki/Berlin/Verkehrswende/Radwege#Radfahrstreifen_in_Mittellage_.2F_Fahrradschleuse_.2F_Fahrradweichen

ginnyTheCat:
Thanks a lot @robhubi.

[...]

Issue #230, opened 2024-01-18
Lanes for bus and bicycles not shown up


robhubi:
**Real situation:**
![grafik](https://github.com/a-b-street/osm2streets/assets/38103976/19b8d4b2-1aa6-4d2a-95c6-f3483293f358)

Tags of [Way 194755333](https://www.openstreetmap.org/way/194755333) :
```
lanes	4
lanes:forward	3
lanes:backward	1
	
cycleway:right	share_busway
cycleway:right:lane	advisory
cycleway:left	lane
cycleway:left:lane	advisory
	
turn:lanes:forward	left|through;right||
	
vehicle:lanes:forward	yes|yes|no|no
bus:lanes:forward	yes|yes|designated|designated
bicycle:lanes:forward	yes|yes|yes|designated
	
vehicle:lanes:backward	yes|yes
bicycle:lanes:backward	yes|designated
	
sidewalk:both	separate
foot	use_sidepath
	
highway	primary
lit	yes
maxspeed	50
name	Sankt-Peter-Hauptstraße
oneway	no
ref	B67a
smoothness	excellent
surface	asphalt
```
**A/B Street (V0.3.49)**
Turn arrows missing, no bus lane, no cycle lane on the right
![grafik](https://github.com/a-b-street/osm2streets/assets/38103976/e17c8ffc-af8d-4911-9362-5c2465c4083e)

ginnyTheCat:
This is fixed on the osm2streets side as far as I can tell.
![imagen](https://github.com/a-b-street/osm2streets/assets/35173023/f0537f4f-f627-48e9-9685-f02b0b07ec93)
I currently can't compile a/b street, but that seems to be caused by the osm2lanes module being moved out of osm2streets, causing the imports to fail or something similar.

dabreegster:
I'm not planning to update A/B Street anytime soon; any development and validation should be happening here for now. Thanks for fixing this with Muv!

Issue #222, opened 2023-08-05
Use lane width tags from OSM


dabreegster:
`sidewalk:{left,right}:width` as one example, probably others. Add tests for these.

dabreegster:
More context here: the code in https://github.com/a-b-street/osm2streets/tree/main/osm2lanes/src transforms OSM way tags and describes the lanes along a road from left-to-right. Right now it uses hardcoded lane width values from https://github.com/a-b-street/osm2streets/blob/ddc26753b36afe49121b0fe6b523a3293f9281ee/osm2lanes/src/lib.rs#L279. Sometimes lane widths are tagged in OSM, so we should use them when they are. There's a few different width tags that we could use; one simple one is `sidewalk:left:width` (https://wiki.openstreetmap.org/wiki/Sidewalks#Sidewalk_as_refinement_to_a_highway).

iakev:
Hi, @dabreegster I would like to take up this issue. As previously highlighted, I would like to focus on the Rust part, however if you could share some links that I can use to get up to speed on OSM domain knowledge and guidance I would highly appreciate it.

dabreegster:
Hi! For some OSM tags related to width, see https://wiki.openstreetmap.org/wiki/Key:width#Width_of_streets. I would start with the cycleway, sidewalk, and parking lane cases (for when those are part of the main road, and not separate objects). You can use Overpass Turbo and TagInfo to find real-world examples in OSM, like https://taginfo.openstreetmap.org/keys/cycleway:left:width#overview and https://overpass-turbo.eu/s/1Bqk (click Wizard to edit the query). Those can be used to start new unit tests in https://github.com/a-b-street/osm2streets/blob/main/osm2lanes/src/tests.rs. Those tests currently ignore width in the output; I would vote for starting new tests similar to the old, that specifically check for width in some of the output lanes (but not all -- we probably won't know for most of them)

The logic changes would happen in https://github.com/a-b-street/osm2streets/blob/main/osm2lanes/src/algorithm.rs. There's an `apply_width` function that happens after the main pass. That might be a good place to apply this logic -- you'll already have the lanes left-to-right, so if you see something like a cycleway on the right, you could search in one direction for it. Or maybe it's easier to parse these lane overrides when that lane is first found -- not sure.

Let me know if you get stuck with anything, whether it's Rust, OSM, or otherwise. Thanks for helping out on this!

Issue #89, opened 2022-09-15
access=no winds up as construction lane


dabreegster:
See https://www.openstreetmap.org/way/745724293 in `leeds_cycleway` for one example, but I think I've spotted others.

In this one case, `psv=designated` should map to a bus lane. We only look for `psv=yes` today.

dabreegster:
More cases in Firenze like https://www.openstreetmap.org/way/24579090

BudgieInWA:
From my reading of the types that are there, constructed is taken to mean no access, so I'm not surprised by that behaviour :P

psv=* should be understood at an early stage in osm2lanes, I think. First, parsing all the different values into a scheme, then combing all the access tags into a unified understanding. That gets simplified back down in osm2streets to "bus (+taxi +bike) lane" for the rendering

dabreegster:
That sounds like a good plan going forward. I'm starting to return to earlier questions about classifying lane types. In London, there are "bus lanes" that also allow cyclists and taxis. https://www.openstreetmap.org/way/24579090 in Firenze is basically a European living street -- pedestrians rule, motor vehicles are only OK with a permit and don't go fast, and thought it says no bicycles, based on my experience there a few weeks back, they happen but aren't the prioritized mode.

Also in the PR I just sent, I'm still hacking on bits of behavior to the "classic" osm2lanes algorithm: https://github.com/a-b-street/osm2streets/commit/45732ea4c623ca99bb5bfd1f8f954ac8753c9588
I decided to not block progress here on the cutover, but I do want to focus on the cutover soon too

BudgieInWA:
> In London, there are "bus lanes" that also allow cyclists and taxis. [openstreetmap.org/way/24579090](https://www.openstreetmap.org/way/24579090) in Firenze is basically a European living street -- pedestrians rule, motor vehicles are only OK with a permit and don't go fast, and thought it says no bicycles, based on my experience there a few weeks back, they happen but aren't the prioritized mode.

I think it would be cool to model all of these particular cases in terms of the "designated" and "yes"/"no" legal use ideas, because then we're actually using the OSM scheme to describe the situation, as it aims to do.

PR #233, opened 2024-02-25
Replace osm2lanes algorithm with muv_osm

### LINKED ISSUE #231: Design of two-way streets with priority (= only one lane)
It seems that two-way streets are rendered as 2 separate lanes, even when there is a priority specified (= when there is actually only one lane, with a priority for backward or forward trafic) 

See exemple :

https://www.openstreetmap.org/way/406015980

image

Trafic coming from here has to give way https://panoramax.ign.fr/#focus=pic&map=19/48.6062/7.751619&pic=f3634a35-8dc3-4627-bc11-e142cf2937fd&speed=250&xyz=38.80/6.76/14

Trafic coming from here has priority https://panoramax.ign.fr/#focus=pic&map=18.98/48.606289/7.751869&pic=c8a0bdd2-ab95-403f-8e57-2a83f608e90f&speed=250&xyz=260.81/-2.48/47

ginnyTheCat:
Start replacing the osm2lanes logic with [muv_osm](https://gitlab.com/LeLuxNet/Muv) as planned in #232. There is still quite some work to do, both on the muv side and on this translation layer. Which osm2lanes test pass should give a good indication for now.

This will replace the currently quite messy [demo](https://github.com/ginnyTheCat/osm2streets/tree/muv) currently live on https://muv.lelux.net/osm2streets-demo, while this branch is live on https://muv.lelux.net/osm2streets.

As this branch currently stands it resolves #231, #230, #222, #89 (only winds up as construction lane as a last resort; new `LaneType`?)


osm2lanes/Cargo.toml
```diff
@@ -8,6 +8,7 @@ abstutil = { git = "https://github.com/a-b-street/abstreet" }
 anyhow = { workspace = true }
 enumset = { version = "1.0.12", features=["serde"] }
 geom = { workspace = true }
+muv-osm = { version = "0.1.0", path = "../../muv8/muv-osm", features = ["lanes", "serde"] }
```
dabreegster, line 11: This'll need to be `git = "https://gitlab.com/LeLuxNet/Muv/"` eventually, until things are on crates.io. (I publish very few things to crates.io since the API is still in such constant flux)

osm2lanes/src/placement.rs
```diff
@@ -87,7 +87,7 @@ impl Placement {
     }
 }
 
-#[cfg(tests)]
+#[cfg(test)]
```
dabreegster, line 90: Eep, good catch!

osm2lanes/src/lib.rs
```diff
@@ -271,6 +272,8 @@ pub struct LaneSpec {
     /// (though local rules might still dictate restrictions).
     /// Turns for specific vehicle types (`turn:bus:lanes` and such) are not yet captured.
     pub allowed_turns: EnumSet<TurnDirection>,
+
+    pub lane: Option<Lane>,
```
dabreegster, line 276: I like this as a way to plumb the more detailed representation along. `LaneSpec` here probably needs lots of work (direction and lane type particularly), but new features there have to be understood by downstream consumers to be useful -- at minimum, the street explorer. Having the more detailed `Lane` could be a way to update those downstream consumers a bit gradually.

dabreegster (commented):
This is really cool! I've read through the code and it makes cursory sense to me. I still don't have much time to dig into this, but please let me know what I can do to help.

osm2lanes/Cargo.toml
ginnyTheCat: Exactly, that was my plan. It's just for the time being as I develop it locally. Not sure if cargo has something similar to Go's `replace` feature.

osm2lanes/src/placement.rs
ginnyTheCat: Haha, not even my catch. Just clippy complaining.

osm2lanes/src/lib.rs
ginnyTheCat: >direction and lane type particularly

I agree. I just didn't touch them for now as you mentioned in another issue that direction especially changing would break a lot of the road placement and routing code, so i left them as is for now.

ginnyTheCat:
We are now down to 3 failing osm2lanes tests.
```
failures:

---- tests::test_osm_to_specs stdout ----
For input (example from https://www.openstreetmap.org/way/8591383):
    lanes=1
    oneway=yes
    sidewalk=both
    cycleway:left=track
    oneway:bicycle=no
    highway=residential
Got:
    sbds
    v^^^
Expected:
    sbbds
    vv^^^

For input (example from https://www.openstreetmap.org/way/369623526):
    lanes=1
    oneway=yes
    sidewalk=both
    parking:lane:right=diagonal
    cycleway:left=opposite_track
    oneway:bicycle=no
    highway=residential
Got:
    sbdps
    v^^^^
Expected:
    sbbdps
    vv^^^^

For input (example from https://www.openstreetmap.org/way/539534598):
    highway=cycleway
    oneway=no
    segregated=no
Got:
    FF
    ^v
Expected:
    F
    ^

thread 'tests::test_osm_to_specs' panicked at osm2lanes/src/tests.rs:304:5:
3/25 spec tests failing
```

The first two are because of `oneway:bicycle` not being applied to cycleways as well. I saw the OSM links but am not sure if this is common or documented somewhere, as it doesn't seem to be quite common in my region. Maybe you have some information about those cases? That would be quite helpful.

dabreegster:
For those first two cases, https://wiki.openstreetmap.org/wiki/Tag:cycleway:left=track?uselang=en says `cycleway:{left,right}=track` implies riding in both directions by default. So even ignoring `oneway:bicycle=no`, shouldn't there be two cycle lanes next to each other in opposite directions? (in the osm2streets interpretation, where there isn't Direction = Both yet)

ginnyTheCat:
Oh, wow. I've never seen that. https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dtrack says `cycleway=track` is oneway by default. Also in reality I've never seen `cycleway`: (except `highway=cycleway`) being assumed `oneway=no`. I'm not sure how I should handle this now 😅.

dabreegster:
I'm also confused the more I read the two wiki pages. >_< Well, uh, this is exactly the kind of complexity that it's lovely for Muv to hide from downstream ;)

If I had to try and rationalize this, `cycleway=track` implies a cycleway on both sides of the road. Except in Belgium, I've never personally seen that situation being bidirectional on both sides -- there's usually not enough space. So defaulting in that case to one-way for each side makes sense. If someone has explicitly mapped cycleway:left and :right, then... actually, no, it's still strange for the default to assume bidirectional. I don't know enough about the tag history or proposal to comment.

ginnyTheCat:
Yea, I'm gonna leave it as oneway default. Explicit oneways are handled and in case we find any region where this is common (you mentioned Belgium) it's trivial to add a different default for that in muv (should a one line change).

Still I'm unsure whether `oneway:bicycle=no` on the main road should imply `cycleway:oneway=no`.

tordans:
A few quick notes because I am late to the thread :)

* IMO the `cycleway:side=track + oneway:bicycle=no` is very missleading mapping. IMO the `oneway:bicycle=no` applies to the centerline, which would be wrong in the given example. 
    image
* IMO the right tagging for that case would be `cycleway:side=track + cycleway:side:oneway=no`, so the `oneway` is scoped.
* For our work at Radverkehrsatlas.de (https://github.com/FixMyBerlin/atlas-geo/blob/develop/app/process/roads_bikelanes/bikelanes/InferOneway.lua) we assume that `track` is `oneway=no` and `lane` is `oneway=yes` which is how it is used in Germany except for so called "Hochbord Radwege" which usually have an explicit `oneway=no`-Tag. — When researching that I did not find any good documentation on assumed defaults either. 
* The `segregated=no` case is interesting. So in fact there are two lanes, but those lanes are not visible in any infrastructure. I tend to think that should still create the two lanes however without any indicator on separation.

ginnyTheCat:
The track situation is awfully confusing as while the `cycleway` page doesn't say anything about the topic, `cyclway:left=track` (implies `oneway=no`) is completly different from `cyclway=track` and `cycleway:right=track` (implies `oneway`).

I built a small [visualization](https://overpass-turbo.eu/s/1Ijt) in overpass turbo, that really shows how inconsistent the coverage is due to this documentation nightmare. Cycletracks (not `highway=cycleway`, just `cycleway:*=track`) without a oneway value are black, explicitly tagged `oneway=no` blue and explicitly tagged oneway cycletracks green.
imagen
imagen
You can see how the oneway cycletracks and cycleways without oneway tagged are switching between each other which would make me think oneway is the default.

dabreegster:
If the tagging schema / docs themselves are so confusing and there's clear evidence of it being inconsistently mapped in practice, then how about:

1) We raise this on mailing lists / Slack / the wiki discussion page, to make the schema more clear
2) muv implements the simplest / most sane default assumption right now
3) We either delete these osm2streets test cases as ambiguous, or fix how they're tagged upstream in OSM and update the tests.

I don't think figuring this out has to block this PR, because it gives so many other benefits. If there is a clear interpretation, the updated street explorer can help people find and fix these confusingly mapped cases

tordans:
> > We raise this on mailing lists / Slack / the wiki discussion page, to make the schema more clear
> 
> I think that's a great idea
> 

I don't want to discourage you but pushing this through all the channels until the situations is actually better is a big pile of work. My take on the steps would be…

1. gather some input on the current situation via a forum post with some evidence like shown above (it is likely, that there are local differences in how the non documented defaults are perceived)
2. write a proposal draft that makes the case for the issue and the solution (see is_sidepath proposal); that includes possible deprecations, change to wiki pages, maybe even recommendations for editors
3. discuss this proposal in forum and so on – takes a while to "simmer"
4. bring the proposal to a voting
5. use that agreement to go to the id-editor-schema-repo and add deprecations, new fields, new presets so the iD editor nudes people to do the right thing in the future
6. try doing the same for JOSM presets
7. probably start a MapRoulette Project to cleanup the existing mess and gather support to work through that

In very easy cases, only step 1 + updates to the wiki pages is possible.

We went through this with the parking tagging (except 7). It is possible, but a lot of very involved work.

---

What to do "instead"? I suggest to define your own sensible default, document that and go with it. As long it is clear how it works and can be changed in the tags if it's wrong and uses proper tags – which is all the case here – this will help guide the bigger community in the ride direction … over time.

---

Side note: I will likely try this process at some point to get rid of `cycleway:null=track|lane|separate|none` in favor of `cycleway:both=track|lane|separate|none` which is one of those historic issues that no one cleaned up but causes issues when processing and editing all the time…

tordans:
> > muv implements the simplest / most sane default assumption right now

I suggest to assume `oneway=no` on `track`s and oneway=yes on `lane`s. 

And in the example above, to apply the `oneway:bicycle=no` only to the centerline, which will create a second lane in MUV which would be a shared lane wich needs to be very narrow or overlapping. I consider this a tagging error that has to be fixed in OSM, not in the processing. But it is valid tagging eg. here https://www.openstreetmap.org/way/25509293 so it needs to be processed.

ginnyTheCat:
Making that tagging consistent is definitely a long and hard journey. I will collect more information as I go along but probably won't start writing a draft for now.

> We went through this with the parking tagging

The new street parking schema is amazing and i'm very thankful you went through with all those many steps.

> But it is valid tagging eg. here

Exactly. This is one of the things that is an annoying side-effect of catering to mistagings. You have to choose mutually exclusively between handling the mistaging or parsing the valid situations the mistagings are appearing as. If you side with the mistaged version, the other version becomes unrepresentable.

> which will create a second lane

That's how muv used to handle `oneway:bicycle=no`. Now it only creates one lane allowing cars in the forward, and bikes in both directions. The narrow second lane currently only appears when using something like `cycleway=opposite_lane`.

ginnyTheCat:
> where the tag is also discussed sound to me that somewhere sometime recently people considere(d) this tagging to be a good idea.

I'm seeing the Talk page now as well. The messages seem to be from 2018. For me (and what I read in the wiki and encountered when mapping) `cycleway:*=*`  implies `oneway:bicycle=yes` (which is just a permission of access). Without cycleways being there it's just be a oneway that allows cycling in both directions (no infrastructure, only a sign allowing cycling in both directions).

streets_reader/src/extract.rs
ginnyTheCat, line 1: Is there a reason it only renders `railway=light_rail` and `railway=rail` and not `railway=tram`?

ginnyTheCat:
https://github.com/a-b-street/osm2streets/blob/3fa61e302ae66a1279c7d0752fe31770423d8af4/streets_reader/src/extract.rs#L72
Is there a reason it currently only renders railway=light_rail and railway=rail and not railway=tram?

dabreegster:
> Is there a reason it currently only renders railway=light_rail and railway=rail and not railway=tram?

Not really, I think it'd be reasonable to add trams here and treat them the same for the moment. We didn't attempt trams before, because they're usually separate ways that _do_ physically overlap roads. https://github.com/a-b-street/abstreet/issues/141 had some notes, with https://www.openstreetmap.org/#map=19/50.07732/19.90185 as a splendid example of why this'll get hard to get right.

Rendering-wise, it'll be a messy start to have overlapping polygons. Logically one of the road lanes has tram tracks on it too, but sometimes those tracks might veer off and cross a few more lanes.

Reasoning about turns/movements at junctions also gets trickier; we'd have to make sure turn conflicts between the modes of traffic work out.

For a traffic simulation, there'd have to be concept that lanes from two OSM ways occupy the same physical space, so lane changing and queueing and everything else needs to work.

dabreegster:
Another tram test case: https://www.openstreetmap.org/#map=18/47.62134/-122.33649

If/when osm2streets is able to do things like group parallel(ish) ways together, trams would probably work better. But to start, I think it'd be fine to just treat them the same as railways. If they happen to physically overlap roads, that's not that much worse than separate footpaths/cycle lanes doing the same when the width assumptions are wrong.

ginnyTheCat:
Makes sense to wait adding them until the placement algorithm can work with them.

ginnyTheCat:
Muv-osm now supports the `lanes:bus` and `lanes:psv` scheme which now allows even more bus lanes to be shown on the map.

Using osm2streets previous algorithm:
imagen

Using muv-osm: (also notice the cycleway now pointing in the right direction)
imagen

dabreegster:
Super cool to see! What're the blockers to merging at this point -- just that situation with oneway:bicycle? I'd be happy to get something initially merged and update tests as the implementation changes. I've been very negligent on osm2streets work, but all this work here has motivated me to take another look at the Street Explorer code (see if svelte-maplibre would simplify things) and UI (see if the JSON from muv can be displayed nicely), revisit the rendering style (both implementation-wise, as @BudgieInWA had started a new API, and in terms of displaying more detail), and also start to improve the osm2streets internal API, like handling a `Direction::Both` and plumbing that to other apps downstream

ginnyTheCat:
> just that situation with oneway:bicycle

I ignored those tests for now. The only one failing now is the third one, which I will see what I will do with. The ideas you listed sound amazing btw.

@tordans I checked some cities with the visualization and aerial imagery and especially in Berlin most of the `cycleway:{left,right,both}=track` don't have a `cycleway:{left,right,both}:oneway` associated with them. From what I could see in the satellite view it looks like most of those cycleways are oneway though. Especially the dual carriageway below would have two-directional cycleways on both sides of the road then.
screenshot of OSM in Berlin with cycleways highlighted

Maybe I'm misunderstanding something about the default but most two-way cycleways I've seen, were ususally mapped as a seperate `highway=cycleway` next to the main road.

web/package.json
```diff
@@ -7,7 +7,7 @@
     "wasm": "wasm-pack build --dev --target web ../osm2streets-js",
     "wasm-release": "wasm-pack build --release --target web ../osm2streets-js",
     "dev": "vite",
-    "build": "vite build",
+    "build": "npm run wasm-release && vite build",
```
dabreegster, line 10: I just pushed a commit to this branch with a change to GH config due to this, FYI

osm2lanes/src/algorithm.rs
```diff
@@ -1,610 +1,238 @@
...
-    } else if tags.is("sidewalk", "separate") && cfg.inferred_sidewalks {
-        // TODO Need to snap separate sidewalks to ways. Until then, just do this.
-        fwd_side.push(fwd(LaneType::Sidewalk));
-        if !back_side.is_empty() {
-            back_side.push(back(LaneType::Sidewalk));
-        }
-    } else if tags.is("sidewalk", "right") {
-        if cfg.driving_side == DrivingSide::Right {
-            fwd_side.push(fwd(LaneType::Sidewalk));
+        let lane_type = if main || rank.lane_type != LaneType::Footway {
```
dabreegster, line 203: Looks like `main` is used here. I don't understand the logic here, comment please?

osm2lanes/src/algorithm.rs
```diff
@@ -1,610 +1,238 @@
...
-    if cfg.inferred_sidewalks || tags.is(osm::HIGHWAY, "living_street") {
-        if need_fwd_shoulder {
-            fwd_side.push(fwd(LaneType::Shoulder));
-        }
-        if need_back_shoulder {
-            back_side.push(back(LaneType::Shoulder));
-        }
+        return (lane_type, dir, turns);
     }
+    (LaneType::Construction, Direction::Fwd, EnumSet::new())
```
dabreegster, line 222: There's also a lifecycle check elsewhere that assigns construction. In this fall-back case, is construction the only case in OSM, or is this osm2streets not yet modeling some situation? No rank has access, so it might be something like a military/delivery-only access?

ginnyTheCat:
Thanks for the review. I often forget to comment but I will add these to hopefully clear up what the algorithm is doing. I didn't know about the visual tests yet but will definitely give them a try as well.

dabreegster:
Ah, checking the visual tests is useful! Comparing https://a-b-street.github.io/osm2streets/?test=arizona_highways with the local version, I noticed:

- Both lanes of https://www.openstreetmap.org/way/23806634 get classified incorrectly as SharedLeftTurn
- https://www.openstreetmap.org/way/528310266 and nearby gets classified as bus lanes

These are good candidates for more unit test cases in `osm2lanes/src/tests.rs` -- the representation in Muv may be fine, but then the mapping is unexpected. I can look through these cases later this week

osm2lanes/src/algorithm.rs
ginnyTheCat: This currently just imitates the old algorithm which would fall back to `Construction` if the way is not accessable. Maybe it's worth having a separate `LaneType` for those cases.

osm2lanes/src/algorithm.rs
```diff
@@ -1,610 +1,238 @@
...
-            fwd_side.insert(idx + 1, fwd(LaneType::Buffer(buffer)));
+// Specifies the importance of different modes of transport in descending order
+// The raw outline is `train > car > bus > shared use > bicycle > foot`
+const RANKS: [Rank; 11] = [
+    Rank::normal(TMode::Tram, LaneType::LightRail),
+    Rank::normal(TMode::Train, LaneType::LightRail),
+    Rank::designated(TMode::Bus, LaneType::Bus),
+    Rank::normal(TMode::Motorcar, LaneType::Driving),
+    Rank::normal(TMode::Bus, LaneType::Bus),
+    Rank {
```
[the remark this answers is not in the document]
ginnyTheCat: It's the like order of importance of all the LaneTypes.

osm2lanes/src/algorithm.rs
```diff
@@ -1,610 +1,238 @@
...
+    Rank::designated(TMode::Bicycle, LaneType::Biking),
+    Rank::designated(TMode::Foot, LaneType::Footway),
+    Rank::normal(TMode::Bicycle, LaneType::Biking),
+    Rank::normal(TMode::Foot, LaneType::Footway),
+    Rank::normal(TMode::All, LaneType::Shoulder),
+];
+
+fn travel_lane(
+    t: &TravelLane,
+    main: bool,
```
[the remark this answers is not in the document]
ginnyTheCat: `main` is super confusing. It's a terrible name and changed now to something more understandable.

osm2lanes/src/algorithm.rs
```diff
@@ -1,610 +1,238 @@
...
-                if part == "designated" {
-                    back_side[idx].lt = LaneType::Bus;
-                }
-            }
-        }
-    }
+fn access_level_allowed(access: AccessLevel) -> bool {
+    matches!(
+        access,
+        AccessLevel::Designated
```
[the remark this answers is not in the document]
ginnyTheCat: > What're your thoughts?

I'm not entirely sure about this function either. As far as I understood the old algorithm checked `access != No` so this might just be the easiest way to implement this function. Maybe we could add a case for `Private`.

I'm, however, not sure whether it makes more sense to be more permissive (allowing `Destination`, `Permit`, or even `Agricultural`) or whether being more restrictive (only `Yes`, `Designated` and `Permissive`) makes more sense considering the downstream users such as the A/B Street routing engine.

ginnyTheCat:
I fixed the three visual issues you found and added osm2street test cases for the first two. I also added some comments and hope the code is more understandable now.

dabreegster:
Thank you, all the comments make sense!

I looked through all of the e2e tests and added a commit confirming the ones that looked reasonable to me. There are still 9 with changes that seem wrong to me. Some possible issues:

- https://www.openstreetmap.org/way/792024859 All 3 lanes show up bus lanes, but it should just be the rightmost. My understanding of bus:lanes:forward = yes|yes|designated is that buses are allowed in the 2 general lanes, but the rightmost is the designated bus lane
- https://www.openstreetmap.org/way/539534608  This should be a bidirectional cycle lane on the right side. I think this is one of the weird tagging cases, so maybe we accept this for now and figure it out later
- https://www.openstreetmap.org/way/4869147 and others in https://a-b-street.github.io/osm2streets/?test=i5_exit_ramp are showing up as construction, possibly because it's the fall-through case for access
- https://www.openstreetmap.org/way/1024116250 in `perth_peanut_roundabout` seems to have a cycle lane on both sides of a one-way, which I'm pretty sure isn't right...
- Some light rail became construction, like https://www.openstreetmap.org/way/372672863
- https://www.openstreetmap.org/way/428087110 wound up with 4 vehicle lanes total and a cycle lane, but `lanes = 3`
- https://www.openstreetmap.org/way/1067864761 lost all vehicle lanes, see https://a-b-street.github.io/osm2streets/?test=st_georges_cycletrack

Hopefully this isn't an overwhelming list. :\  I can also start to chip away at these if it is. It's feeling close though!

dabreegster:
Edit: I've been using the latest OSM data for some of the tagging, but it's entirely possible the .osm files in `tests` are outdated and have weird tagging. So caveat some of those things I said, maybe

ginnyTheCat:
I've been checking the visual tests as well. And already fixed most of them.

The i5 case is interesting as it's tagged as `oneway=reversible` which blocks travel in both directions as it's missing a notice when it can be driven on. I'm not sure what would be the best way to handle this.

dabreegster:
You're right about the I5 case, https://wiki.openstreetmap.org/wiki/Tag:oneway=reversible is pretty clear about `oneway:conditional` being there too. There actually appears to be a schedule at https://wsdot.wa.gov/travel/roads-bridges/express-lanes. I'll raise with some local mappers and for now, this interpretation of no access here is fine, consistent with the tagging

ginnyTheCat:
Yea. muv is on the safe side with reversible as the wiki for example also suggests not to route via reversible roads if the times are unknown.

ginnyTheCat:
I think the only one remaining now is

> https://www.openstreetmap.org/way/539534608 This should be a bidirectional cycle lane on the right side. I think this is one of the weird tagging cases, so maybe we accept this for now and figure it out later

which is a osm2streets tests as well that is currently ignored. We were unsure whether "pulling" a `oneway:bicycle` from a road, into a cycleway makes sense.

dabreegster:
Just had another quick look, liking the improvements! I'll have more time after work, but still a few oddities:

- On [428087110](https://www.openstreetmap.org/way/428087110), still seeing too many vehicle lanes. I checked `src/seattle_triangle/input.osm` to make sure the data's reasonably up to date, and `lanes=3` in there indeed
- https://www.openstreetmap.org/way/692655614 Segregated cycle lanes should have a footpath (or sidewalk?) on one side, but they're showing up as construction
- The double cycle lane is fixed, but I think it's on the wrong side for RHT by default. Example https://www.openstreetmap.org/way/422264711

ginnyTheCat:
I'm was confused with the first one and still am. Maybe I'm misunderstanding something.
imagen
This is what it looks like on my side. The ones with `lanes=3` have 3 lanes, and the other ones 4.

dabreegster:
If I re-import the OSM data for `seattle_triangle`, it looks fine. Something must be strange with the older data in there. I'll reimport and update now.

Everything else looks fantastic; I think we're basically there! I noticed two small things, but they don't need to block merging:
- I think https://www.openstreetmap.org/way/668984461 should have a bus lane, but it looks like `cycleway=share_busway` maybe overrides it
- https://www.openstreetmap.org/way/207816746 has a bus lane on both sides of a one-way four lane road. I'm not sure how to interpret the "both sides" part of https://wiki.openstreetmap.org/wiki/Tag:busway=lane for one-way roads

ginnyTheCat:
I just noticed a small bug in the muv implementation of segregated.
imagen
The idea is that it's always appends the footway on the driving side (left with LHD, right with RHD) as the cycleway the segregated is on is likely a oneway following a road, so it doesn't sandwich the segregated footway between the cycleway and the footway.
This broke though for LHD and I just noticed because the small cycleway in the top is on the right. Now that it's fixed though
imagen
the footway on the bottom cycleway ofc also switches sides. With the `segregated` tag we can really only guess what side it belongs to. I think adding a `sidewalk=left` tag on the bottom cycleway is more practical than always appending segregated paths to the right even for LHD.

dabreegster:
> I think adding a sidewalk=left tag on the bottom cycleway is more practical than always appending segregated paths to the right even for LHD.

Good catch, and yeah, this makes perfect sense. One of the things osm2streets eventually hopes to solve is figuring out the relationship between the separate cyclway/footwayand the main road, so that it could at least know if they're on the left or right, or reason about the polygon gap between them, or maybe consolidate/merge. If we get to that point, we could supply hints about the direction. But until then, the tagging is kind of ambiguous, and it's nice to see it so clearly, so people can try different tagging.


I think everything here is good to go! If we find more problems, we can just make more smaller changes. I'd like to add you to the project credits; do you want to be identified by your GH username, and do you want a link to something other than https://gitlab.com/LeLuxNet/Muv?

dabreegster (approved):
Absolutely phenomenal work, I am so excited to cutover to this and resume effort on osm2streets

ginnyTheCat:
> I think https://www.openstreetmap.org/way/668984461 should have a bus lane, but it looks like cycleway=share_busway maybe overrides it.

This is a odd one. It talks about 2 lanes, but then has a `psv:lanes=designated` at the end. As muv sees this value first, it disgards all the `:lanes` tags. I will handle this more gracefully in the future, but in most cases there isn't a lot we can do when the `:lanes` count mismatch.

But the main issue is that `Motorvehicles` are set to `permit`. I would either remove `permit`from the list of `AccessLevels` we [deem as allowed](https://github.com/a-b-street/osm2streets/pull/233#discussion_r1522065205), or make it rank lower than `Yes`. I'm not sure what makes more sense though.

> looks like `cycleway=share_busway` maybe overrides it

`cycleway=share_busway` is still a TODO on the muv side and currently just added as a normal cycleway. In the future it would merge into the bus lane though.

> I'm not sure how to interpret the "both sides" part

I think it makes sense to handle it the same way we did with the cycleways. Oneways with bus lanes on both sides could still tag `bus_way:both=lane` to get one on both sides.

> we could supply hints about the direction

That's a super interesting idea as well and has a lot of potential.

> I think everything here is good to go!

After that quick fix I think we're good to go.

> do you want to be identified by your GH username

Yup, that's perfect

> do you want a link to something other than https://gitlab.com/LeLuxNet/Muv?

No, I don't think so. Thanks a lot already.

ginnyTheCat:
> I think everything here is good to go! If we find more problems, we can just make more smaller changes.

I agree. The translation layer is quite robust now so most bugs can just be fixed on the muv side with a simple `cargo update muv-osm` on the `osm2streets` side.

dabreegster:
Alright, merging this finally! Thanks so much for all of the work.

Probably Friday or so, I'll carve out time to update the Street Explorer UI, start exposing more Muv details, etc. If you're interested in reviewing the code or just keeping up with changes, I can describe the work in PRs.

> I will handle this more gracefully in the future, but in most cases there isn't a lot we can do when the :lanes count mismatch.

In situations when the tagging is inconsistent or ambiguous, one idea is to collect a list of warnings. We can find an appropriate way to display them to the user (maybe in the lane editor).

ginnyTheCat:
> If you're interested in reviewing the code or just keeping up with changes, I can describe the work in PRs.

Sounds great. I will definitely keep eyes open for that.

> one idea is to collect a list of warnings

I saw the newer osm2lanes did that and I think that is a good idea.

tordans:
@ginnyTheCat 

> I checked some cities with the visualization and aerial imagery and especially in Berlin most of the `cycleway:{left,right,both}=track` don't have a `cycleway:{left,right,both}:oneway` associated with them. From what I could see in the satellite view it looks like most of those cycleways are oneway though. Especially the dual carriageway below would have two-directional cycleways on both sides of the road then. 
> Maybe I'm misunderstanding something about the default but most two-way cycleways I've seen, were ususally mapped as a seperate `highway=cycleway` next to the main road.

"Even in Berlin" (where we have a quite active OSM + Cycling Community we have a lot of unfinished tagging to cleanup. I consider those cases where `track` is mapped on the centerline without a `oneway` incomplete. But it's just so much work to get good defaults and recommendations inside editors and to create QA tools that get used… (We are working on that with https://radverkehrsatlas.de/ in some parts).
My general take is, to move `track` to separate `highway=path` or `highway=cycleway` which makes it so much easier to map the needed details like `oneway`.

So my take is: It is OK for osm2streets to show the tagging mistake so mappers start fixing them…

(Sorry for the late reply…)

ginnyTheCat:
> "Even in Berlin"

In most cities the visualization I built shows around 70% of cycleways not having any oneway tag (neither `yes` nor `no`) in most cities, while in Berlin it's more like 90%. From my memories of Berlin and most cities in Europe `highway=cycleway` are mostly just cycleways on sidewalks and therefore oneway by default. The bidirections cycleways common in North America are in most cases tagged as a `highway=cycleway` as they are more often seperated by poles, barriers, parking etc.

matkoniecz:
I noticed it in https://github.com/openstreetmap/id-tagging-schema/pull/1179

I would say that `cycleway:left`/`cycleway:both`/`cycleway:right`/`cycleway`  `=track` is a format of fixme and should be replaced by properly mapped geometry.

But `highway=tertiary cycleway=track oneway=yes oneway:bicycle=no` may be  used to map as one object oneway road with two-way cycleway along it

tordans:
> But `highway=tertiary cycleway=track oneway=yes oneway:bicycle=no` may be used to map as one object oneway road with two-way cycleway along it

That tagging is so imprecise that we have to make so many assumptions … 
- assume that `cycleway=track` is actually `cycleway:both=track`
- assumes that `oneway:bicycle=no` in combination with `cw=track` does _not_ mean cycling is allowed in both directions where the vehicles drive … _but_ instead applies to the `track`
- and then we should assume that the track on both sides is free to be used in both directions, yes?

Is this how you would interpret it, @matkoniecz ?

| style | left | center | right |
|--------|--------|--------|--------|
| centerline mapping | - | `highway=tertiary cycleway=track oneway=yes oneway:bicycle=no` | - |
| separate mapping | `highway=cycleway/path oneway=no` | `highway=tertiary cycleway:both=separate oneway=yes` | `highway=cycleway/path oneway=no` | 

Of course, there is also a way to say…
- `cycleway=track` is only applied to the main direction of the road, so in most countries it should be interpreted as `cycleway:right=track`

In this case, the `oneway:bicycle=no` makes a little more sense. But still, the parser needs to behave differently for streets that are just `highway=* + oneway=yes + oneway:bicycle=no (+ optional cycleway(:both)=no)` (which is the only way to allow cycling is allowed in both directions (but vehicles are not) ([example](https://www.openstreetmap.org/way/25509293#map=19/52.46963/13.44015)))

| style | left | center | right |
|--------|--------|--------|--------|
| centerline mapping | - | `highway=tertiary cycleway=track oneway=yes oneway:bicycle=no` | - |
| separate mapping | - | `highway=tertiary cycleway:left=no cycleway:right=separate oneway=yes` | `highway=cycleway/path oneway=no` |

matkoniecz:
> Is this how you would interpret it, @matkoniecz ?

I would also consider `fixme=replace it by properly mapped cycleway as separate geometry, and replace it by cycleway:*=separate` to be an implied tag.

I would not assume that it is `cycleway:both=track` - I would assume that it is two-way track on one, unspecified, side. At least in Poland, in Germany two oneway tracks may more likely.

I had (still unfinished) project of visualising bicycle infrastructure and `=track` was going solely to QA section as serious mapping quality issue, with no attempt to guess what people meant by it.

tordans:
@matkoniecz glad we are on the same page :-).

And to pick up the (deleted) conversation from before, that is also why it was talking about scoping the oneway to the virtual cycleway like so…

| style | left | center | right |
|--------|--------|--------|--------|
| current | - | `highway=tertiary oneway=yes cycleway=track oneway:bicycle=no` | - |
| more clear | - | `highway=tertiary oneway=yes cycleway:left=track cycleway:left:oneway=no` | - |

> I have (still unfinished) project of visualising bicycle infrastructure and =track was going solely to QA section as serious mapping quality issue, with no attempt to guess what people meant by it.

Lets talk about https://github.com/FixMyBerlin/atlas-geo processing at some point. It would be great to see someone from non-germany pick this up and see if the processing does fit the local community styles. In a few month we will have proper docs for the category that we create and some time this year I want to create a German-wide map with some Maproulette projects to improve the tagging to make the processing work.

matkoniecz:
> … and which is why I consider the wiki very non helpful with this oneway tagging.

Well, wiki describes tagging as-is

if new better tagging is needed then tagging mailing list discussion/proposal process is needed, not only just editing wiki page 

(and for new style I would keep also oneway:bicycle=no and have this new extra tagging as optional, for typical use such as routing this detail is not needed at all)

> cycleway:left=track

If someone cares about high level of detail and clarity I would focus on getting rid of any `track` values and replace them by separately mapped geometries and (in this case) `cycleway:left=separate`

matkoniecz:
> Lets talk about https://github.com/FixMyBerlin/atlas-geo processing at some point. It would be great to see someone from non-germany pick this up and see if the processing does fit the local community styles. In a few month we will have proper docs for the category that we create and some time this year I want to create a German-wide map with some Maproulette projects to improve the tagging to make the processing work.

Is it by any chance serving tiles/data for Kraków, Poland? Or having some working setup step-by-step instructions?

tordans:
> > Lets talk about https://github.com/FixMyBerlin/atlas-geo processing at some point. It would be great to see someone from non-germany pick this up and see if the processing does fit the local community styles. In a few month we will have proper docs for the category that we create and some time this year I want to create a German-wide map with some Maproulette projects to improve the tagging to make the processing work.
> 
> Is it by any chance serving tiles/data for Kraków, Poland? Or having some working setup step-by-step instructions?

No, we only take the Germany dataset from Geofabrik and there are no plans to expand the region any time soon. There are no docs for this ATM but it most of it should be easy to use to set it up for other countries. Hence the "lets talk about" – I can tell you more in a screenshare or chat some time. Or maybe I will make it to SOTMEU…

tordans:
Just for clarity…

> > cycleway:left=track
> 
> If someone cares about high level of detail and clarity I would focus on getting rid of any `track` values and replace them by separately mapped geometries and (in this case) `cycleway:left=separate`

The same issue is there with `lane`, I just kept the `track` because we had it as an example before.

matkoniecz:
Then I would say that again, if road in general allows bidirectional bicycle travel then - at least for purposes I know - `oneway:bicycle=no` is good enough.

ginnyTheCat:
> Well, wiki describes tagging as-is

The main problem with the wiki is that `cycleway:left=track` and `cycleway:right=track` are different from the oneway they imply with I doubt people tag like.

PR #236, opened 2024-03-13
Display OSM tags and muv JSON in lane popups, in a rather unergonomic

dabreegster:
way

The result isn't really legible yet, but it's a start. I'm rethinking what's useful to show in this popup, and whether the blank space in the left sidebar could be used instead.

CC @ginnyTheCat

### DIFF
--- a/osm2streets/src/render.rs
+++ b/osm2streets/src/render.rs
@@ -148,6 +148,9 @@ impl StreetNetwork {
                     "osm_way_ids",
                     Value::Array(road.osm_ids.iter().map(|id| id.0.into()).collect()),
                 );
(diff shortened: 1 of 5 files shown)


dabreegster:
Sounds good! We can always do some simple JSON rewriting on this end to display some things more compactly. If this current representation is easier to work with in the code, the serialized form shouldn't take priority

ginnyTheCat:
It shouldn't be to much harder to implement if it works out how I imagine it. The main problem, other than making the JSON longer, is that it doesn't work with with things like `access:conditional=no @ (22:00-6:00)` + `bicycle=yes` + `foot=yes`.

PR #239, opened 2024-03-15
Better popups

dabreegster:
Screenshot from 2024-03-15 12-51-44
Screenshot from 2024-03-15 12-52-08
Screenshot from 2024-03-15 12-52-01
@ginnyTheCat, the more compact JSON is looking nice, as are the est_width and segregated fixes


ginnyTheCat:
The Popup looks super good. I like it a lot 👍

Issue #240, opened 2024-03-15
Improve rendering detail in Street Explorer


dabreegster:
The A/B Street renderer includes more detail than osm2streets. Some things will be hard to achieve in MapLibre, like fixed map-size bike/bus symbols, but most should be doable.

- [ ] Parking lane hatches (perpendicular only right now, but an opportunity to support diagonal and others)
- [ ] Change intersection fill color based on the incident lane types
- [ ] sidewalk perpendicular lines
- [ ] the curb between sidewalk and road
- [ ] shared use path dashed outlines
- [ ] bike and bus icons (hard)

PR #242, opened 2024-03-15
Add kerbs

ginnyTheCat:
Adds kerb positions generated by muv-osm to osm2streets as discussed in #232 and tracked by #240. This provides a higher level of detail as it allows one to easily distinguish `cycleway=lane` from `cycleway=track` on the map.

A cycleway moving from the roadway to the sidewalk

It behaved quite unexpected for me for `highway=cycleway` + `sidewalk=*` but this is what the [wiki suggests](https://wiki.openstreetmap.org/wiki/Key:sidewalk#Usage_on_cycleways). In case this is unwanted we can skip the kerb in this case in every country but the Netherlands.


dabreegster:
Thank you! I'll pick this up tomorrow. The rendering of the curb is a bit buggy; `RenderLanePolygons.svelte` probably needs a new case.

Relatedly, `BufferType` here can probably have an overhaul. I think https://wiki.openstreetmap.org/wiki/Proposal:Separation has caught on. I might have a go at some simple rendering for all of these cases. Downstream use cases might look at the quality of the separation of a cycleway from traffic.

dabreegster (approved):
Made some UI tweaks so it shows up, though really faintly. Thanks!

PR #244, opened 2024-03-17
Add support to switch times of day in the UI

ginnyTheCat:
Not sure whether this is in the scope of osm2streets. In tried it some days ago as a demo and it worked quite well, so before it rots in my local braches you can at least have a look at it.

A [pedestrian tunnel below a train station](https://www.openstreetmap.org/way/29049225) closed at night:
tunnel marked as no access

Open during the day:
tunnel marked as cycleway

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -2,6 +2,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Once;
 
 use abstutil::{Tags, Timer};
+use chrono::NaiveDateTime;
 use geom::{Distance, LonLat, PolyLine, Polygon};
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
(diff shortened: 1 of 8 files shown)


osm2streets-js/src/lib.rs
```diff
@@ -59,6 +61,7 @@ impl JsStreetNetwork {
         let mut cfg = MapConfig::default();
         cfg.inferred_sidewalks = input.inferred_sidewalks;
         cfg.inferred_kerbs = input.inferred_kerbs;
+        cfg.date_time = input.date_time;
```
dabreegster, line 64: Neat that the default JSON parsing handles strings from the browser

dabreegster (approved):
Thanks for the PR! This could be useful to some downstream consumer. I remember years ago, there was someone interested in simulating turn restrictions during peak hours. It would take some work for the downstream app to know when to recalculate the network, but no reason not to add in some support now.

ginnyTheCat:
> It would take some work for the downstream app to know when to recalculate the network

On a pure lane by lane level that should be quite easy. `opening-hours` (as does `opening-hours.js`) provides us with an iterator over all changes that are happening from a point in time onwards. I was going to use that to build a similar iterator for the whole `Conditional`s allowing you to have different road networks for different time windows.

Issue #246, opened 2024-03-22
Crossings


dabreegster:
- [ ] Render traffic islands
- [ ] Handle when an intersection gets merged
- [ ] Distinguish signalized from marked
- [ ] Handle Roads that're crossings

tordans:
How to you plan on handling the different taggings from the way segment `footway|cycleway|path=crossing` vs. the taggings from the node `highway=crossing`?

- example way https://www.openstreetmap.org/way/940414779
- example node https://www.openstreetmap.org/node/6320136951

Rapid recently added code that helps to sync some of the tagging of those two representations. I still have to try this out, but it sounds promising because right now the crossing system is pretty messy, unfortunately. I thought that usually the nodes are present first and the ways are only added later, which would make the nodes the winner when tags conflict. Hover I heard that in some regions there are only ways, not notes…

PS: During the last iD Editor community meetup we discussed how to improve the presets for crossings which will improve the situation somewhat. But I still have do document what we talked about in that meetup…

dabreegster:
That example looks like a good one to iterate on: https://a-b-street.github.io/osm2streets/#19.29/52.4737549/13.4406967
![image](https://github.com/a-b-street/osm2streets/assets/1664407/86de673a-6dec-45da-a4a5-b1001e96c0ad)
Right now it's mostly looking for tagging on the node. I'll take a shot at interpreting the ways now.

I honestly have a poor understanding of the tagging schema; I've just been working in example places that I know on the ground and that seem to be tagged reasonably

dabreegster:
Ah, both the missing crosswalks here are because two OSM nodes get consolidated by osm2streets into one polygon. Then the crossing junction connects to three walkable segments, and so it's unclear how to draw the straight line of the crossing. That straight line is erased in the process of building the intersection polygon, but usually it's recovered by looking for the two connecting ends

dabreegster:
![image](https://github.com/a-b-street/osm2streets/assets/1664407/3ee4baf6-56d9-493a-bb8c-560a3269eac0)
This shows the problem more clearly maybe. When there are 3 walkable edges, it's picking two arbitrarily to draw the straight line of the crossing. The left side looks good here, but the right picks the wrong pair.

Quick hack would be to find the pair of points farthest from each other. Proper approach would be to preserve the original crossing geometry when building the intersection.

PR #247, opened 2024-03-22
Start handling highway=crossing nodes

dabreegster:
image

This PR starts a fresh attempt to handle pedestrian crossings in osm2streets. Previously, I think @BudgieInWA and I have thought about these as movements/turns across an intersection, having some influence on stop lines, etc. This new approach is much simpler, exploiting the fact that we already have separate Intersection objects that exactly correspond to a `highway=crossing` node such as https://www.openstreetmap.org/node/2819230451. So all we have to do is plumb along some per-Intersection info and render it as markings.

This approach makes sense only in the context of using separately mapped sidewalks (which is kind of the way things are going anyway). Looking at the image above, the idea of a larger "junction" that groups the big Intersection and all of the things around it starts to look a bit more clear. But going back to an old idea with Ben, let's first get the "one-to-one" OSM representation to be pretty good, then we can think about higher-level "views" on top of it.

This is a simple first step, some next steps are in #246. Comments welcome from anyone, particularly @andrewphilipsmith, @ginnyTheCat, @hungerburg (dredging up https://github.com/a-b-street/abstreet/issues/795 -- no more inferred anything!), @SupaplexOSM. (I might merge quickly so osm2streets.org is updated for convenience, but happy for feedback whenever)

### DIFF
--- a/osm2streets/src/render/intersection_markings.rs
+++ b/osm2streets/src/render/intersection_markings.rs
@@ -1,10 +1,10 @@
 use anyhow::Result;
 use geojson::Feature;
-use geom::{Polygon, Ring};
+use geom::{Distance, Line, PolyLine, Polygon, Ring};
 
 use super::{serialize_features, Filter};
 use crate::road::RoadEdge;
-use crate::{Intersection, LaneType, StreetNetwork};
+use crate::{CrossingKind, Intersection, LaneType, Road, StreetNetwork};
 
 impl StreetNetwork {
     pub fn to_intersection_markings_geojson(&self, filter: &Filter) -> Result<String> {
@@ -14,6 +14,25 @@ impl StreetNetwork {
                 let mut f = Feature::from(polygon.to_geojson(Some(&self.gps_bounds)));
                 f.set_property("type", "sidewalk corner");
                 features.push(f);
+            }
+
+            if let Some(ref crossing) = intersection.crossing {
+                match crossing.kind {
+                    CrossingKind::Signalized | CrossingKind::Marked => {
+                        for polygon in draw_zebra_crossing(self, intersection) {
+                            let mut f = Feature::from(polygon.to_geojson(Some(&self.gps_bounds)));
+                            f.set_property("type", "marked crossing line");
+                            features.push(f);
+                        }
+                    }
+                    CrossingKind::Unmarked => {
+                        for polygon in draw_unmarked_crossing(self, intersection) {
+                            let mut f = Feature::from(polygon.to_geojson(Some(&self.gps_bounds)));
+                            f.set_property("type", "unmarked crossing outline");
+                            features.push(f);
+                        }
+                    }
+                }
             }
         }
         serialize_features(features)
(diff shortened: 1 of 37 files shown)


osm2streets/src/render/intersection_markings.rs
```diff
@@ -15,6 +15,25 @@ impl StreetNetwork {
...
+                match crossing.kind {
+                    CrossingKind::Signalized | CrossingKind::Marked => {
+                        for polygon in draw_zebra_crossing(self, intersection) {
+                            let mut f = Feature::from(polygon.to_geojson(Some(&self.gps_bounds)));
+                            f.set_property("type", "marked crossing line");
+                            features.push(f);
+                        }
+                    }
+                    CrossingKind::Unmarked => {
+                        for polygon in draw_unmarked_crossing(self, intersection) {
```
dabreegster, line 29: Ideas for rendering styles welcome, especially for unmarked crossings -- by definition it's quite unclear how to indicate them! Of course these are going to be region-dependent; I'm starting with something simple and we can add configurability later (along with region-specific lane colors, default widths)

streets_reader/src/extract.rs
```diff
@@ -53,10 +52,21 @@ impl OsmExtract {
...
         }
 
-        // TODO Maybe restricting to traffic_signals is too much. But we definitely don't want to
-        // use crossing=unmarked to infer stop lines
-        if node.tags.is("highway", "crossing") && node.tags.is("crossing", "traffic_signals") {
-            self.signalized_crossings.push(node.pt.to_hashable());
+        if node.tags.is("highway", "crossing") || node.tags.is("railway", "crossing") {
+            let kind = match node.tags.get("crossing").map(|x| x.as_str()) {
+                Some("traffic_signals") => CrossingKind::Signalized,
+                Some("uncontrolled" | "marked") => CrossingKind::Marked,
```
dabreegster, line 58: Based on my understanding of https://wiki.openstreetmap.org/wiki/Tag:highway=crossing?uselang=en

ginnyTheCat:
Looks amazing. Maybe in the future we could render those details for crossing nodes as well, similar to how Mapbox Streets renders them.

streets_reader/src/extract.rs
1ec5: Kind of. The [history around crossing classification](https://wiki.openstreetmap.org/wiki/Crossings#Street_crossings) is quite muddled and quickly devolves into a blame game. Some mappers strongly believe `uncontrolled` should mean marked/signposted/unsignalized, while others point out that `uncontrolled` in some regions such as the UK and U.S. was historically largely used on unmarked crossings due to plain language in both dialects. I think it’s fair to say that the former faction is winning, but only because the latter has moved on to `crossing:markings=*` and `crossing:signals=*`. At this point, a new renderer or router would be better off supporting `crossing:markings=*` and `crossing:signals=*`, which are less ambiguous in practice.

osm2streets/src/render/intersection_markings.rs
1ec5: Probably just a void with no road markings would suffice, unless there are regions where it’s normal for a centerline to continue uninterrupted past an unmarked crossing.

dabreegster: Tough to do today, because osm2streets doesn't know how to continue any road markings into the "intersection polygon" area that gets calculated. That's something @BudgieInWA was working on remedying by detecting "intersections" that're really just merges / forks. I'll keep it in mind as an idea, thank you!

PR #249, opened 2024-03-22
Get rid of experimental code and graphviz. #162

### LINKED ISSUE #162: Remove experimental, keep graphviz
Remove graphviz layer from street explorer, remove the experimental crate, but keep the .dot generation for tests

dabreegster:
Closes #162 

We haven't used any of the graphviz or experimental code in a long time. Some of the enums and ideas from it have made their way into the main code. And lots has changed after the Muv refactor. I'd like to clean up the repo a bit. If we ever want to restore anything here, it should be easy enough from git history.

I'll wait a day or two for feedback and merge


dabreegster:
image
Some graphviz output looks like this. With the current coloring based on intersection kind and nicer popups, and with hovering on an intersection to see its edges, I think lots of the same functionality is in Street Explorer

Issue #248, opened 2024-03-22
Grouping roads/intersections together into logical units


dabreegster:
![image](https://github.com/a-b-street/osm2streets/assets/1664407/124f4b62-d17a-448a-8a2a-0e8d8579e606)
The big pink intersection is surrounded by crossings that overall better define the overall shape of the junction.

![image](https://github.com/a-b-street/osm2streets/assets/1664407/7c5de5ae-421b-40c0-8074-b6c51d88b1a4)
The east/west road similarly has a loop of sidewalks/crossings surrounding it. (Actually this case is more interesting, because there are service roads interrupting things)

In both of these cases, a higher-level logical "view" on the raw edges and nodes is starting to become clear. There have been a few attempts before to handle this kind of thing, like the "zip sidepath" experiments in #210, but they've never made it far. Now that separate footways are showing up much better, I'd like to try again with a fresh idea.

# Part 1: just find a cycle

Starting from a manually chosen node, use DFS to find the smallest distance cycle in the undirected graph. It'd be best to start from nodes connecting only sidewalks (and footway crossings), to represent those corners on a junction. Maybe constrain the search to only/mostly use sidewalks, or wind up with both cycleways and regular road in the resulting loop. Just start by displaying all the matching Intersections and Roads along the way.

# Part 2: trace a polygon

Next, turn that cycle into a "face" / polygon by walking around the cycle's geometry in order. Previous work in the LTN blockfinder, route snapper area mode, and references in topojson / mapshaper. Just start with the original OSM center-lines. If rendered geometry is very nice, then we could try the edges of road and intersection polygons, but that ambition is what makes LTN blockfinding so buggy. So probably just stick to center-lines.

By looking at what a cycle contains, then we can maybe decide to render these polygons. In between sidewalk/crossings and regular road, we could infer that as the full shape of the pavement. Or maybe there's a grass verge between the road and sidewalk.

# Part 3: consider merging / zipping things

This is where previous attempts too quickly leapt. But sometimes it might be useful to simplify the graph structure and copy over the lanes to the "main" road. Many parts of this process are hard, so that's why it'll be an optional step and only after the other two parts work reasonably. If it works, would be an alternate approach to something like https://github.com/nptscot/networkmerge

dabreegster:
One use case is in the taipei example, figuring out which side to put the sidewalk for https://www.openstreetmap.org/way/1064488314 vs https://www.openstreetmap.org/way/1064488313. Need to know orientation relative to the road

PR #250, opened 2024-03-24
Interpreting negative space

dabreegster:
A first step towards #248. Not ready to merge yet, but I wanted to open for feedback on how to use this.

This PR introduces a way to start at a road or intersection, then find a small cycle starting and ending on it, and generate the polygon that (approximately) fills the negative space between everything. Based on the types of roads and intersections in that cycle, the "block" gets classified as a few different cases.

image
The green polygon represents the area between a road and a separately mapped sidewalk. Assuming the road width and sidewalk placement is roughly accurate, then this could have a few interpretations. Sometimes it's a wide pavement, or the planter strip (or "verge"?) often found in suburbia:
image

image
The purple polygon is the space between dual carriageways. In OSM, a road is split into a dual carriageway when there's some kind of phyical separation between the two sides, not just paint. We could try and render something here, maybe.

There are more cases I'll add, like the space between a cycleway and road (hopefully some kind of [separation](https://wiki.openstreetmap.org/wiki/Proposal:Separation) for the cycleway) or the space between cycleway and sidewalk (maybe separation, or maybe just a little curb).

For this first step, I think I just want to settle for finding the polygons, classifying them, and rendering in an off-by-default layer for debugging. I'll first work on finding all of the blocks (problems handling both sides of a road and dealing with dead-ends) and improving the classification.

But more ideas in #248 could come later. We could also change the block-tracing logic to only follow sidewalks and crossings, and wind up with a nice big junction shape around all of something like this:
image
In that case, the big polygon is probably less useful. Seeing all of the Intersections and Roads inside is useful, because then we can reason about vehicle stop lines (behind the crosswalks, probably) and generate high-level movements through the whole junction.

Or similarly, we could try to group all of the parallel-ish OSM ways and have a higher-level representation of all of the street features from right-to-left, no matter how they're split up in OSM. Answering if https://www.openstreetmap.org/way/422264711 has a sidewalk is not straightforward, but with this kind of grouping, it would work.


dabreegster:
Alright, I got the block-finder to pretty much handle everything except dead-ends / edge of the map. Added a few more categories in the classification, but each of them still needs some work. And improved the debugging UI a bit. Going to merge and take smaller steps.

image
image

PR #251, opened 2024-03-29
Trace a large block around sidewalks and crossings only

dabreegster:
A simple next step towards #248 --when tracing a block, if we only consider crossings and sidewalks AND the road has sidewalks on both sides, then we wind up with a nice grouping for the "entire" junction or "entire" road including side-paths.

Here are some junctions of varying complexity:
image
image
image

And some roads:
image
image

I think this approach of grouping entire "bundles" of roads or junctions could be used for various purposes:

- for od2net or other route network generation, a way to simplify the network topology either before or after the routing. CC @Robinlovelace and also see https://github.com/nptscot/networkmerge
- for producing a simpler graph for interactive routing. Sidepaths and the main road could all be grouped together, then a cost function can look at any separated cycleways. CC @tordans
- for generating movements across a complex junction or reasoning about building-to-building width. CC @sgreenbury and @andrewphilipsmith
- for rendering or inferring geometry in this library

I'll merge this so people can play with it at https://osm2streets.org. Click a sidewalk, then "Trace sidewalks on left/right" (the direction is... not always consistent, but one should work). Please do comment with any interesting test cases you find, either working or not!

### DIFF
--- a/osm2streets-js/src/lib.rs
+++ b/osm2streets-js/src/lib.rs
@@ -251,9 +251,9 @@ impl JsStreetNetwork {
     }
 
     #[wasm_bindgen(js_name = findBlock)]
-    pub fn find_block(&self, road: usize, left: bool) -> Result<String, JsValue> {
+    pub fn find_block(&self, road: usize, left: bool, sidewalks: bool) -> Result<String, JsValue> {
         self.inner
-            .find_block(RoadID(road), left)
+            .find_block(RoadID(road), left, sidewalks)
             .map_err(err_to_js)?
             .render_polygon(&self.inner)
             .map_err(err_to_js)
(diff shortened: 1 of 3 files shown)


dabreegster:
Are there cases where the "entire" road including side-paths shouldn't be bounded on all sides by footways and crossings? Cases with sidepaths on only one side of a road will take some work. But are there any other real-world situations where maybe a cyclepath is on the "outside" and the pavement on the "inside"?

Robinlovelace:
Those are some complicated junctions in there! Fully agree that grouping all the elements into blocks makes sense. Not had a play but intend to do so after the Easter break. Thanks for the link to Networkmerge, not production ready as ineffiecient and is mosly focussed on merging multiple lanes into a single lane so may not be that relevant here.

PR #261, opened 2024-04-04
Get rid of the old sausage link transformation

dabreegster:
It actually has no effect on tests, so the transformation must've broken at some point, unnoticed. New general approach is to represent OSM more faithfully. The new junction/corridor bundling experiment is more promising as the basis of an optional operation to produce a simplified graph.


osm2streets/src/transform/mod.rs
```diff
@@ -7,59 +7,33 @@ mod collapse_short_road;
...
         vec![
-            Transformation::CollapseSausageLinks,
             Transformation::CollapseShortRoads,
             Transformation::CollapseDegenerateIntersections,
             // The above may discover more roads to collapse
             Transformation::CollapseShortRoads,
         ]
     }
 
-    /// A full suite of transformations for A/B Street.
```
dabreegster, line 34: I don't plan on updating A/B Street anytime remotely soon. This code could move there if needed.

dabreegster (commented):
The dual carriageway and zip sidepath transforms/operations are also unused/experimental code that're both probably superceded by the newer block-tracing + bundling approach. Should we remove those too? (Less code makes refactors easier, and git history isn't going anywhere if we ever want to refer back)

PR #262, opened 2024-04-05
Enable manual snapshot testing for blocks

dabreegster:
@BudgieInWA, I want to quickly detect which test cases have different output for the new block-tracing logic, so this is how I'm doing it. I'm gitignoring the files and disabling the new test for now, because I don't want to bloat the git history with a bunch of new files this early in development. But this general approach could be an example for more things, like lane markings

### DIFF
--- a/tests/src/.gitignore
+++ b/tests/src/.gitignore
@@ -1,1 +1,2 @@
 */*.orig.*
+*/blocks.json
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -21,10 +21,6 @@ mod tests {
         let mut timer = Timer::new("test osm2streets");
 
         println!("Working on {path}");
-        // Read the output file before modifying it. If it doesn't exist, then we're creating a new
-        // test case.
-        let prior_json = std::fs::read_to_string(format!("{path}/geometry.json"))
-            .unwrap_or_else(|_| String::new());
 
         let clip_pts = if Path::new(format!("{path}/boundary.json").as_str()).exists() {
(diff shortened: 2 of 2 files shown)


BudgieInWA:
From a quick read, this seems to be the approach I was thinking of.

It would be cool to make use of https://github.com/mitsuhiko/insta for managing the snapshot files. It might be fiddly getting it to behave in our test function, but it's probably worth conforming to their expectations anyway. And it might mean losing control over the snapshot file names.

`insta` implements the workflow that I am familiar with: write `*.snap.new` when a difference is detected, and use `cargo insta review` to do "accept" the change, doing the file rename that updates the snapshot.

Using `.new` instead of `.orig` loses the easy `git diff` of snapshot files, but gains the ability to quickly iterate on code, rerunning the test until you're happy with the output.

Either way, Street explorer should learn to render snapshot diffs for spatial data. And it could even learn to do the "accept" step, becoming a replacement for `cargo insta review`.

dabreegster:
`insta` seems like a nice way of uniformly handling snapshot tests, yeah. The text diffs it provides won't be too useful for us. How do you imagine street explorer showing diffs -- just a way to toggle between two outputs easily, or anything beyond that? I'm slowly starting a work project, https://github.com/dabreegster/geodiffr, meant to interactively understand diffs between two arbitrary GJ files. https://github.com/dabreegster/geodiffr/issues/3 is one intended use case I want to try

BudgieInWA:
In particular, detecting which tests are failing and presenting that list, then opening up both the baseline and new file when the test is selected. Just like the preview command would do.

PR #263, opened 2024-04-05
Improve block detection

dabreegster:
Some more steps towards #248. Individual blocks involving dead-end roads now mostly work, thanks to bringing in code from https://github.com/a-b-street/abstreet/blob/main/blockfinding/src/lib.rs. (Why solve problems twice?)

And then improve block classification by looking only at the lane closest to the block, not the whole road. Many cases improve.

I looked at loads of before/afters through this, but didn't take any screenshots this time.

### DIFF
--- a/osm2streets/src/block.rs
+++ b/osm2streets/src/block.rs
@@ -272,15 +272,12 @@ fn classify_block(streets: &StreetNetwork, boundary: &Vec<RoadSideID>) -> BlockK
     let mut has_sidewalk = false;
 
     for road_side in boundary {
-        // TODO All of this logic is wrong. Look at the outermost lane, not the whole road.
-        let road = &streets.roads[&road_side.road];
-        if road.is_driveable() {
-            // TODO Or bus lanes?
+        let lt = road_side.get_outermost_lane(streets).lt;
+        if lt == LaneType::Driving || lt == LaneType::Bus {
             has_road = true;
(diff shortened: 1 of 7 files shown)


osm2streets/src/block.rs
```diff
@@ -168,123 +151,134 @@ impl Block {
...
+    let (orig_from_intersection, start_intersection) = {
+        let dir = start_road_side.get_outermost_lane(streets).dir;
+        let road = &streets.roads[&start_road_side.road];
+        if dir == Direction::Forward {
+            (road.src_i, road.dst_i)
+        } else {
+            (road.src_i, road.dst_i)
+        }
+    };
+    let mut current_intersection = start_intersection;
```
dabreegster, line 185: I feel like all of this logic ought to be simpler if we trace tuples (road, left / right, forwards / backwards). I might try that in a future PR.

osm2streets/src/intersection.rs
```diff
@@ -130,6 +130,56 @@ impl Intersection {
             format!("{} ({})", self.id, osm_ids)
         }
     }
+
+    // TODO Use RoadEdge?
+    // This skips the "interior" piece of any loop roads
+    pub fn get_road_sides_sorted(&self, streets: &StreetNetwork) -> Vec<RoadSideID> {
```
dabreegster, line 136: This would be quite nice to explore in a smaller unit test, because I've found it confusing

PR #264, opened 2024-04-05
Try bundle classification based on the number of major roads

dabreegster:
#248. I'm looking at a particularly uniform area in California that has a clear pattern of junctions, corridors, and land use blocks. Before this PR, the classes were a mess:
image
With this PR, most cases become correct:
image

Some of the misclassifications will be tricky to remedy. Some more ideas to think about later:
- does the cumulative length of roads at different angles inside the bundle tell us anything?
- if we look at the connections between bundles, even if they're misclassified, can we start to put together a simplified graph?


osm2lanes/src/osm.rs
```diff
@@ -4,52 +4,3 @@ pub use osm_reader::{NodeID, OsmID, RelationID, WayID};
 
 // This is a commonly used key in the codebase, so worthy of a bit of typo-prevention
 pub const HIGHWAY: &str = "highway";
-
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
```
dabreegster, line 8: All unused in this codebase and the list of highway types has become out-of-date with everything now. So removing.

PR #269, opened 2024-08-25
Replace `country-geocoder` with `country-boundaries`

### LINKED ISSUE #267: Wrong driving side for non-sovreign LHT jurisdictions inside RHT countries
Bidirectional roadways don't change to the correct side in Hong Kong, Macau, and US Virign Islands. They are `admin_level=3` "states", causing the geocoder to not recognize them as "country", and this difference skipped. 
This results in wrong lane layout (asymmetric lane count roads, and becomes inconsistent with divided roads), road marking orientation, and "movement arrows". I haven't noticed what happens with sidewalks. 
UK RHT overseas territories don't have this problem, being `admin_level=2` https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic#Worldwide_distribution_by_country

JamesHinshelwood:
Instead of relying on [`country-geocoder`](https://github.com/a-b-street/country-geocoder) to find a country code, we now use
[`country-boundaries`](https://github.com/westnordost/country-boundaries-rust/). This makes the lookup faster, but more importantly it also returns correct results for regions and countries which are not `admin_level=0`, such as Hong Kong or Macau.

`country-boundaries` does not include data on whether a country drives on the left or right, however thankfully `muv-osm` does, so we use its data instead (which is generated from `wikidata.org`).

Fixes #267 

Example: https://a-b-street.github.io/osm2streets/#18.47/22.3022565/114.0112691

Before
image

After
image

### DIFF
--- a/streets_reader/src/lib.rs
+++ b/streets_reader/src/lib.rs
@@ -5,8 +5,10 @@ extern crate log;
 
 use abstutil::Timer;
 use anyhow::Result;
+use country_boundaries::{CountryBoundaries, BOUNDARIES_ODBL_60X30};
 use geom::{GPSBounds, LonLat, Ring};
 
+use muv_osm::lanes::highway::driving_side;
 use osm2streets::{DrivingSide, MapConfig, StreetNetwork};
 use osm_reader::Document;
 
@@ -52,16 +54,21 @@ pub fn detect_country_code(streets: &mut StreetNetwork) {
         return;
     }
 
-    let geocoder = country_geocoder::CountryGeocoder::new();
-    let pt = streets.gps_bounds.get_rectangle()[0].into();
+    let geocoder = CountryBoundaries::from_reader(BOUNDARIES_ODBL_60X30).unwrap();
+    let pt = streets.gps_bounds.get_rectangle()[0];
 
-    if let (Some(code), Some(left_handed)) = (geocoder.iso_a2(pt), geocoder.drives_on_left(pt)) {
-        streets.config.driving_side = if left_handed {
-            DrivingSide::Left
-        } else {
-            DrivingSide::Right
-        };
+    // Obtain the possible list of country codes, ordered by their size ascending.
+    let codes = geocoder.ids(country_boundaries::LatLon::new(pt.y(), pt.x()).unwrap());
+    // Use the first code, since it is most specific.
+    if let Some(code) = codes.first() {
         streets.config.country_code = code.to_string();
+        let driving_side = match driving_side(&code) {
+            Some(muv_osm::lanes::Side::Left) => DrivingSide::Left,
+            Some(muv_osm::lanes::Side::Right) => DrivingSide::Right,
+            // Default to driving on the right.
+            None => DrivingSide::Right,
+        };
+        streets.config.driving_side = driving_side;
     } else {
         error!("detect_country_code failed -- {:?} didn't match to any country. Driving side may be wrong!", pt);
     }
--- a/streets_reader/src/lib.rs
+++ b/streets_reader/src/lib.rs
@@ -59,19 +59,18 @@ pub fn detect_country_code(streets: &mut StreetNetwork) {
 
     // Obtain the possible list of country codes, ordered by their size ascending.
     let codes = geocoder.ids(country_boundaries::LatLon::new(pt.y(), pt.x()).unwrap());
-    // Use the first code, since it is most specific.
-    if let Some(code) = codes.first() {
-        streets.config.country_code = code.to_string();
-        let driving_side = match driving_side(&code) {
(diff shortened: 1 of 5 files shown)


streets_reader/src/lib.rs
```diff
@@ -52,16 +54,21 @@ pub fn detect_country_code(streets: &mut StreetNetwork) {
         return;
     }
 
-    let geocoder = country_geocoder::CountryGeocoder::new();
-    let pt = streets.gps_bounds.get_rectangle()[0].into();
+    let geocoder = CountryBoundaries::from_reader(BOUNDARIES_ODBL_60X30).unwrap();
```
JamesHinshelwood, line 57: I opted to use the smallest, but slowest raster here, which is 296 KB. We could opt for a larger, but faster raster too.

streets_reader/src/lib.rs
```diff
@@ -52,16 +54,21 @@ pub fn detect_country_code(streets: &mut StreetNetwork) {
...
+    // Use the first code, since it is most specific.
+    if let Some(code) = codes.first() {
         streets.config.country_code = code.to_string();
+        let driving_side = match driving_side(&code) {
+            Some(muv_osm::lanes::Side::Left) => DrivingSide::Left,
+            Some(muv_osm::lanes::Side::Right) => DrivingSide::Right,
+            // Default to driving on the right.
+            None => DrivingSide::Right,
+        };
+        streets.config.driving_side = driving_side;
```
JamesHinshelwood, line 71: There is an alternative algorithm here which would handle the case where `muv-osm` doesn't have an entry for a our most-specific country code, but might for a less specific one:

```rust
let driving_side = codes.iter().find_map(|code| driving_side(&code)).unwrap_or(DrivingSide::Right);
```

However, in this case I wasn't sure whether we should retain the more specific country-code in `streets.config.country_code` or set that to the more general code too.

Happy to change this if desired.

dabreegster: I think this is necessary. The code for England is now "GB-ENG", not "GB", but Muv only seems to have two-letter codes. So without doing this, some of the tests break.

I made this change in https://github.com/a-b-street/osm2streets/tree/region-driving-side_debug, but very oddly, I'm still seeing the `borough_sausage_links` test case break. If you `cargo test`, you'll get a bunch of git diffs for built-in cases. Many are expected (the country code getting more specific), but when the GeoJSON changes, I visually compare https://a-b-street.github.io/osm2streets with the local version (`cd web; npm run wasm && npm run dev`). I'm not sure what's going wrong though...

dabreegster (commented):
Thank you for the PR! It'll be great to git rid of `country-geocoder` in favor of this much more rigorous crate. I'm seeing some test diffs I don't understand yet, but after we sort those out, happy to merge this

streets_reader/src/lib.rs
JamesHinshelwood: Ah sure, thanks for pointing those out. I'll run the tests and dig further into the diffs. I agree that we should use the most specific code for which we know a driving side.

JamesHinshelwood: I made this change and have updated the test cases. I think the changes are mostly reasonable.

The two cases that were changed were:
* Frederiksted - Because they drive on the left.
* Fremantle - Because the [country boundaries](https://github.com/a-b-street/country-geocoder/blob/main/data.geojson) in `country-geocoder` cut off some of western Australia, meaning this wasn't considered part of AU.

The actual diffs in those files are composed of:
* `country_code` now has a value for those two cases.
* `Intersection` -> `Fork`: An intersection that used to have conflicts no longer does (e.g. a left side-road when driving on the left).
* `Fork` -> `Intersection`: An intersection that used to have no conflicts now does.
* `Intersection` -> `Connection`: This one confuses me a bit. When I look at the intersections it seems like they always should have been `Connection`s  (e.g. https://a-b-street.github.io/osm2streets/#20.23/17.7147832/-64.8799136), but I haven't yet understood `calculate_movements_and_kind` well enough to see why they were `Intersection`s previously.

JamesHinshelwood: I haven't tried running it properly, but I _think_ the difference between my latest commit and your `region-driving-side_debug` is that I now set `streets.config.country_code` to the most specific code for which we have a driving side, whereas your branch still sets it to the most specific code regardless. I guess this causes later lookups based on the `config.country_code` to return different results.

dabreegster: Thank you for the quick fix and digging into the changes! `calculate_movements_and_kind` is still quite experimental code, and the changes to `Connection` there seem benign.

dabreegster (approved):
This is great, thank you for fixing this bug! I haven't spent time on osm2streets in a while, so I really appreciate help improving it

